PR #13, opened 2020-06-06
Refactoring the Runtime class

Chemaclass:
# Description

Improvements applied to the `Phel\Runtime` class:

* Fixed the typo
  * from: "cacheDiretory" (the "c" of the "Dire**c**tory" word is missing!)
  * to: "cacheDirectory"
* Applying `strict_types`
* Remove unnecesary "elses"

### DIFF
--- a/src/php/Runtime.php
+++ b/src/php/Runtime.php
@@ -29,38 +29,38 @@ class Runtime {
     /**
      * @var string|null
      */
-    private $cacheDiretory;
+    private $cacheDirectory;
 
     /**
      * @var Runtime|null
      */
     private static $instance;
 
-    private function __construct(GlobalEnvironment $globalEnv = null, string $cacheDiretory = null) {
+    private function __construct(GlobalEnvironment $globalEnv = null, string $cacheDirectory = null) {
         set_exception_handler(array($this, 'exceptionHandler'));
 
         if (is_null($globalEnv)) {
             $globalEnv = new GlobalEnvironment();
         }
         $this->globalEnv = $globalEnv;
-        $this->cacheDiretory = $cacheDiretory;
+        $this->cacheDirectory = $cacheDirectory;
         $this->addPath('phel\\', [__DIR__ . '/../phel']);
     }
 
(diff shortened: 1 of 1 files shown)


src/php/Runtime.php
```diff
@@ -11,63 +13,46 @@
...
+    public static function newInstance(GlobalEnvironment $globalEnv = null, string $cacheDirectory = null): self {
+        return new Runtime($globalEnv, $cacheDirectory);
     }
 
-    /**
-     * @param Exception $exception
-     */
-    public function exceptionHandler($exception) {
-        if (php_sapi_name() == 'cli') {
+    public function exceptionHandler(Exception $exception): void {
```
jenshaase, line 54: Actually `Exception` is wrong. It must be `Throwable`. See https://www.php.net/manual/en/function.set-exception-handler.php

Chemaclass: Done

src/php/Runtime.php
```diff
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
```
jenshaase, line 3: I haven't use `declare(strict_types=1);` nowhere in this project currently. I think it should be either everywhere or nowhere. In general I'm haven't used this feature yet myself and have no idea what are the effects on the rest of the codebase. It would be better to remove it for now and make an issue to discuss the advantages and disadvantages first.

Chemaclass: The good think about this line itself is that it applies actually to ONLY this file. It ensures that this file has strict_types. It doesn't apply to any other file in the whole project. That's the best thing about it. It allows us to do step by step this migration from non-strict to strict code. Especially for new files or projects.

jenshaase:
Thank @Chemaclass. I merged it into master!

src/php/Runtime.php
Chemaclass: @jenshaase take a look at this post, I forgot to write it down in my previous comment https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md 
I will publish a [medium](https://medium.com/@chemaclass) post about this topic soon. I think it's important to let the people know about this :)

PR #16, opened 2020-06-07
Remove psalm errors

Chemaclass:
# Description

## Fixing psalm and cleaning code

In this PR the main focus was to fix all issues/errors that (even in the level 3) psalm was reporting.
I went one by one trying to understand them and fix them. Most of them were about possible nulls types (an `Assert` fixed it) or about missing or unknown types (adding the explicit type fixed it).

I also clean myself some condition statements (getting rid of non-necessary elses), and, in some places, return early in order to decrease the indentation of the whole block.

## Additional suggestions

Some suggestions that I didn't want to keep modifying (but potentially in further contributions) and, why not, a general review of what I saw so far:
* Use strict comparison: `===` instead of `==`.
* Use single quotes for strings (`'string'`). [Double quotes strings are useful when you want to render/parse some variables into a string itself.](https://stackoverflow.com/questions/3446216/what-is-the-difference-between-single-quoted-and-double-quoted-strings-in-php) Otherwise, a single quote string is a literal string and the way PHP works with strings.
* Try to avoid writing "else". You're already returning early in most of the places (that's great, btw!), therefore the `else` statement is useless, you don't need it at all.
* I saw some `$x`(see `analyzeBindings()`) and `$b`(see `analyzeLoop()`) variables which are hard to mental-map with their real meaning behind. Try to avoid such names. Instead, `$tuple` and `$symbol`, for example.
* Keep your `use` import ordered by alphabetical.

That being said, in general, is a really good code 👍


src/php/Analyzer.php
```diff
@@ -592,6 +597,7 @@ protected function analyzeRecur(Tuple $x, NodeEnvironment $env): RecurNode {
 
         $exprs = [];
         for($i = 1; $i < $tupleCount; $i++) {
+            Assert::isInstanceOf($x[$i], Phel::class);
```
jenshaase, line 600: This is wrong `$x[$i]` can be `Phel|scalar|null`

src/php/Analyzer.php
```diff
@@ -87,10 +85,7 @@ private function isLiteral($x): bool {
     }
 
     /**
-     * @param Phel|scalar|null $x
-     * @param ?NodeEnvironment $nodeEnvironment
-     * 
-     * @return Node
+     * @param mixed $x
```
jenshaase, line 88: `mixed` is much more open than `Phel|scalar|null`. The original value was correct.

src/php/Analyzer.php
```diff
@@ -571,6 +575,7 @@ protected function analyzeRecur(Tuple $x, NodeEnvironment $env): RecurNode {
         $frame = $env->getCurrentRecurFrame();
 
         if (!$frame) {
+            Assert::isInstanceOf($x[0], ISourceLocation::class);
```
jenshaase, line 578: We can be more strict here `$x[0]` must be the Symbol 'recur'. Is is implicitly handled by the `analyze` function that has checked that and then called the this method.

Chemaclass:
Yes, I will do that. Thanks for the feedback!

PR #18, opened 2020-06-07
GitHub workflow

Chemaclass:
# Description

Similar as I have in some of my other projects ([EdifactParser](https://github.com/Chemaclass/EdifactParser/blob/master/.github/workflows/ci.yml) for example), or even in the [PHPUnit](https://github.com/sebastianbergmann/phpunit/blob/master/.github/workflows/ci.yml), I introduce a basic workflow with two jobs:
* Type Checker: runs `psalm`
* Tests: runs the tests to the compiler itself and the core library

Note:
I increased the level (from 3 to 4) in the `psalm.errorLevel` because psalm had 9 errors which are out of the scope of this PR. The current level is 4 (when it doesn't complain).

### DIFF
--- a/src/php/Reader.php
+++ b/src/php/Reader.php
@@ -144,7 +144,7 @@ class Reader {
                     $this->fnArgs = [];
                     $body = $this->readList($tokenStream, Token::T_CLOSE_PARENTHESIS);
 
-                    if (count($this->fnArgs) > 0) {
+                    if (!empty($this->fnArgs)) {
                         $maxParams = max(array_keys($this->fnArgs));
                         $params = [];
                         for ($i = 1; $i <= $maxParams; $i++) {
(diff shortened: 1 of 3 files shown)


src/php/Reader.php
```diff
@@ -144,7 +144,7 @@ public function readExpression(Generator $tokenStream) {
                     $this->fnArgs = [];
                     $body = $this->readList($tokenStream, Token::T_CLOSE_PARENTHESIS);
 
-                    if (count($this->fnArgs) > 0) {
+                    if (!empty($this->fnArgs)) {
```
Chemaclass, line 147: This was the unique error that psalm complained once I changed to `errorLevel=4`.

jenshaase:
Well done!

jenshaase:
It's working great! Thanks!

PR #19, opened 2020-06-07
Cleaning Analyzer class

Chemaclass:
# Description

* Making the class [strict](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md)
* Making the class [final](https://medium.com/@chemaclass/final-classes-in-php-9174e3e2747e)
* Using [strict comparations](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/performance-tips.md): `===` instead of `==`
* Removing unnecessary `else` statements due to return early (really good job there!)
  * It also decreases the indentation of the blocks which helps a lot to follow the logic
* Change from `protected $phpKeywords` to `private const PHP_KEYWORDS` you might guess why :)

### DIFF
--- a/src/php/Analyzer.php
+++ b/src/php/Analyzer.php
@@ -48,7 +48,7 @@ use Phel\Lang\Tuple;
 
 final class Analyzer {
 
-    protected GlobalEnvironment $globalEnvironment;
+    private GlobalEnvironment $globalEnvironment;
 
     private const PHP_KEYWORDS = [
         '__halt_compiler', 'abstract', 'and', 'array', 'as', 'break', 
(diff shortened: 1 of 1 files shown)


src/php/Analyzer.php
```diff
@@ -99,68 +97,78 @@ public function analyze($x, ?NodeEnvironment $nodeEnvironment = null): Node {
...
-                    default:
-                        return $this->analyzeInvoke($x, $nodeEnvironment);
-                }
-            } else {
-                return $this->analyzeInvoke($x, $nodeEnvironment);
-            }
-        } else {
+        }
+
+        if (!$x instanceof Tuple) {
```
jenshaase, line 118: This is the only case I would not split. It think it is much easier to understand if you have it in the following form
```php
<?php
if ($x instanceof Tuple) { // Matches the style of the ifs above
  if ($x[0] instanceof Symbol) {
    switch ... 
  }

  return $this->analyzeInvoke($x, $nodeEnvironment);
}

// This Exception is now at the end and it's clear that a case above is missing
// If the exception is somewhere in between it is much harder to understand the logic of the code.
throw new AnalyzerException('Unhandled type: ' . var_export($x, true), null, null);
```

Chemaclass: achso! ok, got it!

jenshaase:
Nice work! Merged in into master.

PR #21, opened 2020-06-08
Using null coalescing operator

Chemaclass:
# Description

Using null coalescing operator `$foo ?? $bar` instead of `isset($foo) ? $foo : $bar`

[Modern PHP: know the best PHP features up to PHP 7.4](https://medium.com/@chemaclass/modern-php-know-the-best-php-features-til-php-7-4-e3b2bf5916b2)


jenshaase:
Thanks. I just merged it!

PR #22, opened 2020-06-08
Cleaning Printer class

Chemaclass:
# Description

* Making the class [strict](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md) and [final](https://medium.com/@chemaclass/final-classes-in-php-9174e3e2747e).


jenshaase:
Thanks! It is now merged.

PR #28, opened 2020-06-09
Cleaning Emitter class

Chemaclass:
# Description

* Making the class [strict](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md) and [final](https://medium.com/@chemaclass/final-classes-in-php-9174e3e2747e).
* No logic was altered.

### DIFF
--- a/src/php/Emitter.php
+++ b/src/php/Emitter.php
@@ -1,14 +1,14 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Phel;
 
 use Exception;
-use ParseError;
 use Phel\Ast\ApplyNode;
 use Phel\Ast\ArrayNode;
 use Phel\Ast\CallNode;
 use Phel\Ast\CatchNode;
-use Phel\Ast\Node;
 use Phel\Ast\DefNode;
 use Phel\Ast\DefStructNode;
 use Phel\Ast\DoNode;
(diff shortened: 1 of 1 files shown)


jenshaase:
Thanks. It is now merged.

src/php/Emitter.php
```diff
@@ -96,21 +78,21 @@ public function emitAndEval(Node $node): string
...
-                throw $e;
-            }
-        } else {
+        if (!$filename) {
             throw new Exception("can not create temp file.");
         }
 
+        try {
+            file_put_contents($filename, "<?php\n" . $code);
+            $result = require $filename;
```
Chemaclass, line 87: @jenshaase Actually here, I would do the return directly. You don't really need to save the "require result" in a variable to return it later right after if no exception occurred.
```php
return require $filename;
```
The context is pretty strait forward here, and there is either `return something` or `throw $e`.
```php
try {
    file_put_contents($filename, "<?php\n" . $code);
    return require $filename;
} catch (Throwable $e) {
    throw $e;
}
```
Anyway, it's a minor thing. But just to let you know :)

jenshaase: You are right. Missed this! Feel free to change this.

PR #29, opened 2020-06-09
#28 Emitter class - Return require inline

JesusValera:
# Description

Implement @Chemaclass recommendation in the https://github.com/jenshaase/phel-lang/pull/28#discussion_r437644221 comment about return inline the `require`.


jenshaase:
Thanks! Just merged it!

PR #30, opened 2020-06-10
Adding docker using PHP:7.4

Chemaclass:
# Description

* Adding docker-compose with docker using `PHP:7.4` for dev env
* Renamed the script names for the tests:
  * `phpunit` to `test-compiler`
  * `test` to `test-core`
* Added `composer psalm` and `composer csfix` to "composer scripts"
* Ignore `.idea/` directory (from PHPStorm)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
         run: composer install --prefer-dist --no-progress --no-suggest
 
       - name: Run tests to the compiler itself
-        run: composer run-script phpunit
+        run: composer run-script test-compiler
 
       - name: Run tests to the core library
-        run: composer run-script test
+        run: composer run-script test-core
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ data
 doc/public
 .phel-repl-history
 .php_cs.cache
+.idea/
--- a/Readme.md
+++ b/Readme.md
@@ -10,7 +10,18 @@ The rest of this Readme file documents on how to work with the compiler code.
 
 Feel free to ask questions and join discussions on the [Phel Gitter channel](https://gitter.im/phel-lang/community).
 
-## Build the documentation
+
+## Development and contribution
+
+### Your first try!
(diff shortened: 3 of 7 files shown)


jenshaase:
Can you just explain the benefits of adding a Docker file?

Currently the requirement is only PHP 7.4 and Composer. It's not that complicated and most people working with PHP know how to install this. Sure, some people might prefer to use docker but they also know how to setup their own containers.

My idea is to keep the maintaining cost of the project focused on what's really important. Adding such things increases the maintaining cost, with no real benefit.

JesusValera:
> Can you just explain the benefits of adding a Docker file?
> 
> Currently the requirement is only PHP 7.4 and Composer. It's not that complicated and most people working with PHP know how to install this. Sure, some people might prefer to use docker but they also know how to setup their own containers.
> 
> My idea is to keep the maintaining cost of the project focused on what's really important. Adding such things increases the maintaining cost, with no real benefit.

I do not agree with you 100%.
My company's computer has PHP 7.3, we don't have any application running the last version so I don't need it, also I don't want to spent a lot of time with extra configuration (like enabling extensions and adding xDebug for PHP 7.4, in Ubuntu is a bit complicated), adding Docker helps us not only do not depend if the user has or not the exactly PHP version, but to create a "standard" version for everybody even if in their computer does not have any PHP version.

About your second comment, I do not think it "increases the maintaining cost" at all, you only created the `docker-compose` file once and that's all.
We are giving more facilities and opportunities for other people to try the project. :stuck_out_tongue:

Chemaclass:
Yes, @JesusValera answered faster than me. I wanted to say something similar. 
Nowadays, unfortunately, most devs/projects still run in 7.2/7.3.
Having this docker image standards the development environment, specially talking about a system that it's getting developed in different machines.
Jesus already mentioned a real use case above which this PR intends to solve.

jenshaase:
Did another small change: I moved the dockerfile from `devops/php` to `build`.

It is now merged.

PR #32, opened 2020-06-11
Cleaning Table and Struct classes

Chemaclass:
# Description

* Declaring [strict types](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md).
* Defining type properties.
* Adding return types to functions. 
* Strict conditions when possible.
  * The only one case when it's not possible to define as strict condition is in the `Table::equals($other)` method.
* No logic was altered.


jenshaase:
Thank you!

PR #33, opened 2020-06-11
install zip in dockerfile

paullaffitte:
In order to improve performances when running `composer install`, I installed zip to allow composer to use _dist_ instead of _sources_.

image


jenshaase:
Thank you!

PR #36, opened 2020-06-11
Fix Dockerfile dependencies

JesusValera:
# Description

When you try to install the project from Docker, for any reason, when it starts with the dependencies'  installation (`apt install -y git zip`), the process is cancelled.

I try some scenarios and for some strange reason if you remove the `RUN` keywords and replaced them by `&&`. With the proposal PR, it works like a charm.

## Scenario:
1. `docker-compose up`

Screenshot 2020-06-11 at 21 12 26


jenshaase:
Thank you. I merged it.

PR #37, opened 2020-06-11
Cleaning GlobalEnvironment class

JesusValera:
# Description

* Making the class strict & final
* Rename the method `addDefintion()` to `addDefinition()`.
* Define the type and make properties private (better encapsulation).
* Simplify an `isset()` _L45_
   * It is not necessary to check twice if the key and sub-key exists. [DEMO](http://sandbox.onlinephpfunctions.com/code/11557f70adf59254dde4705685dc6275f067de2d)
* Removing unnecessary `else` statements.
* Split the logic from `addRequireAlias()` into `resolveWithAlias()` & `resolveWithoutAlias()`.
* No logic was altered.

### DIFF
--- a/src/php/Analyzer.php
+++ b/src/php/Analyzer.php
@@ -1394,7 +1394,7 @@ final class Analyzer
             );
         }
 
-        $this->globalEnvironment->addDefintion($namespace, $name, $meta);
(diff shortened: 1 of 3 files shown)


jenshaase:
Thank you! I merged it!

PR #43, opened 2020-06-12
Cleaning Runtime class

JesusValera:
# Description

- Remove unnecessary imports.
- Rearrange properties & methods (static properties on top always).
- Define more strict exceptions (`\RuntimeException` instead of generic `\Exception`).
- Changes in `findFile()` method.
    - Return only one object type (`string`) instead of two (`string` or `bool` if error).
    - Replace global `strtr()` by `str_replace()` function (more legiblility).
- Changes in `loadNs()` method:
   - [Add `true` flag](https://www.php.net/manual/en/function.in-array.php) in `in_array()` global function.
    - Remove one identation level because of `findFile()` return type.
- No logic was altered.

### DIFF
--- a/src/php/Runtime.php
+++ b/src/php/Runtime.php
@@ -4,8 +4,6 @@ declare(strict_types=1);
 
 namespace Phel;
 
-use Exception;
-use Phel\Exceptions\AnalyzerException;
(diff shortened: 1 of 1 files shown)


jenshaase:
Thanks. Merged it.

PR #44, opened 2020-06-13
Refactoring Repl 

Chemaclass:
# Description

I mainly refactor the `Repl` class applying some techniques as extract method and moving up some variables from the `run()` method to the class level (as members/class-properties).

# Changes

- No logic was altered.
- Added `array_syntax: short` to the `php_cs`.
- Added `"ext-readline": "*"` to the `composer.json` because it's actually required by the `Readline:readline()`(line:54)
- Modified the `Repl`:
  - Making the class [final](https://medium.com/@chemaclass/final-classes-in-php-9174e3e2747e) and [strict](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md).
  - Move to class properties the variables (defined in the `run()` method) that are outside the `while(true)` loop, and behave like class properties.
  - Applied "extract method" refactoring to the input processing:
    - `readInput()`: Read and validate the input
    -  `analyzeInput()`: Once it was read it and validated then apply the analyzer logic


src/php/Repl.php
```diff
@@ -8,106 +10,114 @@
...
+     */
+    private function readInput($input): void
     {
-        return fgets(STDIN);
+        if (false === $input) {
+            $this->output($this->color("Bye from Ctrl+D!\n", 'yellow'));
+            exit;
+        }
+
+        if ('quit' === $input) {
```
Chemaclass, line 79: I would define not only `quit` but also `exit`.  What do you think, @jenshaase?
```php
if (in_array($input, ['quit', 'exit'], true)) { // extracting this array into a `priv const`
    $this->output($this->color("Bye!\n", 'yellow'));
    exit;
}
```

src/php/phel.php
```diff
@@ -31,13 +31,9 @@ function getHelpText(): string
...
-switch ($argv[1]) {
-    case "repl":
-        $repl = new Repl($currentDir);
-        $repl->run();
-        break;
-
-    default:
-        echo getHelpText();
-        exit;
+if ('repl' === $argv[1]) {
```
jenshaase, line 34: Could we leave this as switch? Currently there is just the `repl` command but there will be more commands in the future.

src/php/Repl.php
jenshaase: I would go either for "quit" or for "exit". But you are right. "exit" is probably more common to PHP programmers, because of the "exit" function. @JesusValera Can you change "quit" to "exit"

src/php/Repl.php
```diff
@@ -8,106 +10,114 @@
...
-
-        $reader = new Reader();
-        $lexer = new Lexer();
-        $analzyer = new Analyzer($globalEnv);
-        $emitter = new Emitter();
-        $printer = new Printer();
-        $exceptionPrinter = new TextExceptionPrinter();
-
         $this->output($this->color("Welcome to the Phel Repl\n", 'yellow'));
         $this->output('Type "quit" or press Ctrl-D to exit.' . "\n");
```
jenshaase, line 42: This also needs to be changed from "quit" to "exit"

composer.json
```diff
@@ -12,7 +12,8 @@
         }
     ],
     "require": {
-        "php": ">=7.4"
+        "php": ">=7.4",
+        "ext-readline": "*"
```
jenshaase, line 16: This should be moved to "require-dev". Since the Repl is just a tool for development and not for production.

src/php/phel.php
Chemaclass: Natürlich :)

jenshaase:
Thanks. It is merged now.

PR #47, opened 2020-06-13
Extracting the color logic to ColorStyle

Chemaclass:
# Description

I saw there was a duplicated function named `function color()` so I created a class to encapsulate that logic: `ColorStyle`.
I also refactored the `Printer::print()` method to get rid of the optional second parameter `$readable`. 

# Changes

- I introduced two named constructors to the `Printer` class:
  - `Printer::readable(): Printer`
  - `Printer::nonReadable(): Printer`
- I get rid of all usages of `new Printer()` in order to use one of its proper named constructors.
- I updated the `core.phel` to use the  `nonReadable` method constructor.
- Some variable naming improvements.
- Strict comparations (when possible).
- Specifying the return type of the functions (when possible).


src/phel/core.phel
```diff
@@ -1269,8 +1269,8 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
   The resulting string is returned."
   [& xs]
   (let [len (count xs)
-        printer (php/new Printer)
-        pp |(php/-> printer (print $ false))]
+        printer (php/new Printer false)
```
Chemaclass, line 1272: @jenshaase is it possible to call a static method from here? I would like to do something like:
```
printer (php/-> Printer::nonReadable())
```
But I don't know if that's even possible. That's why I let the constructor public... but I would say, ideally it should be private, so you can create a Printer only by one of its two named constructors.

Anyway, independently of the proposal and design itself, would it possible?

jenshaase: `(php/:: Printer (nonReadable))`

jenshaase:
Need more time to review this. But the color options should not be moved to the `Printer`. A printer is a concept that you can find in almost every Lisp. I prints the Lisp data. There is no concept of color for this. The colors are just for a better user experience when it comes to error reporting.

Chemaclass:
mmm ok, I think I got what you're saying. I will then move the color outside the Printer but leave the idea of creating the Printer by using named constructors.

jenshaase:
Merged it.

PR #51, opened 2020-06-15
Cleaning reader and compiler

Chemaclass:
# Description

I refactored some classes related to the Reader and Compiler without altering its functionality.

# Changes

- Declared classes as [strict](https://github.com/Chemaclass/php-best-practices/blob/master/technical-skills/strict-types.md) and [final](https://medium.com/@chemaclass/final-classes-in-php-9174e3e2747e).
- Removing unnecessary `else` statements due to return early.
- No logic was altered.
- `Quasiquote`
  - I rearranged the class methods: moving the private after the public one.

### DIFF
--- a/src/php/Reader.php
+++ b/src/php/Reader.php
@@ -79,7 +79,7 @@ final class Reader
     }
 
     /**
-     * @return Phel|null|string|scalar
+     * @return Phel|null|scalar
      */
     public function readExpression(Generator $tokenStream)
     {
--- a/src/php/Reader.php
+++ b/src/php/Reader.php
@@ -111,7 +111,7 @@ final class Reader
                     return $this->readList($tokenStream, Token::T_CLOSE_PARENTHESIS);
 
                 case Token::T_OPEN_BRACKET:
-                    return $this->readList($tokenStream, Token::T_CLOSE_BRACKET, true);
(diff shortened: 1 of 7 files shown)


src/php/Reader.php
```diff
@@ -88,7 +79,7 @@ private function readWhitespace(Generator $tokenStream): void
     }
 
     /**
-     * @return Phel|null|scalar
+     * @return Phel|null|string|scalar
```
jenshaase, line 82: `string` is unnecessary, because it is a `scalar` :)
Or do I miss something?

See https://www.php.net/manual/en/function.is-scalar.php

Chemaclass: You're right. This is a separate "problem".
PHPDoc doesn't recognize scalar as type for `string|bool|int|float`, in fact:
* some people already tried to create some special `csfixer` for this https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/3341 
* and even adding `scalar` Pseudo-type to the php-src: https://github.com/php/php-src/pull/2987

both of them declined.

For this reason, I'm a bit reluctant to use `Phel|null|scalar` because it's mixing too many types at once. But, as I said, that's another epic. I actually have an idea about how could we improve that, but not in this PR ;)

I will revert this line leaving how it was before.

jenshaase:
It's merged now.

PR #53, opened 2020-06-16
Refactoring Analyzer

Chemaclass:
# Description

I refactor the `Analyzer` class in order to encapsulate the private functions that were in that class before into classes with their own responsibility. The main purpose was to get rid of that "God-class" :)

# Changes

No logic was altered.

I refactored the `Analyzer` class using the Extract Class refactoring technique:

```php
// Analyzer::analyze($x, ?NodeEnvironment $env = null): Node 

if ($this->isLiteral($x)) {
    return (new AnalyzeLiteral())($x, $env);
}
if ($x instanceof Symbol) {
    return (new AnalyzeSymbol($this))($x, $env);
}
if ($x instanceof Tuple && $x->isUsingBracket()) {
    return (new AnalyzeBracketTuple($this))($x, $env);
}
if ($x instanceof PhelArray) {
    return (new AnalyzeArray($this))($x, $env);
}
if ($x instanceof Table) {
    return (new AnalyzeTable($this))($x, $env);
}
if ($x instanceof Tuple) {
    return (new AnalyzeTuple($this))($x, $env);
}
```

For simplicity, I group all of these classes inside a module/directory: "src/php/Analyzer". All of them have one public method `__invoke()` and they behave as identically the same as before. They receive the same arguments and they do the same.
Now, it's easier to read and understand the logic behind the Analyzer because all of these classes have a unique responsibility.

A follow-up PR would be split (using this very same idea) the class `AnalyzeTuple` (1200~lines). But I think this PR is big enough to be reviewed and double-check if this is going in a good direction 🛰️


src/php/Analyzer/AnalyzeLiteral.php
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Phel\Analyzer;
+
+use Phel\Ast\LiteralNode;
+use Phel\Lang\Phel;
+use Phel\NodeEnvironment;
+
+final class AnalyzeLiteral
+{
+    public function __invoke($x, NodeEnvironment $env): LiteralNode
```
jenshaase, line 13: Maybe we can add a DocComment for `$x` (`Phel|null|scalar`)?

src/php/Analyzer/AnalyzeArray.php
```diff
@@ -0,0 +1,32 @@
...
+use Phel\Analyzer;
+use Phel\Ast\ArrayNode;
+use Phel\Lang\PhelArray;
+use Phel\NodeEnvironment;
+
+final class AnalyzeArray
+{
+    private Analyzer $analyzer;
+
+    public function __construct(Analyzer $analyzer)
```
jenshaase, line 16: You can also make a `BaseAnalyzer` class that defines the constructor and the `$analyzer` property. That would mean less duplicated code.

src/php/Analyzer/AnalyzeLiteral.php
Chemaclass: Sure, no problem.

src/php/Analyzer/AnalyzeArray.php
Chemaclass: This "duplicated code" isn't really hurting at all, in my opinion. Actually, it makes every class independent from each other. They can be constructed in different ways. Look for example `AnalyzeLiteral`, it doesn't need the `Analyzer` at all.

jenshaase:
Thanks! I merged it!

JesusValera:
Wow, awesome PR @Chemaclass !! :tada:

PR #54, opened 2020-06-16
Implement psalm suggestions

JesusValera:
# Description

I ran the following commands and applied the suggestions when it was possible.
- `vendor/bin/psalm --alter --issues=MissingParamType --dry-run`
- `vendor/bin/psalm --show-info=true` commands

# Changes

- Declare return types explicitly & the type in the properties.
- Remove some innecesary documentation.
- In `NodeEnvironment` I removed an `if` conditional in the constructor in favour of null coalescing operator (`??`).
- Improve the `Reader->readSymbol()` method with nested `ifs`.
- Change the `Readline->readLine()` to return a `string` or `null` instead of ´bool` (unique return type == psalm happy == me happy :D ).


src/php/Munge.php
```diff
@@ -58,6 +58,7 @@ public static function decodeNs(string $s): string
         return self::encodeWithMap($s, array_flip(self::$mungeNsMapping));
     }
 
+    /** @param (array-key|mixed)[] $mapping */
```
jenshaase, line 61: This is actually a `array<string, string>`

src/php/Munge.php
```diff
@@ -58,6 +58,7 @@ public static function decodeNs(string $s): string
...
     private static function encodeWithMap(string $s, $mapping): string
```
jenshaase, line 62: Why not add `array $mapping`?

Chemaclass: This syntax is autogenerated by psalm when you run the auto fixer (from psalm suggestions). Anyway, I would also prefer to use `array` and/or even better the array types.

JesusValera: I agree, I thought about using something like: `array<string, string>` but I didn't know what key and value types were, so I prefer to use the psalm standard.
I will investigate later and replace by the good ones.
BTW, @jenshaase do you know what is the shape of the array?

PR #62, opened 2020-06-18
Implement Sets

paullaffitte:
# Description

This implements the set data structure described by #5. However I didn't implement the `join` function as it seems to be a duplicate of `union`.

However, I didn't update the documentation yet, I would like to get an approbation about this update first.

I tried to figure out how to parse a set with a custom notation (I was thinking about `@(1 2 3)` since this notation is not used yet and is close to another notation used in the language), but I couldn't understand what I should do exactly to get this result.

# Changes

- Added tests on Sets
- Added a Set php class derived from PhelArray
- Added Set phel constructor and functions
  - `(set 1 2 3 1 2 3)` create a set containing the values 1, 2 and 3 (only once since it's a set)
  - `union (set 1) (set 2) (set 3)` update the values of the first set to the given sets unions
  - `intersection (set 1) (set 2) (set 3)` update the values of the first set to the given sets intersections
  - `difference (set 1) (set 2) (set 3)` update the values of the first set to the given sets differences

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -690,28 +690,6 @@ arrays. Use (php/aunset ds key)" )))
     (php/aunset x key)
     x))
 
-(declare reduce)
-
-(defn union
-  "Union multiple sets into a new one"
-  [& sets]
-  (reduce concat (set) sets))
-
-(defn intersection
-  "Intersect multiple sets into a new one"
-  [& sets]
-  (reduce (fn [acc s] (php/-> acc (intersection s))) (union (first sets)) (rest sets)))
-
-(defn difference
-  "Difference between multiple sets into a new one"
-  [& sets]
-  (reduce (fn [acc s] (php/-> acc (difference s))) (union (first sets)) (rest sets)))
-
-(defn symmetricDifference
-  "Symmetric difference between multiple sets into a new one"
-  [& sets]
-  (reduce (fn [acc s] (union (difference acc s) (difference s acc))) (union (first sets)) (rest sets)))
-
 # --------
 # For loop
 # --------
@@ -1080,6 +1058,32 @@ collection in map"))
           (if (>= (count b) n)
             (recur b)
             res))))))
+
+
+# -------------
+# Set operation
+# -------------
+
+(defn union
+  "Union multiple sets into a new one"
+  [& sets]
+  (reduce concat (set) sets))
+
+(defn intersection
+  "Intersect multiple sets into a new one"
+  [set & sets]
+  (reduce |(php/-> $1 (intersection $2)) set sets))
+
+(defn difference
+  "Difference between multiple sets into a new one"
+  [set & sets]
+  (reduce |(php/-> $1 (difference $2)) set sets))
+
+(defn symmetric-difference
+  "Symmetric difference between multiple sets into a new one"
+  [set & sets]
+  (reduce |(union (difference $1 $2) (difference $2 $1)) set sets))
+
 
 # ------------------
 # Function operation
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -710,7 +710,7 @@ arrays. Use (php/aunset ds key)" )))
 (defn symmetricDifference
   "Symmetric difference between multiple sets into a new one"
   [& sets]
-  (reduce (fn [acc s] (php/-> acc (symmetricDifference s))) (union (first sets)) (rest sets)))
+  (reduce (fn [acc s] (union (difference acc s) (difference s acc))) (union (first sets)) (rest sets)))
 
 # --------
 # For loop
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -690,33 +690,27 @@ arrays. Use (php/aunset ds key)" )))
     (php/aunset x key)
     x))
 
+(declare reduce)
+
 (defn union
-  "Union multiple sets into a new set"
+  "Union multiple sets into a new one"
   [& sets]
-  (let (s (set))
-    (apply concat s sets)))
+  (reduce concat (set) sets))
 
 (defn intersection
-  "Intersect multiple sets into one"
-  [s & sets]
-  (foreach [se sets]
-    (php/-> s (intersection se)))
-  s)
-
-(defn difference-into
-  "Difference between multiple sets into one"
-  [s & sets]
-  (foreach [se sets]
-    (php/-> s (difference se)))
-  s)
+  "Intersect multiple sets into a new one"
+  [& sets]
+  (reduce (fn [acc s] (php/-> acc (intersection s))) (union (first sets)) (rest sets)))
 
 (defn difference
-  "Difference between multiple sets"
+  "Difference between multiple sets into a new one"
   [& sets]
-  (if (= (count sets) 0)
-    (set)
-    (apply difference-into (first sets) (rest sets))))
+  (reduce (fn [acc s] (php/-> acc (difference s))) (union (first sets)) (rest sets)))
 
+(defn symmetricDifference
+  "Symmetric difference between multiple sets into a new one"
+  [& sets]
+  (reduce (fn [acc s] (php/-> acc (symmetricDifference s))) (union (first sets)) (rest sets)))
 
 # --------
 # For loop
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -6,6 +6,7 @@
   (:use Phel\Lang\IContactable)
   (:use Phel\Lang\PhelArray)
   (:use Phel\Lang\Table)
+  (:use Phel\Lang\Set)
   (:use Phel\Lang\Nil)
   (:use Phel\Lang\ICons)
   (:use Phel\Lang\ISlice)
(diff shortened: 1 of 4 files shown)


src/php/Lang/Set.php
```diff
@@ -0,0 +1,86 @@
...
+     * @return Set
+     */
+    public static function create(...$values): Set
+    {
+        return new Set($values);
+    }
+
+    public function push($x): IPush
+    {
+        if (!in_array($x, $this->data)) {
```
jenshaase, line 39: This is not a good idea, because it is slow. `in_array` has the complexity `O(n)`. But checking if a value is in a set must be `O(1)`. 

You have to do this the following way.
1. Compute the hash of the value (see `Table` class)
2. Store the hash as key of the PHP Array and the value as the value of the PHP Array.

The push method looks like this:

```
public function push($x): IPush
{
    $hash = $this->computeHash($x);
    $this->data[$hash] = $x; // Don't need to check if $x is already there, just override.

   return $this;
}
```

Afterwards, you probably can simplify the methods for intersection, union and diff.

paullaffitte:
Thanks for the feedback, I will work on it again as soon as I have some free time.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,86 @@
...
+
+namespace Phel\Lang;
+
+use ArrayAccess;
+use Countable;
+use InvalidArgumentException;
+use Iterator;
+use Phel\Printer;
+
+class Set extends PhelArray implements ArrayAccess, Countable, Iterator, ICons, ISlice, ICdr, IRest, IPop, IRemove, IPush, IConcat
```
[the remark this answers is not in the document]
paullaffitte: Ok, I will make the required change to get to this. Btw, maybe we could use traits in order to reuse some common functions. For instance, the `hash` function is implemented in the same way all over the project. It could be nice to factorize the code, and traits are probably a better approach than inheritance. I will open an issue about it to discuss about it.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return new Set($values);
+    }
+
+    public function isTruthy(): bool
+    {
+        return true;
+    }
+
+    public function hash(): string
```
jenshaase, line 46: This method is unnecessary.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public function rewind()
+    {
+        reset($this->data);
+    }
+
+    public function cons($x): ICons
+    {
+        array_unshift($this->data, $x);
```
jenshaase, line 83: This method does not check if `$x` is already in the set. You can just call `push` here, since set do not have ordering.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public function union(Set $set): IConcat
+    {
+        return $this->concat($set->toPhpArray());
+    }
+
+    public function intersection(Set $set): IConcat
+    {
+        return $this->applySet($set, 'array_intersect');
```
jenshaase, line 111: I think, we must call `array_intersect_key` here.

Why not just:
```
$this->data = array_intersect_key($this->data, $set->toPhpArray());
```
This is much simpler to read. Also you don't need the to wrap the method into a string.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public function difference(Set $set): IConcat
+    {
+        $rightDiff = new Set($set->toPhpArray());
```
jenshaase, line 116: I don't really understand what you are doing here.

I would just go for:
```
$this->data = array_diff_key($this->data, $set->toPhpArray());
```

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+
+        return $this;
+    }
+
+    public function union(Set $set): IConcat
+    {
+        return $this->concat($set->toPhpArray());
+    }
+
+    public function intersection(Set $set): IConcat
```
jenshaase, line 109: Return type must be `Set`

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+    {
+        return $this->applySet($set, 'array_intersect');
+    }
+
+    public function difference(Set $set): IConcat
```
jenshaase, line 114: Return type must be `Set`

src/phel/core.phel
```diff
@@ -677,6 +690,32 @@ arrays. Use (php/aunset ds key)" )))
     (php/aunset x key)
     x))
 
+(defn union
+  "Union multiple sets"
+  [& sets]
+  (apply concat sets))
```
jenshaase, line 696: This method fails, if we just call `(union)`.

src/phel/core.phel
```diff
@@ -677,6 +690,32 @@ arrays. Use (php/aunset ds key)" )))
...
+
+(defn intersection
+  "Intersect multiple sets into one"
+  [s & sets]
+  (do
```
jenshaase, line 701: `do` is not required any more.

src/phel/core.phel
```diff
@@ -677,6 +690,32 @@ arrays. Use (php/aunset ds key)" )))
...
+    (foreach [se sets]
+      (php/-> s (intersection se)))
+    s))
+
+(defn difference-into
+  "Difference between multiple sets into one"
+  [s & sets]
+  (do
```
jenshaase, line 709: `do` is not required any more.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,163 @@
...
+
+namespace Phel\Lang;
+
+use ArrayAccess;
+use Countable;
+use InvalidArgumentException;
+use Iterator;
+use Phel\Printer;
+
+class Set extends Phel implements Countable, Iterator, ICons, IPush, IConcat
```
jenshaase, line 11: After the last commit you need to extends from `AbstractType` instead of `Phel`

src/php/Lang/Set.php
paullaffitte: It seems required since I extends AbstractType which contains the abstract method `hash`. I gave it a try and unsurprisingly it yield the fatal error `Class Phel\Lang\Set contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Phel\Lang\AbstractType::hash)`.

src/php/Lang/Set.php
paullaffitte: Indeed, `array_diff_key` is definitely a better approach, I updated my code to use this function. But the biggest weirdness in my code comes from the way PHP implements `array_diff`.

Take this code:
```php
$foo = [1, 2];
$bar = [2, 3];
$foobar = array_diff($foo, $bar);
```
Here `$foobar` will not equals to `[1, 3]` as expected, but to `[1]`. That's why I have to execute  array_diff two times, to get the diff from both arrays, and then merge them. The same is true for `array_diff_key`.

But if you know a better solution, I would be happy to know it, because it's actually not very pretty neither readable.

src/phel/core.phel
paullaffitte: I believe this also means that `union` should always create a new set instead of updating the first one? And that this rule should apply to `intersection` and `difference` too?

src/php/Lang/Set.php
```diff
@@ -0,0 +1,161 @@
...
+        return $this;
+    }
+
+    public function intersection(Set $set): Set
+    {
+        $this->data = array_intersect_key($this->data, $set->toPhpArray());
+        return $this;
+    }
+
+    public function difference(Set $set): Set
```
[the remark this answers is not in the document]
paullaffitte: Oh I see, so it was just a misunderstanding about how sets works. I will keep this function as it's still useful but rename it as symmetricDifference then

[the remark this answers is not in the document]
jenshaase: Please only implement the Set Difference in PHP. You can then implement the Symmetric Difference in Phel.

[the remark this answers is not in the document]
paullaffitte: Oh sorry, I resolved the conversation before refreshing the page and already pushed my code. I will update it right now.

src/php/Lang/Set.php
```diff
@@ -0,0 +1,147 @@
...
+    {
+        $difference = array_diff_key($this->data, $set->toPhpArray());
+        $this->data = [];
+        $this->concat($difference);
```
jenshaase, line 109: Why do we need to call concat here again?

I would just do
```
$this->data = $difference;
```

src/phel/core.phel
```diff
@@ -677,6 +690,27 @@ arrays. Use (php/aunset ds key)" )))
     (php/aunset x key)
     x))
 
+(declare reduce)
```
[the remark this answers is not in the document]
paullaffitte: Is there a reason to avoid reduce or is it just a personal preference? Before I used 2 functions and foreach, but using reduce helped me making this function way shorter and more readable.

src/php/Lang/Set.php
paullaffitte: I guess it's just an artifact from my refactor, there is actually no reason to use it.

src/phel/core.phel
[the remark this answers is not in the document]
jenshaase: Sorry, maybe I didn't made it clear. Using `reduce` is a good idea. I just wanted to move the set functions after the reduce function has been defined. Then you don't need the "declare" call. Does it make sense now?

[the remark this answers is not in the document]
paullaffitte: Oh ok, perfect then.

PR #65, opened 2020-06-18
Upgrade tests

paullaffitte:
# Description

This feature has the goal to make the tests easier to debug.

# Changes

- Extraction of the library part of the testing code (which was duplicated across core and http) in a separate file
- Print expected and got values on error
- Centralize test suites in an `index.phel` file (however there is an issue about requiring modules, so the http module cannot be required, thus cannot neither be tested)


jenshaase:
I think I resolve the bug in #66 

But there are some other problems here. You cannot use `def` inside of `defn`. `def` can only be defined globally. The compiler should throw an error. I made an issue for that: #68

Refactoring the test was a good idea. But I would go a step further. For the first release of Phel I would like to have a standard test library integrated into Phel. I already have some ideas for that.

Therefore, I think it is better to reject this pull request and refactor the test code when there is a test library.

What do you think?

paullaffitte:
Yeah, I thought about making a test library too. I think it would be a very good idea. So it would be maintained in a separate repository right?

Also I didn't push another change because again I couldn't make it work with the http test module, but I found a way pretty elegant to print on test failure. Currently I only print "Foobar - Expected: foo, Got: bar", but actually it's wrong since the test function could be something else than `=`. So by replacing assert with a macro calling the original assert method, I could make it print something like "Foobar - Expected: rhs testFunction lhs". So for `(assert = "foo" "bar" "Foobar")`, it would output "Foobar - Expected: bar = foo". Which is way better and can work with any test function.

The code of the macro simply being like this.
```phel
 # assertStr being the original assert function with a parameter for the function name added
(defmacro assert
  [test left right label]
  `(assertStr ,test ,(print-str test) ,left ,right ,label))
```

I would love to work on the standard test library, so let me know when you want to do it.

jenshaase:
I will close this. We can discuss in more detail here: https://github.com/jenshaase/phel-lang/issues/78

PR #74, opened 2020-06-22
Refactor TupleSymbols

Chemaclass:
# Description

I refactor and clean the `Analyzer/TupleSymbol/*.php`.

# Changes

- improving naming (for example: changing `$x` to `$tuple` or `$nodeEnvironment` to `$env`)
- apply extract method refactoring for almost every `else` I saw
- use named constructor (`AnalyzerException::withLocation()`) to help the creation of an `AnalyzerException`
- extract chunks of logic in private methods to keep the level of abstraction easy-peasy to understand
- no functionality was changed

PR #75, opened 2020-06-22
Some random fixes

JesusValera:
# Changes

- Declare the property type in a few method calls.
- Make the `Exceptions` classes final and strict.
- Update the class `PhelCodeException` to make it `abstract`.
- Replace some `new` instances by `self`.
- Small improvs.

### DIFF
--- a/src/php/Lang/Struct.php
+++ b/src/php/Lang/Struct.php
@@ -20,7 +20,7 @@ abstract class Struct extends Table
     {
         if (!in_array($offset, $this->getAllowedKeys(), false)) {
             $keyName = Printer::nonReadable()->print($offset);
-            $structName = static::class;
+            $structName = get_class($this);
             throw new InvalidArgumentException(
                 "This key '$keyName' is not allowed for struct $structName"
             );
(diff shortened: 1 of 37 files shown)


src/php/Lang/Struct.php
```diff
@@ -19,9 +18,9 @@ abstract class Struct extends Table
 
     protected function validateOffset($offset)
     {
-        if (!in_array($offset, $this->getAllowedKeys())) {
+        if (!in_array($offset, $this->getAllowedKeys(), false)) {
             $keyName = Printer::nonReadable()->print($offset);
-            $structName = get_class($this);
+            $structName = self::class;
```
jenshaase, line 23: Shouldn't this be `static` (Late static binding)?

JesusValera: Yes, you are right -> https://stackoverflow.com/a/4719222/4988034

JesusValera: Done. :)

JesusValera:
> The build (psalm) failed after the last merge. Can you fix it?

Yes, I am investigating.

By any reason, `psalm` doesn't like the `static::class;`.
At the moment, I am going to put back the `get_class($this)`.

JesusValera:
Okk, the problem was not the `static::class` but the PR #74.
Let me fix it: :)

jenshaase:
Thanks. It's merged now.

PR #76, opened 2020-06-23
Use traits to implement common function among Phel types

### LINKED ISSUE #71: Use traits to factorize Phel classes implementations
As discussed here https://github.com/jenshaase/phel-lang/pull/62#discussion_r442647840, we could factorize the implementation of Phel classes by using traits, since inheritance seems to not be the right way to do it. There is some function reused over the code like `hash`, `count`, `current`, `next`, `key`, `valid`, `rewind` or `offsetHash` that share the same implementation over `PhelArray`, `Table`, `Tuple` and soon `Set`.

Factorize this code would help to improve the maintainability of those classes and would make them way shorter, keeping only the code specific to the feature inside them.

paullaffitte:
# Description

Closes #71 by refactoring classes inheriting from `AbstractType` in order to factorize the code with traits, and to make it more concise, so it will be also more readable and maintainable. Since this is only refactoring, it should keep the exact same behavior that Phel previously had.

# Changes

- Create traits
  - `HashableTrait`
  - `PrintableTrait`
  - `CountableTrait`
  - `IterableTrait`
  - `ComparableTrait`
- Use those traits in Phel Types
  - `Keyword`
  - `PhelArray`
  - `Symbol`
  - `Table`
  - `Tuple`

### DIFF
--- a/src/php/Lang/ComparableTrait.php
+++ b/src/php/Lang/ComparableTrait.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Phel\Lang;
+
+trait ComparableTrait
+{
+    public function equals($other): bool
+    {
+        return $this == $other;
+    }
+}
--- a/src/php/Lang/CountableTrait.php
+++ b/src/php/Lang/CountableTrait.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Phel\Lang;
+
+trait CountableTrait
+{
+    public function count(): int
+    {
+        return count($this->data);
+    }
+}
--- a/src/php/Lang/HashableTrait.php
+++ b/src/php/Lang/HashableTrait.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Phel\Lang;
+
+trait HashableTrait
+{
+    public function hash(): string
+    {
+        return spl_object_hash($this);
+    }
+}
--- a/src/php/Lang/IterableTrait.php
+++ b/src/php/Lang/IterableTrait.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace Phel\Lang;
+
+trait IterableTrait
+{
+    public function current()
+    {
+        return current($this->data);
+    }
+
+    public function key()
+    {
+        return key($this->data);
+    }
+
+    public function next(): void
+    {
+        next($this->data);
+    }
+
+    public function rewind(): void
+    {
+        reset($this->data);
+    }
+
+    public function valid(): bool
+    {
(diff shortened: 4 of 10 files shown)


jenshaase:
@Chemaclass What is your opinion on this? I think it will reduce duplicated code, but will make that individual classes harder to read. What I don't like is that there are traits, that work on `$this->data`.  `$this->data` is not defined in the traits but in the classes with semantically different meanings.

Chemaclass:
I like the intention, but I see a problem here: the traits rely on a property member (as @jenshaase mentioned) such a 
`$this->data` and it's something that doesn't belong to the trait itself but the class which is using it. 

To "fix" this particular issue you could define an abstract method in the trait, so everyone using that trait must implement that method. I'm talking about:
```php
trait CountableTrait
{
    public abstract function data(): array;

    public function count(): int
    {
        return count($this->data());
    }
}

final class PhelArray extends AbstractType implements ...
{
    use CountableTrait;

    public function data(): array 
    {
        return $this->data;
    }
    // ...
}
```

That might be solution in order to ensure the dependency-variables used in a trait from other classes that use that particular trait. Saying that I don't see a huge benefit (at least in this context) of having these `ComparableTrait`, `CountableTrait`, `HashableTrait` & `PrintableTrait` for the reason that it will make that individual classes harder to read for just one method.

On the other hand, `PrintableTrait` has a collection of methods, which all depend on that `$this->data` property. Here it might be interesting to have that abstract method with the trait, for example, but I'm not sure if the "loss of readability" worth that trait extraction or not. 

TL;DR: I'm with @jenshaase, I think it might not only make that individual classes will be harder to read but also the 
`$this->data` can ends to be more like a hack" than something that we want to have.

paullaffitte:
I see two issues in implementing an abstract function `data` in each traits requiring it. First, it will duplicate code again. Second, and this is where there is a catch, `next` and `reset` (abstract methods from `Iterable`) require a variable, but using this technique, we will get instead a reference, so it will just don't work.

Maybe traits are just not the right way to do this. But I don't have any better idea about it for now. So maybe we just want to reject the pull request and close the related issue.

Chemaclass:
Yeah, I agree with you there @paullaffitte. I feel that these traits don't solve the "problem", actually, I see it more like they come to solve a problem introducing another one. I also don't have a better solution, for now, so I would go for the rejection and maybe we'll see another potential improvement in the upcoming days/weeks :)

jenshaase:
Thanks for you open discussion on this topic. I will close this now.

PR #82, opened 2020-06-24
Sets documentation

### LINKED ISSUE #77: Docs: Set Datastructure
The set data structure should be added to https://phel-lang.org/documentation/array-table-and-struct/

paullaffitte:
# Description

Added documentation on sets. Closes #77 

# Changes

- Fixed some typos in the "Array, Table and Struct" documentation.
- Added sets documentation in "Array, Table and Struct" and renamed it "Array, Table, Struct and Set".
- Updated sets' tests in order to reflect that sets are unsorted according to the new documentation.

### DIFF
--- a/doc/content/documentation/array-table-and-struct.md
+++ b/doc/content/documentation/array-table-and-struct.md
@@ -3,7 +3,7 @@ title = "Array, Table, Struct and Set"
 weight = 9
 +++
 
-Phel has two mutable datastructures. An array is an indexed sequential datastructure and a table is an associative datastructure. While PHP has one single datastructure to represents arrays and hash tables, Phel splits them in two individual datastructures.
+Phel has three mutable datastructures. An array is an indexed sequential datastructure, a table is an associative datastructure and a set is a unsorted sequential data structure that cannot contain twice the same value. While PHP has one single datastructure to represents arrays and hash tables, Phel splits them in two individual datastructures.
 
 An array can be created using the `@[` reader macro or the `array` function.
 
@@ -19,7 +19,7 @@ A table can be created using the `@{` reader macro or the `table` function.
 (table :key1 "value1" :key2 "value2")
 ```
 
-A set is an unsorted sequential data structure that cannot contain twice the same value. It can be created using the `set` function.
+A set is can be created using the `set` function.
 
 ```phel
 (set 1 2 3)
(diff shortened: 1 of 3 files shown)


doc/content/documentation/array-table-and-struct.md
```diff
@@ -107,6 +123,48 @@ An array can also be used as a stack. Therefore, the `push`, `peek` and `pop` fu
   (pop arr)) # Evaluates to 1, arr is empty @[]
 ```
 
+## Set operations
+
+A set provide very specific functions described in set theory. Such as `union`, `intersection`, `difference` and `symmetric-difference`.
```
jenshaase, line 128: A set provide**s** very specific functions **that are** described in the set theory. **These are** `union`,...

doc/content/documentation/array-table-and-struct.md
```diff
@@ -1,9 +1,9 @@
 +++
-title = "Array, Table and Struct"
+title = "Array, Table, Struct and Set"
 weight = 9
 +++
 
-Phel has two mutable datastructures. An array is an indexed sequential datastructure and a table is an associative datastructure. While PHP has one single datastructure to represents arrays and hash tables, Phel splits them in two individual datastructures.
+Phel has three mutable datastructures. An array is an indexed sequential datastructure and a table is an associative datastructure. While PHP has one single datastructure to represents arrays and hash tables, Phel splits them in two individual datastructures.
```
jenshaase, line 6: You should give a summary of all three data structures here. Currently you say that there are three, but you list only two :)

paullaffitte: Oops, yes. But actually I think I will stick with "two", since `set` doesn't seems to be mutable in any way in its current implementation. Indeed, set operations only create new sets, but doesn't update them.

doc/content/documentation/array-table-and-struct.md
```diff
@@ -84,19 +101,24 @@ To remove multiple values from arrays, the `remove` and `slice` functions can be
...
 
 (count @{:a 1 :b 2}) # Evaluates to 2
 (count @{}) # Evaluates to 0
+
+(count (set 1 2 3)) # Evaluates to 3
+(count (set)) # Evaluates to 0
 ```
 
-## Array as a Stack
+## Array/Set as a Stack
```
jenshaase, line 121: It may work that sets can be used as stack, but sets don't have a strict order. Therefore, from a functional view point they can not be used as stacks, because stacks required ordered data structures.

PR #85, opened 2020-06-26
Cleaning analyzer types

Chemaclass:
# Description

Introduced constants for every Symbol name (let, do, loop, etc) and clean some analyze symbol classes.

# Changes

- Included `Symbol::NAME_*` constants
- Used that constants instead of raw strings where does it makes sense
- Minor little cleaning
- No functionality was modified


src/php/Lang/Symbol.php
```diff
@@ -8,6 +8,28 @@
...
+    public const NAME_IF = 'if';
+    public const NAME_APPLY = 'apply';
+    public const NAME_LET = 'let';
+    public const NAME_PHP_NEW = 'php/new';
+    public const NAME_PHP_OBJECT_CALL = 'php/->';
+    public const NAME_PHP_OBJECT_STATIC_CALL = 'php/::';
+    public const NAME_PHP_ARRAY_GET = 'php/aget';
+    public const NAME_PHP_ARRAY_SET = 'php/aset';
+    public const NAME_PHP_ARRAY_PUSH = 'php/apush';
+    public const NAME_PHP_ARRAY_UNSET = 'php/aunset';
```
Chemaclass, line 25: @jenshaase I was thinking between `NAME_PHP_AGET` and `NAME_PHP_ARRAY_GET`. I think the first one is closer to the real symbol itself, but the second one is more expressive, but what do you think?

src/php/Reader.php
```diff
@@ -129,7 +129,7 @@ public function readExpression(Generator $tokenStream)
                     throw $this->buildReaderException('Unterminated list');
 
                 case Token::T_QUOTE:
-                    return $this->readWrap($tokenStream, 'quote');
+                    return $this->readWrap($tokenStream, Symbol::NAME_QUOTE);
```
Chemaclass, line 132: @jenshaase  I'm not sure about this one 🤔 because what's about ''unquote'? Should we treat it as a symbol as well?

jenshaase: Yes, `unquote`, `quote` and `unquote-splicing` are all Symbol names. They are mostly used in the Quasiqoute class.

src/php/Lang/Symbol.php
jenshaase: `NAME_PHP_ARRAY_GET` is fine for me

src/php/Reader.php
Chemaclass: Ok, then I will create those constants as well :)

PR #86, opened 2020-06-27
fix set first and second

paullaffitte:
# Description

`(first (set))` should return `nil` but return `false` instead. Also `second` cannot be used on a set since it uses `php/aget`.

# Changes

- Added set tests
- Fixed `first` on empty set
- Updating `second` to use `first` and `next` to be compatible with sets


jenshaase:
Thanks!

PR #87, opened 2020-06-27
Add test QuoteSymbol class

JesusValera:
# Description

The idea is to unit test all symbols but not sure 100% if this is the right approach.
So, I pick the simplest one.

Take a look @jenshaase and give me some feedback. :P

---
I checked you can pass as the first argument to `Tuple` anything and it will work.
In fact, in the code, we are only getting the `$tuple[1]`.

Not sure how much "secure" it is... :thinking:

### DIFF
--- a/src/php/Analyzer/TupleSymbol/QuoteSymbol.php
+++ b/src/php/Analyzer/TupleSymbol/QuoteSymbol.php
@@ -14,7 +14,7 @@ final class QuoteSymbol
 {
     public function __invoke(Tuple $tuple, NodeEnvironment $env): QuoteNode
     {
-        if (!($tuple[0] instanceof Symbol && (string) $tuple[0] === Symbol::NAME_QUOTE)) {
+        if (!($tuple[0] instanceof Symbol && $tuple[0]->getName() === Symbol::NAME_QUOTE)) {
             throw AnalyzerException::withLocation("This is not a 'quote.", $tuple);
         }
 
--- a/src/php/Analyzer/TupleSymbol/QuoteSymbol.php
+++ b/src/php/Analyzer/TupleSymbol/QuoteSymbol.php
@@ -14,7 +14,7 @@ final class QuoteSymbol
 {
     public function __invoke(Tuple $tuple, NodeEnvironment $env): QuoteNode
     {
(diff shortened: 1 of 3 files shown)


tests/php/Analyzer/TupleSymbol/QuoteSymbolTest.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Phel\Exceptions\PhelCodeException;
+use Phel\Lang\Symbol;
+use Phel\Lang\Tuple;
+use Phel\NodeEnvironment;
+use PHPUnit\Framework\TestCase;
+
+final class QuoteSymbolTest extends TestCase
+{
+    /** @test */
```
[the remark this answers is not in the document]
JesusValera: Changed the annotation by the method name prefix.

I added the guard-clause condition in `QuoteTuple` class, in order to ensure the current object is, in fact, a `QuoteSymbol` (as it was already done in `LoopSymbol`).

From my point of view, the `if` conditional has a micro computational cost but the class is much more secure.
I think it worths to avoid possible conflicts in the future. And with PHP 8 and JIT feature will be much faster!

I add a one more test to check that.

src/php/Analyzer/TupleSymbol/QuoteSymbol.php
```diff
@@ -6,15 +6,20 @@
...
 use Phel\Exceptions\AnalyzerException;
+use Phel\Lang\Symbol;
 use Phel\Lang\Tuple;
 use Phel\NodeEnvironment;
 
 final class QuoteSymbol
 {
     public function __invoke(Tuple $tuple, NodeEnvironment $env): QuoteNode
     {
+        if (!($tuple[0] instanceof Symbol && $tuple[0] == Symbol::NAME_QUOTE)) {
```
jenshaase, line 17: It should be `$tuple[0]->getName() === Symbol::NAME_QUOTE`

JesusValera: Uups. Of course :)

jenshaase: Use the `getName()` function instead of string casting.

JesusValera: Apply suggestion and update the names of the tests. I think now is much better!

PR #89, opened 2020-06-28
Add tests to AnalyzeTuple class

JesusValera:
# Description

- Create `AnalyzeTupleTest` class.
    - The purpose of this test is to check if the analyze factory works and depending on the input, it returns the correct class.

**PS:** I was not able to test the `default` switch case.

### DIFF
--- a/tests/php/Analyzer/AnalyzeTupleTest.php
+++ b/tests/php/Analyzer/AnalyzeTupleTest.php
@@ -46,7 +46,7 @@ final class AnalyzeTupleTest extends TestCase
         self::assertInstanceOf(DefNode::class, $this->tupleAnalyzer->__invoke($tuple, NodeEnvironment::empty()));
     }
 
-    public function testSymbolWithNameNS(): void
+    public function testSymbolWithNameNs(): void
     {
         $tuple = Tuple::create(Symbol::create(Symbol::NAME_NS), Symbol::create('def-ns'));
         self::assertInstanceOf(NsNode::class, $this->tupleAnalyzer->__invoke($tuple, NodeEnvironment::empty()));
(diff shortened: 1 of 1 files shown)


JesusValera:
> Looks good to me.
> 
> To test the default case you can do somethings like this:
> 
> ```
> Tuple::create(Tuple::create(Symbol::create(Symbol::NAME_FN), Tuple::create()))
> ```
> 
> this is equivalent to the following Phel code
> 
> ```
> ((fn []))
> ```
> 
> In PHP it would look like this:
> 
> ```
> (function() {})()
> ```

It's not working.
The code enter into the first conditional (L42) instead of the `default case`.

```php
if (!$tuple[0] instanceof Symbol) {
    return (new InvokeSymbol($this->analyzer))($tuple, $env);  // <- HERE
}
switch ($tuple[0]->getFullName()) {
    // . . .
    default:
        return (new InvokeSymbol($this->analyzer))($tuple, $env); // NOT HERE :)
}
```

tests/php/Analyzer/AnalyzeTupleTest.php
```diff
@@ -0,0 +1,192 @@
...
+        $this->tupleAnalyzer = new AnalyzeTuple(new Analyzer(new GlobalEnvironment()));
+    }
+
+    public function testSymbolWithNameDef(): void
+    {
+        $tuple = Tuple::create(Symbol::create(Symbol::NAME_DEF), Symbol::create('increment'), 'inc');
+        self::assertInstanceOf(DefNode::class, $this->tupleAnalyzer->__invoke($tuple, NodeEnvironment::empty()));
+    }
+
+    public function testSymbolWithNameNS(): void
```
JesusValera, line 49:
```suggestion
    public function testSymbolWithNameNs(): void
```

PR #94, opened 2020-07-11
Refactoring RecurSymbol and NodeEnvironment

Chemaclass:
# Description

I just clean up a little bit the `RecurSymbol` and `NodeEnvironment` classes

# Changes

- Refactor a little bit the `RecurSymbol` and `NodeEnvironment` classes
- Renamed `exprs` to `expressions`
- No functionality was altered

### DIFF
--- a/src/php/NodeEnvironment.php
+++ b/src/php/NodeEnvironment.php
@@ -31,7 +31,7 @@ final class NodeEnvironment
 
     /**
      * A list of RecurFrame
-     * @var RecurFrame[]
+     * @var array<RecurFrame|null>
      */
     private array $recurFrames;
 
@@ -44,7 +44,7 @@ final class NodeEnvironment
      * @param Symbol[] $locals A list of local symbols
      * @param string $context The current context (Expression, Statement or Return)
      * @param Symbol[] $shadowed A list of shadowed variables
-     * @param RecurFrame[] $recurFrames A list of RecurFrame
+     * @param array<RecurFrame|null> $recurFrames A list of RecurFrame
      * @param string|null $boundTo A variable this is bound to
      */
     public function __construct(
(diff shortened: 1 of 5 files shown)


src/php/NodeEnvironment.php
```diff
@@ -153,7 +153,7 @@ public function withAddedRecurFrame(RecurFrame $frame): NodeEnvironment
     public function withDisallowRecurFrame(): NodeEnvironment
     {
         $result = clone $this;
-        $result->recurFrames = array_merge($this->recurFrames, [null]);
+        $result->recurFrames = array_merge($this->recurFrames, [RecurFrame::withoutParams()]);
```
Chemaclass, line 156: @jenshaase I wanted to double-check with you this change. 
**I did this in order to avoid `null` values** inside `$this->recurFrames`, but I'm not 100% sure that this might break something, even though all tests are green.  As far as I saw, this is used in `RecurSymbol` class:
```php
$currentFrame = $env->getCurrentRecurFrame();

if (!$currentFrame) {
    throw AnalyzerException::withLocation("Can't call 'recur here", $tuple[0]);
}

if (count($tuple) - 1 !== count($currentFrame->getParams())) {
    throw AnalyzerException::withLocation("Wrong number of arguments for 'recur.", $tuple);
}
```
Could you bring some light here? :)

jenshaase: This should be `null`. `recurFrames` is a stack of frames. If `null` is on the stack it means that the `recur` function is not allowed at this point. A `RecurFrame` without Parameters means that `recur` is call without arguments (e.g. `(recur)`). It is best explained with some examples:

```
(loop []
  (recur)) # Recur is allowed inside a loop (no parameters, endless loop)
```

```
(loop []
  (do
    (println "test")
    (recur) # Recur is not allowed here (it is only allowed at the end)
    (println "abc"))) 
```

```
(loop []
  (do
    (println "test")
    (println "abc")
    (recur) # Recur is allowed here
)) 
```

Chemaclass: I understand this now, thanks! I will rollback that thing later and I will let you know :)

PR #97, opened 2020-07-18
Def inside of Def should not work

Chemaclass:
### Issue https://github.com/jenshaase/phel-lang/issues/68

# Description

Forbid def inside def

# Changes

- Added a flag (`$defAllowed`) inside the `NodeEnvironment` to keep track of the `DefSymbol` to know if it was already defined.

### DIFF
--- a/src/php/Analyzer/TupleSymbol/DefSymbol.php
+++ b/src/php/Analyzer/TupleSymbol/DefSymbol.php
@@ -8,6 +8,7 @@ use Phel\Analyzer\WithAnalyzer;
 use Phel\Ast\DefNode;
 use Phel\Ast\Node;
 use Phel\Exceptions\AnalyzerException;
+use Phel\Exceptions\PhelCodeException;
 use Phel\Lang\AbstractType;
 use Phel\Lang\Keyword;
 use Phel\Lang\Symbol;
@@ -21,24 +22,15 @@ final class DefSymbol
 
     private const POSSIBLE_TUPLE_SIZES = [3, 4];
 
+    /**
+     * @throws PhelCodeException
+     */
     public function __invoke(Tuple $tuple, NodeEnvironment $env): DefNode
     {
-        if (!$env->isDefAllowed()) {
-            throw AnalyzerException::withLocation("'def inside of a 'def is forbidden", $tuple);
-        }
-
-        $env->setDefAllowed(false);
-        $tupleSize = count($tuple);
-
-        if (!in_array($tupleSize, self::POSSIBLE_TUPLE_SIZES)) {
-            throw AnalyzerException::withLocation(
-                "Two or three arguments are required for 'def. Got " . count($tuple),
-                $tuple
-            );
-        }
+        $this->ensureDefIsAllowed($tuple, $env);
+        $this->verifySizeOfTuple($tuple);
 
         $nameSymbol = $tuple[1];
-
         if (!($nameSymbol instanceof Symbol)) {
             throw AnalyzerException::withLocation("First argument of 'def must be a Symbol.", $tuple);
         }
@@ -47,7 +39,7 @@ final class DefSymbol
             ->getGlobalEnvironment()
             ->getNs();
 
-        [$metaTable, $init] = $this->createMetaAndInit($tuple);
+        [$metaTable, $init] = $this->createMetaTableAndInit($tuple);
 
         $this->analyzer
             ->getGlobalEnvironment()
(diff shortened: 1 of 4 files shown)


src/php/NodeEnvironment.php
```diff
@@ -179,4 +184,16 @@ public function getBoundTo(): string
     {
         return $this->boundTo;
     }
+
+    public function isDefAllowed(): bool
+    {
+        return $this->defAllowed;
+    }
+
+    public function setDefAllowed(bool $defAllowed): self
```
jenshaase, line 193: NodeEnvironment should always be immutable. So instead of a setter the method should be called `withDefAllowed` and return a new instance of NodeEnvironment (see other methods in this class).

I made it immutable because it is much easier to reason about it.

tests/php/IntegrationTest.php
```diff
@@ -60,7 +60,7 @@ protected function doIntegrationTest(string $filename, string $phelCode, string
             );
         }
         $compiledCode = trim(implode('', $compiledCode));
-        $this->assertEquals($generatedCode, $compiledCode, 'in ' . $filename);
+        self::assertEquals($generatedCode, $compiledCode, 'in ' . $filename);
```
jenshaase, line 63: Why is this static now?

Chemaclass: Because all asssertion (in PHPUnit at least) are static functions:
```php
public static function assertEquals($expected, $actual, string $message = ''): void
```
I thought it might be interesting to agree on one style of writing them. For this reason, I wrote Sebastian asking for this particular topic, what does he recommend: https://twitter.com/Chemaclass/status/1284800657027936256

From my point of view, if a function is static, I would prefer to call it using either `self::` of `static::`, this way it will visually clear that this function is static from the calling context.

src/php/NodeEnvironment.php
Chemaclass: Absolutely right, good catch! I will change it! :)

tests/php/IntegrationTest.php
Chemaclass: We got the answer from Sebastian: https://twitter.com/s_bergmann/status/1284810322675630080
You @jenshaase have the last word here :)

src/php/Analyzer/TupleSymbol/DefSymbol.php
```diff
@@ -18,59 +20,90 @@ final class DefSymbol
...
     }
 
-    private function createMetaAndInit(Tuple $tuple): array
+    private function ensureDefIsAllowed(Tuple $tuple, NodeEnvironment $env): NodeEnvironment
+    {
+        if (!$env->isDefAllowed()) {
+            throw AnalyzerException::withLocation("'def inside of a 'def is forbidden", $tuple);
+        }
+
+        return NodeEnvironment::withDefNotAllowed($env);
```
jenshaase, line 64: I would only do the check in here and not update the `$env`

tests/php/IntegrationTest.php
jenshaase: Use `$this`

src/php/NodeEnvironment.php
```diff
@@ -66,6 +71,15 @@ public static function empty(): NodeEnvironment
         return new NodeEnvironment([], self::CTX_STMT, [], []);
     }
 
+    public static function withDefNotAllowed(?NodeEnvironment $env = null): NodeEnvironment
```
jenshaase, line 74: If you want such a method you should make it not static: `public function withDefNotAllowed()`. For the test you can easily build it like this: `NodeEnvironment::empty()->withDefNotAllowed()`.

I don't want to have too many static method for `NodeEnvironment` because they than may be used in the Analyzer in the wrong way. E.g. if you call `NodeEnvironment::withDefNotAllowed()`, this will create a new empty node environment and all other states are lost.

src/php/Analyzer/TupleSymbol/DefSymbol.php
```diff
@@ -88,4 +121,17 @@ private function getInitialMetaAndInit(Tuple $tuple): array
...
     }
+
+    /**
+     * @param AbstractType|scalar|null $init
+     */
+    private function analyzeInit($init, NodeEnvironment $env, string $namespace, Symbol $nameSymbol): Node
+    {
+        $initEnv = $env
+            ->withBoundTo($namespace . '\\' . $nameSymbol)
+            ->withContext(NodeEnvironment::CTX_EXPR)
```
jenshaase, line 132: I would change the `env` here and not in the method `ensureDefIsAllowed`

PR #98, opened 2020-07-19
Replace __invoke() to analyze()

Chemaclass:
# Description

In order to improve readability I renamed the `__invoke()` functions to use `analyze()` instead.

# Changes

- Use `analyze()` instead of `__invoke()` for the Analyzer objects and TupleSymbol.
- Introduce an interface (`TupleSymbolAnalyzer`) for all `TupleSymbol` so they are easier to build. 
  - Look the refactoring done in `AnalyzeTuple`.

### DIFF
--- a/src/php/Analyzer/AnalyzeTuple.php
+++ b/src/php/Analyzer/AnalyzeTuple.php
@@ -37,7 +37,7 @@ final class AnalyzeTuple
     use WithAnalyzer;
 
     /** @throws AnalyzerException|PhelCodeException */
-    public function __invoke(Tuple $tuple, NodeEnvironment $env): Node
+    public function toNode(Tuple $tuple, NodeEnvironment $env): Node
     {
         if (!$tuple[0] instanceof Symbol) {
             return (new InvokeSymbol($this->analyzer))($tuple, $env);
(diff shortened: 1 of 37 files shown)


src/php/Analyzer/AnalyzeTuple.php
```diff
@@ -36,58 +37,71 @@ final class AnalyzeTuple
...
 
-        switch ($tuple[0]->getFullName()) {
+    private function getSymbolName(Tuple $tuple): string
+    {
+        return isset($tuple[0]) && $tuple[0] instanceof Symbol
+            ? $tuple[0]->getFullName()
+            : self::EMPTY_SYMBOL_NAME;
+    }
+
+    private function createSymbolByName(string $symbolName): TupleSymbolToNode
```
jenshaase, line 58: I would call it `createSymbolAnalyzerByName`, since it doesn't create a Symbol but an Analyzer.

Chemaclass:
> Instead of `toNode` I would call it `analyze`. The interface is a good idea, but isn't it bad that the return type of the methods is different? (For example `DefNode` instead of `Node`). Seem like PHP isn't that strict on this currently.

Well, actually, `DefNode` is a `Node` because it extends from it. The interface ensures that the returning type is a `Node`, so any of its children are, by definition, a Node as well.

PR #100, opened 2020-07-25
Add DoSymbol unit tests

Chemaclass:
# Description

Added some unit tests for the DoSymbol class

# Changes

- Created `DoSymbolTest`

### DIFF
--- a/tests/php/Analyzer/TupleSymbol/DoSymbolTest.php
+++ b/tests/php/Analyzer/TupleSymbol/DoSymbolTest.php
@@ -25,7 +25,9 @@ final class DoSymbolTest extends TestCase
     public function testEmptyTuple(): void
     {
         $env = NodeEnvironment::empty();
-        $tuple = Tuple::create();
+        $tuple = Tuple::create(
+            Symbol::create(Symbol::NAME_DO),
+        );
 
         $expected = new DoNode(
             $env,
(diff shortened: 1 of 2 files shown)


tests/php/Analyzer/TupleSymbol/DoSymbolTest.php
```diff
@@ -0,0 +1,60 @@
...
+
+    public function setUp(): void
+    {
+        $this->analyzer = new Analyzer(new GlobalEnvironment());
+    }
+
+    public function testEmptyTuple(): void
+    {
+        $env = NodeEnvironment::empty();
+        $tuple = Tuple::create();
```
jenshaase, line 28: This should be:
```
$tuple = Tuple::create(
  Symbol::create(Symbol::NAME_DO),
);
```
which is the equivalent to:
```
(do)
```

Chemaclass:
@jenshaase would it make sense if we create a save guard in `DoSymbol::analyze()` to verify that the Symbol is a `DO`? Similar as we have in `LoopSymbol`. I think we should have these checks in all TupleSymbols, don't we?
```php
if (!($tuple[0] instanceof Symbol && $tuple[0]->getName() === Symbol::DO)) {
    throw AnalyzerException::withLocation("This is not a 'do.", $tuple);
}
```

jenshaase:
@Chemaclass Yes it makes sense to add these guards to all of the TupleSymbols analyzer. I merge this now and you can add it in another PR if you want.

Chemaclass:
I already added the guard for this `DoSymbol`. But it's actually good that we are on the same page with that guard, so we can add more tests and add these guards at the same time :)

PR #101, opened 2020-08-02
Refactoring emitter

Chemaclass:
# Description

The `Emitter` class has currently +1.5k lines, so I thought there is a potential room for refactoring :)

# Changes

The main focus of this PR is NOT to refactor the NodeEmitters or the logic behind but to move the logic from the fat `Emitter` class to other little classes.

- I applied something similar as it was already done to the `Analyzer` class.
- I moved away to other classes a lot of private methods that were previously inside the `Emitter` class.
- I grouped the emitters as:
  - NodeEmitters inside a directory (one node-emitter per ast-node). `NodeEmitterFactory` creates them.
  - and the rest of the literal-emitters are inside another class `LiteralEmitter`.
- The only "logic that I changed" was: removing the static scope of the `Muge` class.

As a follow-up PR we could refactor these new classes in order to clean up a bit more the code :)


JesusValera:
Awesome!! <3

src/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitter.php
```diff
@@ -0,0 +1,67 @@
...
+use Phel\Ast\PhpVarNode;
+use Phel\Emitter\OutputEmitter\NodeEmitter;
+
+final class ApplyEmitter implements NodeEmitter
+{
+    use WithOutputEmitter;
+
+    public function emit(Node $node): void
+    {
+        assert($node instanceof ApplyNode);
```
Chemaclass, line 18: Ideally, this line shouldn't be necessary at this point.
Ideally, we should be able to define the node-child-type in the method signature, like
```php
final class ApplyEmitter implements NodeEmitter
{
    use WithOutputEmitter;

    public function emit(ApplyNode $node): void
    { // ...
```
This is not currently possible with PHP 7.4. But I think it will be allowed with PHP 8.

JesusValera: Yes, I think this will possible on PHP 8.
What about creating an `issue` in order to remember the change when we upgrade the PHP version.

By the way, do you have in mind update to version 8 when it is available or keep on on version 7.4? @jenshaase

Chemaclass: I think it makes total sense to update to version PHP 8. Using docker it should be something trivial :)

jenshaase: I would like to support 7.4 a little bit longer. One of my goals was that Phel can be executed on cheap hosting providers and they normally need a little bit of time to support the newest version of PHP.

PR #103, opened 2020-08-10
Refactor & tests for ApplyEmitter

JesusValera:
# Description

The last test executes the following command:
```clojure
>>> (apply (fn [x] x) [1])
1
>>> 
```

# Changes

- Split the class `ApplyEmitter` logic into different and small methods, also, remove duplication.
- Create `ApplyEmitterTest` class.


tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,93 @@
...
+            'array_reduce([...((\Phel\Lang\Tuple::createBracket(, , );) ?? [])], function($a, $b) { return ($a + $b); });'
+        );
+    }
+
+    public function testPhpVarNodeButNoInfix(): void
+    {
+        $node = new PhpVarNode(NodeEnvironment::empty(), 'str');
+        $args = [
+            new LiteralNode(NodeEnvironment::empty(), 'abc'),
+            new LiteralNode(NodeEnvironment::empty(), 'def'),
```
jenshaase, line 62: The last argument of `apply` must always be a sequence (`Tuple`). Therefore this line must be replaced with this:
```
new TupleNode(NodeEnvironment::empty(), [
  new LiteralNode(NodeEnvironment::empty(), 'def')
])
``

File: tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php [hunk 2]
```diff
@@ -0,0 +1,93 @@
...
+        ];
+
+        $arrayNode = new ApplyNode(NodeEnvironment::empty(), $node, $args);
+        $this->applyEmitter->emit($arrayNode);
+
+        $this->expectOutputString('str(, ...(() ?? []));');
```
jenshaase, line 68: If you fix the comment above, this should result in valid PHP code.

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public function testNoPhpVarNode(): void
```
jenshaase, line 71: No this is not what you wanted, sorry ;)

This is something like
```
(apply (defn foo [arg] (println arg)) [true])
```

Which branch/line do you want to cover with this test. Maybe I can give you a better example.

src/php/Ast/PhpVarNode.php
```diff
@@ -33,7 +33,7 @@ final class PhpVarNode extends Node
         '**',
         '^',
         '<<',
-        '>>'
+        '>>',
```
jenshaase, line 36: I think we should define a rule, if there a comment at the end of the array or not ;)
What do you think?

Chemaclass: I agree. I would vote 👍 
There is a phpcs for it: [trailing_comma_in_multiline_array](https://mlocati.github.io/php-cs-fixer-configurator/#version:2.15|fixer:trailing_comma_in_multiline_array)

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
JesusValera: I want to test the `$this->noPhpVarNode($node)` method (last one).

In other words, I need to test the branch when the `$node->getFn()` won't be an instance of `PhpVarNode` class.

That's why I tried but not sure if the example fits either how to write it. :stuck_out_tongue: 
```clojure
>>> (defn foo [arg] (println arg))
1
>>> (apply foo [true])
true
nil
>>> 
```
Any suggestion?

jenshaase: I would recommend this:

```
(apply (fn [& x] x) [1])
```

but this one is even easier (but doesn't make much sense)

```
(apply (fn [&]) [1])
```

With can be expressed as:

```
$tuple = Tuple::create(
  Symbol::create('apply'),
  Tuple::create(
    Symbol::create('fn'),
    Tuple::createBracket(
      Symbol::create('&')
    ),
  ),
  Tuple::createBracket(
    1
  )
)
```

JesusValera: @jenshaase I have some troubles.

`ApplyEmitter` only `emit()` `Node` objects, so, I guess it should emit the `ApplyNode` object.
The `ApplyNode` constructor needs `NodeEnvironment`, `Node $fn`, `array $arguments` and an optional `SourceLocation`.
I can pass a `NodeEnvironment::empty()` as the first argument, but I don't know what to do with the `Node $fn`.
I tried with a `FnNode` but it needs a lot of parameters (one of them is also another node...). I am not sure if this is right.

In your example, you are creating a Tuple instead of a Node and I don't know how to implement it into the test. Could you take a look, please? :smiley:

jenshaase: Yeah, sorry I was in the wrong context. But you figured it out yourself.

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->applyEmitter->emit($applyNode);
+
+        $this->expectOutputString('str(, ...((\Phel\Lang\Tuple::createBracket();) ?? []));');
+    }
+
+    public function testNoPhpVarNode(): void
+    {
+        $fnNode = new FnNode(
+            NodeEnvironment::empty(),
+            [Symbol::create('&'), Symbol::create('x')],
```
jenshaase, line 68: You can remove the `&` Symbol here. In the full pipeline the Analyzer will remove this and sets `isVariadic` to true (see next comment).

Should look like this then:
```
[Symbol::create('x')]
```

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,92 @@
...
+            new PhpVarNode(NodeEnvironment::empty(), 'x'),
+            [],
+            false,
```
jenshaase, line 71: This is `true` now.

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,92 @@
...
+            ]),
+        ];
+
+        $applyNode = new ApplyNode(NodeEnvironment::empty(), $fnNode, $args);
+        $this->applyEmitter->emit($applyNode);
+
+        $this->expectOutputString('(new class() extends \Phel\Lang\AFn {
+  public const BOUND_TO = "";
+
+  public function __invoke($_AMPERSAND_, $x) {
```
jenshaase, line 87: This changes to `__invoke(...$x)`

jenshaase (changes requested):
Almost done. Just added a few hints.

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
JesusValera: In your previous message you said:
> but this one is even easier (but doesn't make much sense)
> ```clojure
> (apply (fn [&]) [1])
> ```

And if I removed the `Symbol::create('&')`, the final query would be:
```clojure
(apply (fn [x]) [1])
```
Are they the same?

jenshaase: Good question!

First of all, the generated code of `(fn [& x])` and `(fn [&])` is almost the same. I the second example the analyzer will automatically generated a dummy symbol. So the analyzer will transform it to something like this `(fn [& __phel_1233])`

Now let me explain the `&` Symbol. This is just a hint for the analyzer that the next argument is a variadic argument. This is equal to the `...` before a function argument in PHP. These two are equal
```php
<?php
function(...$args) {}
```

```
(fn [& args])
```

So the analyzer recognizes the `&` symbol. But for the Emitter this `&` has no meaning. The emitter just wants to know if the last argument of the function is a variadic argument (so it has to add the three dots). Therefore, the analyzer removes the `&` symbol from the argument list and sets the variadic flag to true.

Hope this makes it clear now ;)

JesusValera: Thank you!

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+    public function testNoPhpVarNode(): void
+    {
+        $fnNode = new FnNode(
+            NodeEnvironment::empty(),
+            [Symbol::create('x')],
+            new PhpVarNode(NodeEnvironment::empty(), 'x'),
+            [],
+            true,
+            false
```
Chemaclass, line 72:
```suggestion
        $isVariadic = true,
        $recurs = false
```
Is a way of simulate "named arguments". This way it will be clear what these bool parameters represent. In general, bool arguments are a bad design (because it's readability and temporal-mapping) but a way to improve them is to add a variable name that represents its value.

JesusValera: You are right!
Thank you :dancers:

tests/php/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,93 @@
...
+
+    public function testNoPhpVarNode(): void
+    {
+        $fnNode = new FnNode(
+            NodeEnvironment::empty(),
+            [Symbol::create('x')],
+            new PhpVarNode(NodeEnvironment::empty(), 'x'),
+            [],
+        $isVariadic = true,
+        $recurs = false
```
Chemaclass, line 72:
```suggestion
            $isVariadic = true,
            $recurs = false
```

JesusValera:
Check it out @jenshaase :slightly_smiling_face:

jenshaase (changes requested):
Sorry that it took so long. But I needed some time to check this out myself. In general the tests look great but I was not happy with the outputted PHP Code. Therefore, I corrected the tests in the other comments. Let me also explain what is happening there:

The Node Environment has a context flag, that says the Emitter how the code should be emitted. The are three states:

* `NodeEnvironment::CTX_EXPR`: This is a expression. Expressions are part of the code which evaluates to a value.
* `NodeEnvironment::CTX_STMT` (default): This is a Statement. Statements do something (mostly sideeffects, like print).
* `NodeEnvironment::CTX_RET`: This is a Return. This is a Expression with a `return` in front.

If you have a look a the LiteralEmitter you can see that it only outputs the value if it is not in the NodeEnvironment::CTX_STMT context. This makes sense if you look at this Phel Code:

```
(do
  1
  2
  3)
```

`do` is a expression that only returns the last value (`3` in this case). If you transform it to PHP it can look like this:
```
1;
2;
return 3;
```

However, the first three lines do not make sense. So the Phel Compiler removes them and the output is this:
```
return 3;
```

tests/php/Unit/Emitter/OutputEmitter/NodeEmitter/ApplyEmitterTest.php
```diff
@@ -0,0 +1,93 @@
...
+                new LiteralNode(NodeEnvironment::empty(), 3),
+                new LiteralNode(NodeEnvironment::empty(), 4),
+            ]),
+        ];
+
+        $applyNode = new ApplyNode(NodeEnvironment::empty(), $node, $args);
+        $this->applyEmitter->emit($applyNode);
+
+        $this->expectOutputString(
+            'array_reduce([...((\Phel\Lang\Tuple::createBracket(, , );) ?? [])], function($a, $b) { return ($a + $b); });'
```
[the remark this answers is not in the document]
JesusValera: Oh, now it makes sense!!
Thank you for your explanation. :tada:

JesusValera:
> Sorry that it took so long. But I needed some time to check this out myself. In general the tests look great but I was not happy with the outputted PHP Code. Therefore, I corrected the tests in the other comments. Let me also explain what is happening there:
> 
> The Node Environment has a context flag, that says the Emitter how the code should be emitted. The are three states:
> 
> * `NodeEnvironment::CTX_EXPR`: This is a expression. Expressions are part of the code which evaluates to a value.
> * `NodeEnvironment::CTX_STMT` (default): This is a Statement. Statements do something (mostly sideeffects, like print).
> * `NodeEnvironment::CTX_RET`: This is a Return. This is a Expression with a `return` in front.
> 
> If you have a look a the LiteralEmitter you can see that it only outputs the value if it is not in the NodeEnvironment::CTX_STMT context. This makes sense if you look at this Phel Code:
> 
> ```
> (do
>   1
>   2
>   3)
> ```
> 
> `do` is a expression that only returns the last value (`3` in this case). If you transform it to PHP it can look like this:
> 
> ```
> 1;
> 2;
> return 3;
> ```
> 
> However, the first three lines do not make sense. So the Phel Compiler removes them and the output is this:
> 
> ```
> return 3;
> ```

Which scenarios we would use `NodeEnvironment::empty()->withContext(NodeEnvironment::CTX_RET` and/or `NodeEnvironment::empty()->withContext(NodeEnvironment::CTX_STMT`?

jenshaase:
> To sum up, we will use CTX_RET for defn or fn for example, SMTM when it was side effects like print something and EXPR when it's just an expression, isn't?

Correct

jenshaase (approved):

jenshaase:
Thanks! It is now merged.

PR #104, opened 2020-08-15
Refactoring FnSymbol

Chemaclass:
# Description

I refactored the `FnSymbol` class and added some unit tests for it.

# Changes

- Refactor `FnSymbol` without altering any logic.
- Added some unit tests for `FnSymbol`. Not completely covered, but some logic. We could add more unit tests if necessary in another iteration.

### DIFF
--- a/src/php/Analyzer/TupleSymbol/FnSymbol.php
+++ b/src/php/Analyzer/TupleSymbol/FnSymbol.php
@@ -6,6 +6,7 @@ namespace Phel\Analyzer\TupleSymbol;
 
 use Phel\Analyzer\WithAnalyzer;
 use Phel\Ast\FnNode;
+use Phel\Ast\Node;
 use Phel\Exceptions\AnalyzerException;
 use Phel\Lang\Symbol;
 use Phel\Lang\Tuple;
@@ -16,54 +17,79 @@ final class FnSymbol implements TupleSymbolAnalyzer
 {
     use WithAnalyzer;
 
+    private array $params = [];
+    private array $lets = [];
+    private bool $isVariadic = false;
+    private bool $hasVariadicForm = false;
+    private string $state = 'start';
+
     public function analyze(Tuple $tuple, NodeEnvironment $env): FnNode
     {
-        $tupleCount = count($tuple);
-        if ($tupleCount < 2) {
+        $this->verifyArguments($tuple);
+        $this->buildParams($tuple);
+        $this->addDummyVariadicSymbol();
+        $this->checkAllVariablesStartWithALetterOrUnderscore($tuple);
+
+        $recurFrame = new RecurFrame($this->params);
+
+        return new FnNode(
+            $env,
+            $this->params,
+            $this->analyzeBody($tuple, $recurFrame, $env),
+            $this->buildUsesFromEnv($env),
+            $this->isVariadic,
+            $recurFrame->isActive(),
+            $tuple->getStartLocation()
+        );
+    }
+
+    private function verifyArguments(Tuple $tuple): void
+    {
+        if (count($tuple) < 2) {
             throw AnalyzerException::withLocation("'fn requires at least one argument", $tuple);
         }
 
         if (!($tuple[1] instanceof Tuple)) {
             throw AnalyzerException::withLocation("Second argument of 'fn must be a Tuple", $tuple);
         }
+    }
 
-        $params = [];
-        $lets = [];
-        $isVariadic = false;
-        $hasVariadicForm = false;
-        $state = 'start';
+    private function buildParams(Tuple $tuple): void
+    {
+        /** @var Tuple $xs */
         $xs = $tuple[1];
+
         foreach ($xs as $param) {
-            switch ($state) {
+            switch ($this->state) {
                 case 'start':
                     if ($param instanceof Symbol) {
                         if ($this->isSymWithName($param, '&')) {
-                            $isVariadic = true;
-                            $state = 'rest';
+                            $this->isVariadic = true;
+                            $this->state = 'rest';
                         } elseif ($param->getName() === '_') {
-                            $params[] = Symbol::gen()->copyLocationFrom($param);
+                            $this->params[] = Symbol::gen()->copyLocationFrom($param);
                         } else {
-                            $params[] = $param;
+                            $this->params[] = $param;
                         }
                     } else {
                         $tempSym = Symbol::gen()->copyLocationFrom($param);
-                        $params[] = $tempSym;
-                        $lets[] = $param;
-                        $lets[] = $tempSym;
+                        $this->params[] = $tempSym;
+                        $this->lets[] = $param;
+                        $this->lets[] = $tempSym;
                     }
                     break;
                 case 'rest':
-                    $state = 'done';
-                    $hasVariadicForm = true;
+                    $this->state = 'done';
+                    $this->hasVariadicForm = true;
                     if ($this->isSymWithName($param, '_')) {
-                        $params[] = Symbol::gen()->copyLocationFrom($param);
+                        $this->params[] = Symbol::gen()->copyLocationFrom($param);
                     } elseif ($param instanceof Symbol) {
-                        $params[] = $param;
+                        $this->params[] = $param;
                     } else {
                         $tempSym = Symbol::gen()->copyLocationFrom($tuple);
-                        $params[] = $tempSym;
-                        $lets[] = $param;
-                        $lets[] = $tempSym;
+                        $this->params[] = $tempSym;
+                        $this->lets[] = $param;
+                        $this->lets[] = $tempSym;
                     }
                     break;
                 case 'done':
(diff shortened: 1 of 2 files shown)


tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,178 @@
...
+        // This is the same as: (anything (fn [& param-1 param-2]))
+        $tuple = Tuple::create(
+            Symbol::create('anything'),
+            Tuple::create(
+                Symbol::create(Symbol::NAME_FN),
+                Symbol::create('&'),
+                Symbol::create('param-1'),
+                Symbol::create('param-2'),
+            ),
+        );
```
jenshaase, line 130: This is not `(anything (fn [& param-1 param-1))` but `(anything (fn & param-1 param-2))`. The test should be like this:
```
$tuple = Tuple::create(
  Symbol::create(Symbol::NAME_FN),
  Tuple::createBracket(
    Symbol::create('&'),
    Symbol::create('param-1'),
    Symbol::create('param-2'),
  )
)
```
Which is equal to `(fn [& param-1 param-2])`.

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,178 @@
...
+        // This is the same as: (anything anything)
+        $tuple = Tuple::create(
+            Symbol::create('anything'),
+            Symbol::create('anything')
+        );
+
+        $this->analyze($tuple);
+    }
+
+    public function testIsVariadic(): void
```
jenshaase, line 51: I would rename it to `testIsNotVariadic`

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,178 @@
...
+    {
+        // This is the same as: (anything (anything))
+        $tuple = Tuple::create(
+            Symbol::create('anything'),
+            Tuple::create(Symbol::create('anything'))
+        );
```
jenshaase, line 57: This should be:
```
$tuple = Tuple::create(
  Symbol::create(Symbol::NAME_FN),
  Tuple::createBracket(
    Symbol::create('anything')
  )
)
```
Which is equal to `(fn [anything])`

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+        // This is the same as: (anything (fn [paramName]))
+        $tuple = Tuple::create(
+            Symbol::create('anything'),
+            Tuple::create(
+                Symbol::create(Symbol::NAME_FN),
+                Symbol::create($paramName)
+            ),
+        );
```
jenshaase, line 83: This should be:
```
$tuple = Tuple::create(
  Symbol::create(Symbol::NAME_FN),
  Tuple::createBracket(
    Symbol::create($paramName)
  )
)
```

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,178 @@
...
+
+        $this->analyze($tuple);
+    }
+
+    /** @dataProvider providerGetParams */
+    public function testGetParams(Tuple $fnTuple, array $expectedParams): void
+    {
+        $tuple = Tuple::create(Symbol::create('anything'), $fnTuple);
+        $node = $this->analyze($tuple);
```
jenshaase, line 139: This is also not correct. See above.

src/php/Analyzer/TupleSymbol/FnSymbol.php
```diff
@@ -16,116 +17,172 @@ final class FnSymbol implements TupleSymbolAnalyzer
...
 
+    private const STATE_START = 'start';
+    private const STATE_REST = 'rest';
+    private const STATE_DONE = 'done';
+
+    private array $params = [];
+    private array $lets = [];
+    private bool $isVariadic = false;
+    private bool $hasVariadicForm = false;
+    private string $state = self::STATE_START;
```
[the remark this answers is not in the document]
Chemaclass: Ah, got it. I will try out with
```php
private function resetParamsAndLetsState(): void
{
    $this->params = [];
    $this->lets = [];
    $this->isVariadic = false;
    $this->hasVariadicForm = false;
    $this->buildParamsState = self::STATE_START;
}
```

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,204 @@
...
+        yield 'DoNode body => (fn [x] x)' => [
+            'tuple' => Tuple::create(
+                Symbol::create(Symbol::NAME_FN),
+                Tuple::createBracket(
+                    Symbol::create('x'),
+                ),
+                Symbol::create('x'),
+            ),
+            'expectedBodyInstanceOf' => DoNode::class,
+        ];
```
Chemaclass, line 197: @jenshaase Could you help me to come write an `tuple` example which will ended up with an  
`expectedBodyInstanceOf => LetNode::class`

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
```diff
@@ -0,0 +1,204 @@
...
+            ],
+        ];
+    }
+
+    /** @dataProvider providerGetBody */
+    public function testGetBody(Tuple $tuple, string $expectedBodyInstanceOf): void
+    {
+        $node = $this->analyze($tuple);
+
+        self::assertInstanceOf($expectedBodyInstanceOf, $node->getBody());
```
Chemaclass, line 183: @jenshaase instead of the `$expectedBodyInstanceOf`, do you know which object might be the result of each `$tuple` once analyzed (`$node->getBody()`)?

I mean, what this would be if I want to assert an `$expectedBody`
```
self::assertEquals($expectedBody, $node->getBody());
```

jenshaase: `$node->getBody` returns the body of the function, wrapped in either a `DoNode` or a `LetNode`. I explained the let node in the other comment. Here is the `DoNode`. For example:
```
Tuple::create(
  Symbol::create(Symbol::NAME_FN),
  Tuple::createBracket(
    Symbol::create('x')
  ),
  Symbol::create('x'),
)
```
This is equal to `(fn [x] x)`. This will be extended by the Analyzer to `(fn [x] (do x))`.

The body of the `FnNode` is now a `DoNode`. The `DoNode` is now a little bit more complex, since `x` is resolved as local variable:
```
new DoNode(
   $nodeEnviroment, 
   array(), 
   new LocalVarNode($nodeEnviroment, $shadowVar, $location)
)
```

If you want to test the body I would recommend to use no arguments and return only literals, like numbers or strings. For example: `(fn [] 1)`.

```
Tuple::create(
  Symbol::create(Symbol::NAME_FN),
  Tuple::createBracket(
  ),
  1,
)
```

The `DoNode` should look now something like this (not tested):
```
new DoNode(
  NodeEnviroment::empty(),
  array(),
  new LiteralNode(NodeEnviroment::empty(), 1)
);
```

On last thing: The FnNode wraps everything in a `DoNode`, because it is the possible to have multiple expression. For example `(fn [] 1 2 3 4 5)` is then equal to `(fn [] (do 1 2 3 4 5))`. It is just easier to implement if the `FnNode` body is just on node instead of a array.

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
Chemaclass: @jenshaase I tried to write that test case "LetNode example (fn [[x y]] x)", but I got an Exception:
> Phel\Exceptions\PhelCodeException : Can not resolve symbol 'first'

I think the line `Destructure:182` is somehow involved in this...
```php
$accessValue = Tuple::create(
    (Symbol::create('first'))->copyLocationFrom($current),
    $lastListSym
)->copyLocationFrom($current);
```
Any idea why we might get this error? I didn't touch the `Destructure` class logic yet.

jenshaase: Oh yes, I forget about this. The `let` expression is also extended and requires a `first` and `next` function (and maybe some others). You have to add them to the `GlobalEnvironment`. I would rewrite the setup function like this:
```
public function setUp(): void
{
  $env = new GlobalEnvironment();
  $env->addDefinition('phel\\core', 'first', new Table()); 
  $env->addDefinition('phel\\core', 'next', new Table());
  $this->analyzer = new Analyzer();
}
```
This code just lets the analyzer know that these functions exists. It doesn't care if the really exists.

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
Chemaclass: Ok, I don't think it worth now to add such unit tests (for the body...). I think they are good enough for this first iteration. Maybe in a second iteration could we add such tests, but I think this PR is big and good enough as it is :)

tests/php/Analyzer/TupleSymbol/FnSymbolTest.php
Chemaclass: Done.

PR #107, opened 2020-08-22
Refactoring ForeachSymbol

Chemaclass:
# Description 📚 

Refactoring `ForeachSymbol`.

# Changes ⛓️ 

- Refactored the `ForeachSymbol` class extracting some logic chunks into smaller private methods.
- Created `ForeachSymbolTest`.


tests/php/Analyzer/TupleSymbol/ForeachSymbolTest.php
```diff
@@ -0,0 +1,129 @@
...
+            Symbol::create(Symbol::NAME_FOREACH),
+            Tuple::create(
+                Symbol::create(''),
+                Tuple::create(Symbol::create('any'), '', '')
+            ),
+            Tuple::create()
+        );
+
+        $node = $this->analyze($mainTuple);
+        self::assertEquals(Symbol::create('any'), $node->getValueSymbol());
```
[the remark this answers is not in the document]
jenshaase: With 2 args it is like this

```
(foreach [x []])
```
with is the same as this PHP version

```
foreach (array() as $x) {}
```

With 3 args it looks like this

```
(foreach [k v {}])
```
with is the same as this PHP version
```
foreach (array() as $k => $v) {}
```

tests/php/Analyzer/TupleSymbol/ForeachSymbolTest.php
```diff
@@ -0,0 +1,129 @@
...
+                )
+            ),
+            'error' => true,
+        ];
+
+        yield '(foreach (()))' => [
+            'mainTuple' => Tuple::create(
+                Symbol::create(Symbol::NAME_FOREACH),
+                Tuple::create(
+                    Symbol::create(''),
```
jenshaase, line 87: Symbol shouldn't be a empty string. From the point of the parser this can never happen. Maybe we need to validate this in the Symbol.

tests/php/Analyzer/TupleSymbol/ForeachSymbolTest.php
```diff
@@ -0,0 +1,129 @@
...
+            'mainTuple' => Tuple::create(
+                Symbol::create(Symbol::NAME_FOREACH),
+                Tuple::create(
+                    Symbol::create('x')
+                )
+            ),
+            'error' => true,
+        ];
+
+        yield '(foreach (()))' => [
```
jenshaase, line 83: This is equal to the 2 args version. `(foreach [x []])`

jenshaase (changes requested):
I added some examples.

tests/php/Analyzer/TupleSymbol/ForeachSymbolTest.php
Chemaclass: Would this be enough/ok?
```php
public function __construct(?string $namespace, string $name)
{
    assertNotEmpty($name, 'Symbol name can not be empty');
```

jenshaase: I would throw an `AnalyzerException`.

Chemaclass: @jenshaase I think I will create another PR for this. Actually, should we only check that the `$name` is not empty, but it can contain any value, right? Or do you have any acceptance-criteria for that in mind?

jenshaase: Yes, it can contain any other value.

tests/php/Unit/Analyzer/TupleSymbol/ForeachSymbolTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+    public function testValueSymbolFromTupleWith3Args(): void
+    {
+        // (foreach [key value {}])
+        $mainTuple = Tuple::create(
+            Symbol::create(Symbol::NAME_FOREACH),
+            Tuple::createBracket(
+                Symbol::create('key'),
+                Symbol::create('value'),
+                Tuple::createBracket()
```
jenshaase, line 86: It doesn't matter for the test. But to keep it correct, this must be `Table::empty()` and the comment above must be `// (foreach [key value @{}])` (`@` is missing). The same applies to the the next test.

jenshaase (changes requested):
Minor changes required. But otherwise it look very good!

tests/php/Unit/Analyzer/TupleSymbol/ForeachSymbolTest.php
Chemaclass: Done

jenshaase (approved):

jenshaase:
Thanks a lot! I merged it now!

PR #108, opened 2020-08-23
Split integration and unit tests

Chemaclass:
# Description 🔖 

Split the integration and unit tests in different test suites.

# Changes 📚 

- Move all integration tests into `php/tests/integration` folder.
- Move all unit tests into `php/tests/unit` folder.
- `phpunit.xml` config updated according each different test suite.
- Run the integration and unit tests in different steps during [CI](https://github.com/jenshaase/phel-lang/pull/108/checks).
- Update the `Readme > Testing` section.
- Replace all `$this->assertEquals(` to `self::assertEquals(` in order to unify the style.

## Why is this distinction important❓

Unit tests and Integration tests are fundamentally different for multiple reasons. They test differently with different purposes and objectives. Unit tests focus on unit logics (not classes but behavior!), where its logic must be fully isolated from the outside.  A test is not a unit test if: 

* It talks to the database
* It communicates across the network
* It touches the file system
* It can't run at the same time as any of your other unit tests
* You have to do special things to your environment (such as editing config files) to run it.

### Links 🔗 

* https://www.artima.com/weblogs/viewpost.jsp?thread=126923
* https://www.guru99.com/unit-test-vs-integration-test.html


jenshaase (commented):
Maybe we should wait with this until the other pull request have been merged? What do you think?

Chemaclass:
@jenshaase I think we can merge it as soon as there are no conflicts with `master`. 
The open PRs need to adjust and solve their own conflicts (there is one (or maybe two) open PR which we are aware of), so there is no problem with that, no worries. @JesusValera and I are aware of this PR and the potential conflicts. We are going to solve the conflicts, alles im Griff :)

jenshaase:
Ok. I merged it.

PR #110, opened 2020-09-04
Rename CTX_* constants by CONTEXT_*

JesusValera:
# Description

From my point of view, it is better to have long but pronounceable nouns than abbreviations. :zipper_mouth_face: 

# Changes

- Replace `CTX_STMT` by `CONTEXT_STATEMENT`
- Replace `CTX_EXP` by `CONTEXT_EXPRESSION`
- Replace `CTX_RET` by `CONTEXt_RETURN`

### DIFF
--- a/src/php/NodeEnvironment.php
+++ b/src/php/NodeEnvironment.php
@@ -8,7 +8,7 @@ use Phel\Lang\Symbol;
 
 final class NodeEnvironment
 {
-    public const CONTEXT_EXPRESSION = 'expresssion';
+    public const CONTEXT_EXPRESSION = 'expression';
     public const CONTEXT_STATEMENT = 'statement';
     public const CONTEXT_RETURN = 'return';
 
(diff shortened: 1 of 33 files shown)


Chemaclass (approved):
Yes ⚡ !

src/php/NodeEnvironment.php
```diff
@@ -8,9 +8,9 @@
 
 final class NodeEnvironment
 {
-    public const CTX_EXPR = 'expr';
-    public const CTX_STMT = 'stmt';
-    public const CTX_RET = 'ret';
+    public const CONTEXT_EXPRESSION = 'expr';
```
jenshaase, line 11: You could also rename the strings to 'expression', 'statement' and 'return'. I don't think the strings are used somewhere other than with these constants.

jenshaase (changes requested):

src/php/NodeEnvironment.php
JesusValera: Agreed! :)

jenshaase (approved):
Super!

PR #111, opened 2020-09-06
Refactor Lang folder

JesusValera:
# Description

Small changes on `src/php/Lang/*`.

# Changes

- Add `declare(strict_types=1)` to some classes.
- Remove the comment tag `@return` when it is explicitly in the code.
- Fix small typos like `$offest` to `$offset` or some return values from some methods.
- `Lang/Keyword` replace `if` clause by a coalesce operator.
- Split imports when the line is too long ([PSR-12](https://www.php-fig.org/psr/psr-12/#41-extends-and-implements)).
- Replace the call to the class itself by `self::` keyboard.
- Other minor changes.

# Extra

I have updated some PhpDocs with a final dot when it was a comment but without it when it was a `@param` tag.
Looking at the `php cs fixer` I found this rule: [`phpdoc_annotation_without_dot`](https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16|fixer:phpdoc_annotation_without_dot).
I think we should follow a unique standard and for me, this one fits good (apparently there isn't a rule as final dot only on comments but no on @param tag). :thinking: 

Also, take a look at this rule too: [phpdoc_add_missing_param_annotation](https://mlocati.github.io/php-cs-fixer-configurator/#version:2.16|fixer:phpdoc_add_missing_param_annotation)
Maybe I could create a new `issue` when we can discuss those and other `cs fixer` rules. :smile: 

What do you think @Chemaclass & @jenshaase?

### DIFF
--- a/src/php/Lang/AbstractType.php
+++ b/src/php/Lang/AbstractType.php
@@ -17,7 +17,7 @@ abstract class AbstractType implements IMeta, ISourceLocation
     /**
      * Check if $other is equals to $this.
      *
-     * @param mixed $other The other value.
+     * @param mixed $other The other value
      */
     abstract public function equals($other): bool;
 }
--- a/src/php/Lang/ICdr.php
+++ b/src/php/Lang/ICdr.php
@@ -1,15 +1,13 @@
 <?php
 
+declare(strict_types=1);
+
 namespace Phel\Lang;
 
 interface ICdr
 {
-
     /**
-     * Return the sequence without the first element. If the sequence is empty
-     * returns null.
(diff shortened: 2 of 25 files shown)


Chemaclass:
I would vote 👍🏼 to add those two new rules:
* `phpdoc_annotation_without_dot` 
* `phpdoc_add_missing_param_annotation`

jenshaase:
Feel free to add the two new rules!

jenshaase (approved):
Cool. Looks much better now!

PR #117, opened 2020-09-17
Move out phel entry point logic

Chemaclass:
# Description

Clean up the Main class.

# Changes

- Move out the entry point logic from the Main class to the entry point script (phel).


jenshaase (approved):
Looks good.

PR #122, opened 2020-10-24
Adding tests to IfSymbol Tuple

Chemaclass:
# Description

Adding tests to IfSymbol Tuple.

# Changes

- I refactored a little bit and I created some unit tests for the `Tuple\IfSymbol`

### DIFF
--- a/src/php/Analyzer/TupleSymbol/IfSymbol.php
+++ b/src/php/Analyzer/TupleSymbol/IfSymbol.php
@@ -12,8 +12,7 @@ use Phel\Lang\Tuple;
 use Phel\NodeEnvironment;
 
 /**
- * (if condition truthy).
- * (if condition truthy falsy).
+ * (if test then else?).
  */
 final class IfSymbol implements TupleSymbolAnalyzer
 {
@@ -25,9 +24,9 @@ final class IfSymbol implements TupleSymbolAnalyzer
 
         return new IfNode(
             $env,
-            $this->condition($tuple, $env),
-            $this->truthy($tuple, $env),
-            $this->falsy($tuple, $env),
+            $this->testExpression($tuple, $env),
+            $this->thenExpression($tuple, $env),
+            $this->elseExpression($tuple, $env),
             $tuple->getStartLocation()
         );
     }
@@ -41,7 +40,7 @@ final class IfSymbol implements TupleSymbolAnalyzer
         }
     }
 
-    private function condition(Tuple $tuple, NodeEnvironment $env): Node
+    private function testExpression(Tuple $tuple, NodeEnvironment $env): Node
     {
         $envWithDisallowRecurFrame = $env
             ->withContext(NodeEnvironment::CONTEXT_EXPRESSION)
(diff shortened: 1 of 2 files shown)


tests/php/Unit/Analyzer/TupleSymbol/IfSymbolTest.php
```diff
@@ -0,0 +1,80 @@
...
+        yield 'No arguments provided: (if)' => [
+            'tuple' => Tuple::create(
+                Symbol::create(Symbol::NAME_IF)
+            ),
+        ];
+
+        yield 'Only one argument provided: (if one)' => [
+            'tuple' => Tuple::create(
+                Symbol::create(Symbol::NAME_IF),
+                Symbol::create('one'),
```
Chemaclass, line 42: @jenshaase I got a question for you here: is it really necessary to decorate these literal values as Symbols? I mean, for example, the test will still pass(green) with these:
```php
'tuple' => Tuple::create(
    Symbol::NAME_IF,
    'one',
),
// OR
'tuple' => Tuple::create(
    Symbol::NAME_IF,
    'one',
    'two',
    'three',
    'four',
),
```

src/php/Analyzer/TupleSymbol/IfSymbol.php
```diff
@@ -11,34 +11,51 @@
...
+    {
+        $envWithDisallowRecurFrame = $env
+            ->withContext(NodeEnvironment::CONTEXT_EXPRESSION)
+            ->withDisallowRecurFrame();
 
-        return new IfNode(
-            $env,
-            $testExpr,
-            $thenExpr,
-            $elseExpr,
```
Chemaclass, line 36: I saw that we use in the [control flow docu](https://github.com/jenshaase/phel-lang/tree/master/doc/content/documentation/control-flow.md) this wording: test, then, else.
Screenshot 2020-10-24 at 15 01 55

But I think `(if condition truthy falsy)` is more expressive. What do you think? If you like the idea, I would create a PR renaming from the docu and related code 
- test -> condition
- then -> truthy
- else -> falsy

jenshaase: In other Lisp documentations you will find the wording `test`, `then` and `else`. Therefore, I used it like they did.

tests/php/Unit/Analyzer/TupleSymbol/IfSymbolTest.php
jenshaase: Yes you can do this. It doesn't matter for the test. But it is different Phel code and you may want to update it in the string above.

```
(if one two three four)
```
vs
```
(if "one" "two" "three" "four")
```

jenshaase (changes requested):
Looks good. But leave the wording `test`, `then` and `else`.

src/php/Analyzer/TupleSymbol/IfSymbol.php
Chemaclass: Ah, then I will modify it in my PR as well to use this same naming

jenshaase (approved):
Cool! Thanks.

PR #123, opened 2020-11-02
Decoupling Command dependencies

Chemaclass:
## 📚 Description

In order to start decoupling the details from the (main:business)logic, I think the best way is to start from the outside.

In this PR I decouple all dependencies (the ones that make sense) from the commands: `Phel\Commands\*`

## 🔖 Changes

- Create a `CommandFactory` 
- Inject the `ReplCommand`, `RunCommand`, and `TestCommand` dependencies via its constructor.
- Rename `Commands` (plural) directory to `Command` (singular), to follow the singular convention naming directory.


JesusValera (approved):

jenshaase (approved):
Finally found some time to review this. Looks very good!

PR #126, opened 2020-12-01
Principles of Component Cohesion & Coupling focus on the Compiler Module

Chemaclass:
# Description

I think it would be better to define the boundaries between the different modules behind this project.

# Changes

- I moved the Emitter, Lexer, Reader, Analyzer, Ast inside the Compiler module
- Adjusted namespaces
- Created some Factory for the creation of some objects

The "main thing" that I applied in this refactoring:
- Principles of Component Cohesion & Coupling

You will learn them in books such:
- Clean Architecture (Uncle Bob)
- Principle of Package Design (Matthias Noback)

Some links with some info about them:
- https://blog.avenuecode.com/principles-of-package-and-component-design
- https://adriancitu.com/tag/the-common-reuse-principle/


jenshaase (approved):
Lots of stuff :) Thanks! The code becomes super awesome.

PR #132, opened 2020-12-06
Creating NodeEnvironmentInterface

Chemaclass:
## 📚 Description

Invert the NodeEnvironment dependencies by creating an interface for it.

I have some doubts about the usage of the constants from the interface directly:
``` php 
interface NodeEnvironmentInterface
{
    public const CONTEXT_EXPRESSION = 'expression';
    public const CONTEXT_STATEMENT = 'statement';
    public const CONTEXT_RETURN = 'return';
    // ...
#  -----------------------------------------
NodeEnvironmentInterface::CONTEXT_EXPRESSION
``` 
But I decided to leave them there (in the interface) for simplicity right now.

Apart from this, I think it's important to mention that the interface has too many functions, which is a smell for the "Interface Segregation Principle" violation, but that's out of the scope of this PR.

## 🔖 Changes

- Use the `NodeEnvironmentInterface` as an argument where its concrete `NodeEnvironment` was specified.

### DIFF
--- a/src/php/Compiler/Analyzer.php
+++ b/src/php/Compiler/Analyzer.php
@@ -28,7 +28,7 @@ final class Analyzer implements AnalyzerInterface
         $this->globalEnvironment = $globalEnvironment;
     }
 
-    public function resolve(Symbol $name, NodeEnvironment $env): ?Node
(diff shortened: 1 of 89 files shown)


jenshaase (approved):
Looks good!

PR #134, opened 2020-12-23
Refactor Printer

JesusValera:
## 📚 Description

This PR solves the issue https://github.com/jenshaase/phel-lang/issues/129.

The `Printer` class goes from being a "God Class" to an entry point for each type of Print.

## 🔖 Changes

- Improve the `Printer` class
- Create `PrinterInterface`
- Create a specific class for each Print type.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,12 +33,21 @@
         }
     },
     "scripts": {
-        "test-compiler": "vendor/bin/phpunit",
-        "test-compiler:coverage": "vendor/bin/phpunit --coverage-html data/coverage-report",
+        "test-all": [
+            "@test-quality",
+            "@test-compiler",
+            "@test-core"
+        ],
+        "test-quality": [
+            "@csrun",
+            "@psalm"
+        ],
+        "test-compiler": "./vendor/bin/phpunit",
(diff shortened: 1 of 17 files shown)


Chemaclass (approved):
I am not a fan of using `@psalm-suppress`, but in this case, I think it's necessary because of the missing param type from the interface. 

Actually, instead of doing multiple `if`s, I would suggest using a map (key-value dictionary) to increase the speed of the type-search to the max [O(1)]. But we could do that in another iteration/PR. I am happy with the progress done in this PR, good job! 👍🏼

jenshaase:
Can't we use Psalm's template inheritance (https://psalm.dev/docs/annotating_code/templated_annotations/) instead of `@psalm-suppress`:

```php
<?php
/**
 * @template T
 */
interface IFoo {}

/**
 * @implements IFoo<int>
 */
class Foo implements IFoo {}
```

Chemaclass (approved):
Much better using this psalm-generics feature! 🎉

JesusValera:
I didn't know about the @template & @implements. Super nice this way @jenshaase 🎉

jenshaase:
It is merged. Thanks for the work!

PR #135, opened 2020-12-23
Bug: Fixes typo in post-global-valid

mabasic:
## 📚 Description

When sending a POST request with form data, `(get (http/request-from-globals) :parsed-body)` would always return `nil`.

## 🔖 Changes

- Fixes typo `:conent-type` to `:content-type`. 

Note: Unable to write a test for this without changing phel/http library.

@jenshaase I have managed to write a test for this, but I had to make a small change to `request-from-globals`. I will create a new PR for that change.


jenshaase (approved):
Good catch!

PR #136, opened 2020-12-26
Adds test for post with form data.

mabasic:
## 📚 Description

It adds a test for issue fixed in this PR #135.

## 🔖 Changes

- Adds test for post with form data. 
- Adds `post` parameter to `request-from-globals` in order to be able to test it. @jenshaase This seemed like the fastest way of doing this. We could add a method like `get-post-from-globals` like we have `get-method-from-globals`, but this seemed simpler. Open for suggestions.

### DIFF
--- a/src/phel/http.phel
+++ b/src/phel/http.phel
@@ -146,11 +146,11 @@
 
 (defn request-from-globals-args
   "Extracts a request from args"
-  [& [_server _get _post _cookies _files]]
-  (let [method (get-method-from-globals _server)
-        uri (uri-from-globals _server)
-        headers (headers-from-server _server)
-        server-protocol (get _server "SERVER_PROTOCOL")
+  [server get-parameter post-parameter cookies files]
+  (let [method (get-method-from-globals server)
+        uri (uri-from-globals server)
+        headers (headers-from-server server)
+        server-protocol (get server "SERVER_PROTOCOL")
         version (if (string? server-protocol)
                   (php/str_replace "HTTP/" "" server-protocol)
                   "1.1")]
@@ -158,11 +158,11 @@
       method
       uri
       headers
-      (when (post-global-valid method headers) (php-array-to-table _post))
-      (php-array-to-table _get)
-      (php-array-to-table _cookies)
-      (php-array-to-table _server)
-      (normalize-files _files)
+      (when (post-global-valid method headers) (php-array-to-table post-parameter))
+      (php-array-to-table get-parameter)
+      (php-array-to-table cookies)
+      (php-array-to-table server)
+      (normalize-files files)
       version)))
 
 (defn request-from-globals
--- a/src/phel/http.phel
+++ b/src/phel/http.phel
@@ -144,16 +144,13 @@
       (= method "POST")
       (php/in_array types (php/array "application/x-www-form-urlencoded" "multipart/form-data")))))
 
-
-(defn request-from-globals
-  "Extracts a request from `$_SERVER`"
-  [& [server post]]
-  (let [server (or server php/$_SERVER)
-        method (get-method-from-globals server)
-        uri (uri-from-globals server)
-        headers (headers-from-server server)
-        post (or post php/$_POST)
-        server-protocol (get server "SERVER_PROTOCOL")
+(defn request-from-globals-args
+  "Extracts a request from args"
+  [& [_server _get _post _cookies _files]]
+  (let [method (get-method-from-globals _server)
+        uri (uri-from-globals _server)
+        headers (headers-from-server _server)
+        server-protocol (get _server "SERVER_PROTOCOL")
         version (if (string? server-protocol)
                   (php/str_replace "HTTP/" "" server-protocol)
                   "1.1")]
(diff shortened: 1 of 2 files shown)


jenshaase:
Cool. But I would change the code a little bit. I would split the request-from-globals method into two method. Create a new method request-from-globals-args. This methods has 5 arguments: server, get, post, cookies and files. The request-from-globals method than has zero arguments and just calls request-from-globals-args with $_Server, $_Post, etc. The request-from-globals-args method can than be used for testing.

mabasic:
I try to make this happen.

mabasic:
@jenshaase I had to change the arguments names from `server get post cookies files` to `_server _get _post _cookies _files` because `(get server "SERVER_PROTOCOL")` would throw a strange exception because `get` was being overridden.

Btw. It took me quite some time to figure out the error. This was the output:

```
Error: Typed property Phel\Compiler\Emitter\OutputEmitter\SourceMap\SourceMapConsumer::$vlq must not be accessed before initialization
in /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Compiler/Emitter/OutputEmitter/SourceMap/SourceMapConsumer.php:29 (gen: /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Compiler/Emitter/OutputEmitter/SourceMap/SourceMapConsumer.php:29)

PHP Notice:  Undefined index: args in /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Exceptions/TextExceptionPrinter.php on line 108
PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Phel\Exceptions\TextExceptionPrinter::buildPhpArgsString() must be of the type array, null given, called in /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Exceptions/TextExceptionPrinter.php on line 108 and defined in /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Exceptions/TextExceptionPrinter.php:148
Stack trace:
#0 /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Exceptions/TextExceptionPrinter.php(108): Phel\Exceptions\TextExceptionPrinter->buildPhpArgsString()
#1 /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Runtime/Runtime.php(71): Phel\Exceptions\TextExceptionPrinter->printStackTrace()
#2 [internal function]: Phel\Runtime\Runtime->exceptionHandler()
#3 {main}
  thrown in /home/mabasic/code/src/github.com/mabasic/phel-lang/src/php/Exceptions/TextExceptionPrinter.php on line 148
```

Nothing useful there...

jenshaase:
I will look into it.

src/phel/http.phel
```diff
@@ -144,29 +144,32 @@
...
-  "Extracts a request from `$_SERVER`"
-  [& [server]]
-  (let [server (or server php/$_SERVER)
-        method (get-method-from-globals server)
-        uri (uri-from-globals server)
-        headers (headers-from-server server)
-        server-protocol (get server "SERVER_PROTOCOL")
+(defn request-from-globals-args
+  "Extracts a request from args"
+  [& [_server _get _post _cookies _files]]
```
jenshaase, line 149: This line should be like this:
```
[server get-parameter post-parameter cookies files]
```
The `&` is not required because the method should have 5 arguments. I also renamed the `get` argument to `get-parameter` because otherwise this will overwrite the global `get` function in this scope.

Afterwards you have to adopt `_server` to `server`, `_get` to `get-parameter`, ... in the function body.

tests/phel/test/http.phel
```diff
@@ -230,7 +230,7 @@
   (is
     (=
       "GET"
-      (get (h/request-from-globals (php-associative-array
+      (get (h/request-from-globals-args (php-associative-array
```
jenshaase, line 233: After you applied the changes above the function has now five arguments that you need to provide. Therefore the test will look like this:
```
(deftest test-request-get
  (is
    (=
      "GET"
      (get 
        (h/request-from-globals-args
          (php-associative-array "REQUEST_METHOD" "GET")
          (php-associative-array)
          (php-associative-array)
          (php-associative-array)
          (php-associative-array)) 
        :method)
    )
    "request method get"))
```

The same applies for the other tests below that use `h/request-from-globals-args`

jenshaase (changes requested):
@mabasic I found the errors. Can you fix them?

tests/phel/test/http.phel
mabasic: I used & to avoid having empty associative arrays in tests that don't use those parameters since it seems cleaner. What do you think?

mabasic:
@jenshaase Sure thing, but I am unsure about leaving out &. I find it cleaner not to have to specify all parameters. See comment on code.

jenshaase:
It makes no sense for this method. Normally you would use `request-from-globals` and not `request-from-globals-args`. This method is just for testing and therefore can have all arguments.

mabasic:
Ok. Will update the code later

mabasic:
@jenshaase  Updated.

jenshaase (approved):

jenshaase:
Ok cool. I merged it!

PR #138, opened 2020-12-26
Modularize printer component

JesusValera:
## 📚 Description

The `PrinterInterface` wasn't an interface for the `Printer` but for the printer elements (eg: `ArrayPrinter`, `BooleanPrinter`...), it was a bit confusing how it was done.

Also, the `Printer` class was in the root folder, and the _Printer_ folder was with the other modules mixed as they don't have a relation.

The idea of this PR was to rename the interface to `TypePrinterInterface` and place all printer stuff together.

## 🔖 Changes

- Rename `PrinterInterface` to `TypePrinterInterface`
- Place `Printer` inside the Printer folder and the printer elements inside `Printer/TypePrinter/`
- Add tests for the availables printer types

----

## ⚠️ Extra

~~The `phpunit` coverage HTML was not working because we upgraded `phpunit` to  version 9.4 AND `xdebug` to version 3.
For `phpunit` I upgraded the `phpunit.xml.dist`, some rules were outdated.
For the `xdebug` after trying hard, the only way I found to make it work was replacing the `xdebug` by `phpdbg` ([link](https://susi.dev/phpdbg-for-coverage))(xdebug.mode=coverage has to be set in php.ini).~~

~~I think this is not the best way, but for the moment it could work.
I would like to focus on this topic further in the future, what do you think @jenshaase?~~

Upgrade `phpunit.xml.dist` file and PHPUnit to 9.5.
Test coverage report is working again 🎉


jenshaase (changes requested):
Can we rename `ElementPrinter` to `TypePrinter`?

JesusValera:
> Can we rename `ElementPrinter` to `TypePrinter`?

Absolutely!

src/php/Printer/Printer.php
```diff
@@ -53,23 +53,23 @@ private function __construct(bool $readable)
      */
     public function print($form): string
     {
-        $printerName = $this->getPrinterName($form);
-        $printer = $this->createPrinterByName($printerName);
+        $printerName = $this->getElementPrinterName($form);
```
jenshaase, line 56: Rename to `getTypePrinterName`

src/php/Printer/Printer.php
```diff
@@ -53,23 +53,23 @@ private function __construct(bool $readable)
...
+        $printer = $this->createElementPrinterByName($printerName);
```
jenshaase, line 57: Rename to `createTypePrinterByName`

tests/php/Unit/Printer/TypePrinter/SymbolPrinterTest.php
```diff
@@ -0,0 +1,43 @@
...
+        yield [
+            'expected' => 'name',
+            'symbol' => Symbol::createForNamespace(null, 'name'),
+        ];
+        yield [
+            'expected' => 'name',
+            'symbol' => Symbol::create('name'),
+        ];
+        yield [
+            'expected' => '',
```
jenshaase, line 39: Why is this empty? Shouldn't this be `\\?#__\|\/`

jenshaase (changes requested):

src/php/Printer/Printer.php
Chemaclass: Yes, I was about to write the same 😆

tests/php/Unit/Printer/TypePrinter/SymbolPrinterTest.php
JesusValera: Because the named constructor does a `strpos($name, '/');` (name in this case is the `\\?#__\|\/` string). The idea is to split the string for "removing" the namespace.
This particular example was not so good as the delimiter '/' was in the last position, that's why the `expected` var was empty. Test updated :)

Chemaclass (commented):
The code seems good to me. I don't understand why Scrutinizer is red 🤔 
Any ideas, @jenshaase? Can you share the config file that we have in Scrutinizer for phel? (I can not see it, missing rights for it). 

Might it be the problem due to the change in `test-compiler:coverage` using the `phpdbg` command? 

This is what I have in my config for php7.4 projects in Scrutinizer
```
build:
  environment:
    php: 7.4
  nodes:
    analysis:
      project_setup:
        override:
          - 'true'
      tests:
        override:
          - php-scrutinizer-run
    tests: true
    coverage:
      tests:
        override:
          -
            command: 'vendor/bin/phpunit tests --coverage-clover tests/clover.xml --configuration phpunit.xml'
            coverage:
              file: 'tests/clover.xml'
              format: 'clover'
checks:
  php: true
tools:
  php_cs_fixer: true
  ```

My thoughts are that if you are using it as a command for coverage tests the `test-compiler:coverage` then that might be the reason for the failure, but I don't know...

Chemaclass:
Ah! Got it, the problem is the version of PHPUnit! Update PHPUnit from `9.4.2` to `9.5` and it will work. You don't need the `phpdgb` thing anymore, @JesusValera

jenshaase:
I restarted the Scrutinizer job. It was just an error on their site. It is now green again.

Chemaclass (approved):
Nice one 🥳

JesusValera:
@jenshaase upgraded PHPUnit to 9.5.
Now it works without hacks and strange things :P

jenshaase (approved):

jenshaase:
Thank you. I merged it!

PR #144, opened 2020-12-30
Adding var-dumper as require-dev dependency

Chemaclass:
## 📚 Description

Sometimes is really useful to have the "**var-dumper Symfony Component**" to be able to "dump" some values during run time for debugging purposes. This dependency provides you with two really nice functions:

- `dump(...$var)`: it nicely dumps anything (in a beautiful format, actually!)
- `dd(...$var)`: it dumps and dies, similar to the previous dump 😄 

The fact that it's a **require-dev** dependency shouldn't affect the prod run time for the final consumer/client.

## 🔖 Changes

- Adding `symfony/var-dumper` as **require-dev** dependency

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
         "phpunit/phpunit": "^9.5",
         "vimeo/psalm": "^3.11",
         "friendsofphp/php-cs-fixer": "^2.16",
-        "infection/infection": "^0.16.3"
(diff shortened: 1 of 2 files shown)


JesusValera (approved):
Agree 👍🏼

jenshaase (approved):

jenshaase:
It's merged now!

PR #148, opened 2021-01-01
Introducing parser factory

Chemaclass:
## 📚 Description

Introduce some factories in order to create the expression parsers & readers.

## 🔖 Changes

- Created:
  - `ExpressionParserFactory` as dependency for the `Parser`
  - `ExpressionReaderFactory` as dependency for the `Reader`


jenshaase (approved):
Looks good!

PR #149, opened 2021-01-03
Created CommandFacade

Chemaclass:
## 📚 Description

Introducing the Facade pattern in the Command module

## 🔖 Changes

- Created CommandFacade, and injecting it into the PhelFacade.

---

### Facade + Factory patterns Combo

The Facade is the only entry point from a module. It will receive its factory in order to create whatever it needs and perform the actions from its "creations".

🤔 I am thinking about how to improve the current design of the project one step forward by following this pattern in all modules (if possible and if it makes sense). I will work in another branch to try some ideas 💻 🏭

### DIFF
--- a/phel
+++ b/phel
@@ -4,6 +4,7 @@
 $slash = DIRECTORY_SEPARATOR;
 require_once __DIR__ . "{$slash}src{$slash}php{$slash}PhelFacade.php";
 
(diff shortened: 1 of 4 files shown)


jenshaase (approved):

jenshaase:
Looks good. It is merged now.

PR #150, opened 2021-01-03
Reorganizing Compiler Module

Chemaclass:
## 📚 Description

In order to avoid increasing the size of the root directory from the Compiler Module, and keep the common things together, I thought it might be interesting to have them grouped in submodules in order to simplify the design.

This is an idea that I had since I realized the Compiler Module gets growing and getting bigger, but I never found time to try it out.

The motivation for this PR is the very same as for https://github.com/jenshaase/phel-lang/pull/126 but applied to the context of the Compiler Module this time.

## 🔖 Changes

These are the directories(submodules) in which the Compiler module gets splitter, leaving on its root directory only the `CompilerFactory`.
- Compiler
  - Analyzer
  - Ast
  - Compiler
  - Emmiter
  - Environment
  - Lexer
  - Parser

@jenshaase, to check out the code I think it will be easier to pull the branch and check it out locally because I changed a lot of files (mostly because of the namespaces). I am also open to renaming any folder and/or also move to another folder any file if I moved it wrong.


jenshaase:
I would like to make the following changes:

* Move AST to Analyzer. The result of the Analyzer is an Abstract Syntax Tree (AST).
* Also the Environment folder can be moved to the Analyzer. The node/global environment is mainly used by the Analyzer.
* I would not make a Compiler directory inside the Compiler Directory. The File Compiler for example combines every step of the build chain. Therefore, I think it is ok to leave it in the main Compiler Directory
* Optional: Split the Parser directory into a Parser and Reader directory. Then we have all steps of the compiler pipeline in one folder.

Currently the compiler pipeline is this:
1. Lexer
2. Parser
3. Reader
4. Analyzer
5. Emitter

Chemaclass:
I prefer your idea. I will change it to end up with your suggestion folder structure 🧵

Chemaclass:
@jenshaase double check that I didn't miss anything from your suggestions, please. Feel free to point any possible improvement related to the scope of this PR again. Otherwise, we could merge this one. 

I like the result so far. It certainly helps us to visualize the scope and responsibilities of the Compiler Module itself. The very same principle applies to all rest modules 😄

jenshaase (approved):
Looks very good now! I like it!

PR #159, opened 2021-01-09
Implement str-contains? in phel

Chemaclass:
## 📚  Description

I was thinking about to implement something like the new [`str_contains`](https://www.php.net/manual/en/function.str-contains) (able in PHP 8). So I did.

## 🔖 Changes

- Added the `str-contains?` function in phel with some unit tests
- Updated the API docu using `php doc/build-api-page.php > doc/content/documentation/api.md`

### DIFF
--- a/doc/content/documentation/api.md
+++ b/doc/content/documentation/api.md
@@ -689,6 +689,13 @@ Shorthand for `(if (not condition) else then)`.
 ```
 Increments `x` by one.
 
+## `includes?`
+
+```phel
+(includes? s substr)
+```
+True if s includes substr.
+
 ## `indexed?`
 
 ```phel
--- a/doc/content/documentation/api.md
+++ b/doc/content/documentation/api.md
@@ -689,13 +689,6 @@ Shorthand for `(if (not condition) else then)`.
 ```
 Increments `x` by one.
 
-## `includes?`
-
-```phel
-(includes? s substr)
-```
-True if s includes substr.
-
 ## `indexed?`
 
 ```phel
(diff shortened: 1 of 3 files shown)


JesusValera (approved):

jenshaase:
What I don't like is the name `includes?` can also mean that you want to check if a value is in an array. To solve it there are two options in my opinion.

1. Move all string method to a separate namespace. It is no much clearer that `includes?` belongs to strings. However, you then have to to load the namespace in every file want to work with strings. Clojure did this. However, in the Web-Dev-World you work a lot with string. Therefore, I would like to have them in `phel\core`.
2. Have a name prefix on string methods like `string-includes?` or `str-includes?`.

What do you think?

Chemaclass:
mmm, you got a valid point. In that case, to keep the compatibility with the upcoming (in php8) str_contains, I will rename it to `str-contains` and everyone is happy. Thumbs up for your confirmation, @jenshaase 

PS: I don't think we need the `?` in this function name, do we? Like `str-contains?`
Or do we have that convention to end with `?` all bool-returning functions in phel? 🤔

jenshaase:
I would like to have the conventions that all functions that return a boolean, have `?` and the end. 

I think I mostly did that but I didn't wrote it down yet. It is on my list. Therefore, I would like to have a `adrs/phel` directory.

Chemaclass:
Perfect. Then you can merge this one :)

jenshaase (approved):

jenshaase:
Ok merged! Thank you!

PR #162, opened 2021-01-10
Adding argv to the core

Chemaclass:
## 📚 Description

🍨 Issue: https://github.com/jenshaase/phel-lang/issues/161 

Be able to use [$argv](https://www.php.net/manual/en/reserved.variables.argv.php) from phel 

## 🔖 Changes

- Added `argv` as constant in `core.phel`.

## 💡 Extra

As a follow-up idea we could develop a library to parse key-value arguments. Something like
```bash
 php phel run src/phel/local.phel k1=v1 k2=v2
```

And (parse-key-values argv), as simple example:
```
# (php/dump (parse-key-values argv))

array:5 [
  0 => "phel"
  1 => "run"
  2 => "src/phel/local.phel"
  'k1' => "v1"
  'k2' => "v2"
]
```

### DIFF
--- a/doc/content/documentation/api.md
+++ b/doc/content/documentation/api.md
@@ -137,6 +137,10 @@ returns logical false, and returns that value and doesn't evaluate any of the
 other expressions, otherwise it returns the value of the last expression.
 Calling the and function without arguments returns true.
 
+## `argv`
+
+Array of arguments passed to script.
+
 ## `array`
 
 ```phel
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -28,6 +28,10 @@
 
 (def *ns* "\\phel\\core")
 
+(def argv
+  "Array of arguments passed to script."
+  (php/aget php/$GLOBALS "argv"))
+
 # --------------------------------------------
 # Basic methods for quasiquote and destructure
 # --------------------------------------------


jenshaase:
I think this doesn't help very much with the `run` command. If I run
```
vendor/bin/phel run src/my-script.phel first-arg second-arg
```
I would expect the arguments to be `src/my-script.phel`, `first-arg` and `second-arg`.

What if I the Phel cli would provide additional flags (e.g. compiler flags). I wouldn't expect them in the `argv` list. For example I run
```
vendor/bin/phel -someflag run src/my-script.phel first-arg second-arg
```
then I would expect the arguments to be `src/my-script.phel`, `first-arg` and `second-arg`.

What if I ran this on the compiled code
```
php src/my-script-compiled.php first-arg second-arg
```
then I would expect the arguments to be `src/my-script.phel`, `first-arg` and `second-arg`.

I have no good solution for that now. But we should investigate more time on that topic.

Chemaclass:
Well, the purpose of the current PR is to be able to use `argv` as a mirror for the `$argv`'s PHP. We can surely think more later and expand this functionality, even providing some filtering for flags or even key-value options. But for now, I thought this `argv` would be pretty good as a start point.

jenshaase:
Ok you are right! Merged it!

PR #166, opened 2021-01-11
Adding scrutinizer to the readme

Chemaclass:
## 🔖 Changes

- Added scrutinizer badge to the readme


jenshaase:
Could we add the test coverage as well (from scrutinizer). That would be more important to me.

jenshaase:
Merged.

PR #167, opened 2021-01-13
Fix Printer (object types)

Chemaclass:
## 📚 Description

I found a bug in the Printer when I tried to print a struct. The conditional was asking for `AbstractStruct::class === $printerName`, but that's make no sense. The struct is a child of that AbstractStruct, therefore we should check if it's an `instance of`.

## 🔖 Changes

- Fixed the conditional bug for object types.
- Cleanup a bit the `Printer` class
- Fix another minor bug: replace `float` to `double`  
  - Reason: For historical reasons "double" is returned in case of a float, and not simply "float" ([source](https://www.php.net/manual/en/function.gettype.php))

### DIFF
--- a/src/php/Printer/Printer.php
+++ b/src/php/Printer/Printer.php
@@ -26,6 +26,7 @@ use Phel\Printer\TypePrinter\SymbolPrinter;
 use Phel\Printer\TypePrinter\TablePrinter;
 use Phel\Printer\TypePrinter\TuplePrinter;
 use Phel\Printer\TypePrinter\TypePrinterInterface;
+use RuntimeException;
 
 final class Printer
 {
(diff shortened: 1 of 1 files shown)


src/php/Printer/Printer.php
```diff
@@ -53,49 +54,64 @@ private function __construct(bool $readable)
...
             return new PhelArrayPrinter($this);
         }
-        if (AbstractStruct::class === $printerName) {
-            return new StructPrinter($this);
-        }
-        if (Table::class === $printerName) {
+        if ($form instanceof Table) {
             return new TablePrinter($this);
         }
+        if ($form instanceof AbstractStruct) {
```
jenshaase, line 97: I think you need to test that before the `Table` since `AbstractStruct` extends `Table`

jenshaase (changes requested):

src/php/Printer/Printer.php
Chemaclass: That's why `Table` condition is before and `AbstractStruct` is after.

jenshaase: Yes you are correct!

jenshaase:
It is merged now!

PR #170, opened 2021-01-13
Remove array_merge from loop

JesusValera:
## 📚 Description

I noticed [PHP Extensions (EA Extended)](https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-) plugin was notifying it's not recommended to use `array_merge` function inside loops.
If you think, it makes sense, you are overriding the same array with the same data plus the new value over and over for each loop iteration.

image

## 🐘  Reference
 - [Never Use array_merge in a Loop | Chemaclass](https://medium.com/swlh/never-use-array-merge-in-a-loop-74917b056ff9)


Chemaclass (approved):
Exactly! 🥇

jenshaase (approved):

jenshaase:
Nice! It is merged now!

PR #171, opened 2021-01-13
Added format + printf phel functions

Chemaclass:
## 📚 Description

Added some more `Phel` functions to the core.

## 🔖 Changes

- Added `format`: Returns a string produced according to the formatting string fmt.
- Added `printf`: Output a formatted string.

Screenshot 2021-01-14 at 08 00 52

Screenshot 2021-01-14 at 08 01 48

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1376,14 +1376,14 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
   nil)
 
 (defn format
-  "Returns a string produced according to the formatting string fmt."
+  "Returns a formatted string. See PHP's [sprintf](https://www.php.net/manual/en/function.sprintf.php) for more information."
   [fmt & xs]
   (apply php/sprintf fmt xs))
 
 (defn printf
-  "Output a formatted string."
+  "Output a formatted string. See PHP's [printf](https://www.php.net/manual/en/function.printf.php) for more information."
   [fmt & xs]
-  (php/print (apply format fmt xs))
+  (apply php/printf fmt xs)
   nil)
 
 # ----------------
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1375,6 +1375,17 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
   (php/print "\n")
   nil)
 
+(defn format
+  "Returns a string produced according to the formatting string fmt."
+  [fmt & xs]
+  (apply php/sprintf fmt xs))
+
+(defn printf
+  "Output a formatted string."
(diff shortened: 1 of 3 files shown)


src/phel/core.phel
```diff
@@ -1375,6 +1375,17 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
   (php/print "\n")
   nil)
 
+(defn format
+  "Returns a string produced according to the formatting string fmt."
```
jenshaase, line 1379: Could we reference the PHP Doc here? Like this:

"Returns a formatted string. See PHP's `sprintf` for more information`"

src/phel/core.phel
```diff
@@ -1375,6 +1375,17 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
...
+  [fmt & xs]
+  (apply php/sprintf fmt xs))
+
+(defn printf
+  "Output a formatted string."
```
jenshaase, line 1384: Same as above

jenshaase (changes requested):

tests/phel/test/core.phel
```diff
@@ -933,6 +933,24 @@
...
+  (is (= "[\"a\" \"b\"]" (format "%s" ["a" "b"])) "format with tuple of strings")
+  (is (= "@[\"a\" \"b\"]" (format "%s" @["a" "b"])) "format with array of strings")
+  (is (= "@{\"a\" \"b\"}" (format "%s" @{"a" "b"})) "format on table")
+  (is (= "(set \"a\" \"b\")" (format "%s" (set "a" "b"))) "format on set")
+  (is (= "x" (format "%s" 'x)) "format on symbol")
+  (is (= ":test" (format "%s" :test)) "format on keyword")
+  (is (= "1" (format "%d" 1)) "format on number")
+  (is (= "1" (format "%d" true)) "format on true")
+  (is (= "0" (format "%d" false)) "format on false")
+  (is (= "" (format "%s" nil)) "format on nil"))
```
Chemaclass, line 949: @jenshaase, what do you think about this test suite? I copy almost all scenarios from `test-print-str`, but I am not sure if it makes sense. Would you suggest removing some or adjust them somehow?

jenshaase: I only would add a `(format "")` test case.

jenshaase (changes requested):

tests/phel/test/core.phel
Chemaclass: Done

jenshaase (approved):
Super!

PR #172, opened 2021-01-13
Refactor StringPrinter class

JesusValera:
## 📚 Description

Refactoring the `StringPrinter`, maybe there is a better approach to improve this class @Chemaclass. 😄 

## 🏷️  Changes

- Replace `switch` by `if`s
- Move the `special characters` to an array
- Replace conditions by a method name

### DIFF
--- a/src/php/Printer/TypePrinter/StringPrinter.php
+++ b/src/php/Printer/TypePrinter/StringPrinter.php
@@ -9,6 +9,18 @@ namespace Phel\Printer\TypePrinter;
  */
 final class StringPrinter implements TypePrinterInterface
 {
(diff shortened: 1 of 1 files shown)


jenshaase (commented):
I don't like it. For example `isMask110XXXXX` is called three times, but one time would be enough.

JesusValera:
> I don't like it. For example `isMask110XXXXX` is called three times, but one time would be enough.

I know, I have to rethink it tomorrow.
That's why I put it in draft. :)

Chemaclass (approved):

jenshaase (approved):

jenshaase:
Thank you. I merged it

PR #177, opened 2021-01-16
Rename php.dockerfile to Dockerfile

JesusValera:
## 📚 Description

It's a minor change, but I think it's redundant and much more clear if we rename `php.dockerfile` to simply `Dockerfile`.
What do you think?

(I updated also the reference from `docker-compose.yml`, and reinstall the image to check the Docker still running) 😛


Chemaclass (approved):

jenshaase (approved):

jenshaase:
Merged!

PR #178, opened 2021-01-17
Adding unit tests to TextExceptionPrinter

Chemaclass:
## 📚 Description

I did some refactoring for the "Exception Printers" in order to be able to write some unit tests for them.

## 🔖 Changes

- Created an `ExceptionPrinterTrait` which contains the duplicated functions from both ExceptionPrinters.
- Created some interfaces in order to decouple the ExceptionPrinters from their direct dependencies.

I was not able to unit test the `FilePositionExtractor` for "SourceMapComment" case (See it in `FilePositionExtractorTest`). There is still some coupling there with the `SourceMapConsumer`, but I didn't want to go deeper. We might go there in another iteration. I think this one is more than good enough for a PR itself :)


jenshaase:
I'm unsure about the naming of `CommentExtractor`. It extracts basically the original file name and the source map data from the file. Maybe a better name would be `SourceMapExtractor`. What do you think?

Chemaclass:
@jenshaase, right. I will rename it, thanks

src/php/Exceptions/Extractor/CommentExtractorInterface.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Exceptions\Extractor;
+
+use Phel\Exceptions\Extractor\ReadModel\ExtractedComment;
+
+interface CommentExtractorInterface
```
jenshaase, line 9: The interface needs to be renamed too.

src/php/Exceptions/Extractor/CommentExtractorInterface.php
```diff
@@ -0,0 +1,12 @@
...
+{
+    public function getExtractedComment(string $fileName): ExtractedComment;
```
jenshaase, line 11: And the method name.

src/php/Exceptions/Extractor/ReadModel/ExtractedComment.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Exceptions\Extractor\ReadModel;
+
+final class ExtractedComment
```
jenshaase, line 7: This should be renamed to `SourceMapInformation`

jenshaase (changes requested):
Sorry, I probably was a bit unclear in my last comment :) 
But can you also please rename the interface and method names :)

jenshaase (approved):

jenshaase:
Look good now. Thank you for that!

PR #180, opened 2021-01-19
Clean up ParserNodes

Chemaclass:
## 🔖 Changes

- Minor clean up for all ParserNodes
- Added `declare_strict_types` phpcs rule


jenshaase (changes requested):
Shouldn't we add a PHP CS Fixer rule for `strict_types`. (`'declare_strict_types' => true`)

jenshaase (approved):

jenshaase:
Thanks. It is merged now!

JesusValera:
Does it work?
I thought there were some tests which extend directly from some classes, and it was not possible to add to all the classes. 🤔

PR #185, opened 2021-01-21
Add infection to composer

JesusValera:
## 📚 Description

`infection` was not "enabled" in the composer scripts because it was not working.

The problem was I/we missed to add the `XDEBUG_MODE=coverage` flag to make it works. With the flag, `infection` works like a charm.

[Here is the issue](https://github.com/infection/infection/issues/1473#event-4233608893) I opened in the infection project about how to solve this problem. 😃


Chemaclass (approved):
Thanks for that :)

PR #188, opened 2021-01-22
Fix errors from the printers + catch potential error in repl

Chemaclass:
## 📚 Description

https://github.com/jenshaase/phel-lang/issues/187 Prevent REPL from being closed on Exception

## 🔖 Changes

- Add a global try-catch block in the REPL loop to ensure it won't be close by any error. Instead, it will output the error message.
- Fix the ExceptionPrinters; the `$frame['args']` aren't present always, so using the `?? []` would be sufficient to avoid triggering a new error while rendering the actual "current error" ;)

## 🖼️  Screenshots

Screenshot 2021-01-22 at 17 10 17

## QA
Before this was exiting the REPL:
```clojure
(print (\readline_info "library_version"))
```
With this PR it doesn't exit the REPL anymore.

### DIFF
--- a/src/php/Command/ReplCommand.php
+++ b/src/php/Command/ReplCommand.php
@@ -66,7 +66,7 @@ final class ReplCommand
                     break;
                 }
             } catch (Throwable $e) {
-                $this->io->output($e->getMessage() . PHP_EOL);
+                $this->io->output($e->getTraceAsString() . PHP_EOL);
             }
         }
 
--- a/src/php/Command/ReplCommand.php
+++ b/src/php/Command/ReplCommand.php
@@ -58,7 +58,12 @@ final class ReplCommand
             $this->io->output("\e[?2004h"); // Enable bracketed paste
             $input = $this->io->readline('>>> ');
             $this->io->output("\e[?2004l"); // Disable bracketed paste
-            $this->checkInputAndAnalyze($input);
+
+            try {
+                $this->checkInputAndAnalyze($input);
+            } catch (Throwable $e) {
+                $this->io->output($e->getMessage());
+            }
         }
     }
 
--- a/src/php/Command/ReplCommand.php
+++ b/src/php/Command/ReplCommand.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Phel\Command;
 
-use Phel\Command\Repl\ColorStyle;
+use Phel\Command\Repl\ColorStyleInterface;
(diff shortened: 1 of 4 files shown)


JesusValera (approved):
🥇  That's true, I also had this problem in the past. 😄

src/php/Command/ReplCommand.php
```diff
@@ -55,24 +55,31 @@ public function run(): void
...
+                $input = $this->io->readline(self::PROMPT);
+                $this->io->output(self::DISABLE_BRACKETED_PASTE);
+
+                try {
+                    $this->checkInputAndAnalyze($input);
+                } catch (ExitException $e) {
+                    break;
+                }
+            } catch (Throwable $e) {
+                $this->io->output($e->getMessage() . PHP_EOL);
```
jenshaase, line 69: Is the message enough or would it be better to show the complete stack trace?

mabasic: Full stack trace would be cool for debugging purposes... but a message is also fine... idk

Chemaclass: Yeah, you're right. `getTraceAsString` is nicer than `getMessage`, I will change it :)

src/php/Command/ReplCommand.php
```diff
@@ -55,24 +55,31 @@ public function run(): void
...
+                $input = $this->io->readline(self::PROMPT);
+                $this->io->output(self::DISABLE_BRACKETED_PASTE);
+
+                try {
+                    $this->checkInputAndAnalyze($input);
+                } catch (ExitException $e) {
+                    break;
+                }
+            } catch (Throwable $e) {
+                $this->io->output($e->getTraceAsString() . PHP_EOL);
```
jenshaase, line 69: Maybe I'm wrong. But `getTraceAsString` only returns the stack trace, but not the message? Both would be good :)

jenshaase (changes requested):

src/php/Command/ReplCommand.php
Chemaclass: This is what you see with the current code. I think the error is there:  `PHP Notice: undefined index ... etc`

Screenshot 2021-01-22 at 17 26 38

jenshaase: Ok. Thanks

jenshaase (approved):

jenshaase:
It is merged now! Thank you.

Issue #184, opened 2021-01-21
Allow multiline REPL expressions


Chemaclass:
#  User Story

As a user, I want to be able to write an expression in multiple lines in REPL.

## Current behavior

If you try to write an expression without the ending parenthesis because you want to keep writing in another line, you get an error like the following:
```clojure
>>> (+ 1 2
Unterminated list
in string:1

1| (+ 1 2
   ^^^^^^
```

## Acceptance criteria

Don't finish the parsing/compilation of an expression by one line, but by the ending parenthesis. 
- Good number of closing parenthesis -> execute parsing and compilation.

```clojure
>>> (+ 1 2
3)
6
```

### Error consideration

When a higher number of closing parenthesis:

```clojure
>>> (+ 1 2
3))
6
Syntax error reading source at (REPL:2:4).
Unmatched delimiter: )
```

#### HInt

Keep reading lines from the prompt until you find the proper number of closing parenthesis or higher

Chemaclass:
Close in favor of the PR #191

PR #191, opened 2021-01-23
Allow multiline REPL expressions

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/184

Keep reading lines from the prompt till the number of closing parentheses matches the open ones. 
Then store it in the REPL history and analyze it.

## 🔖 Changes

- Allow multiline REPL expressions
- Created an `InputValidator` to know if the REPL should read more lines or parse them.
- Ignore chars founded after the comment char `#` (per line)

## Output example

``` clojure
>>> (+ 1
... 1
... #)
... 2#)
... )
4
```


JesusValera (approved):
👏🏼

jenshaase:
This will definitely not work. You can't just count `(` and `)`. For example, you can put a bracket inside of string. And what about the other brackets. You also need to make sure that you have matching `"`.

I don't now what is the best way to make this feature. One solution would be to check after every readline if the input buffer is parsable. But that you have to find away to exit the input buffers.

Chemaclass:
Right, right. I got another idea. I will try something else

src/php/Command/Repl/InputValidator.php
```diff
@@ -0,0 +1,111 @@
...
+        }
+    }
+
+    /**
+     * @throws RuntimeException
+     */
+    private function validateNumberOfBrackets(): void
+    {
+        if ($this->totalCloseRoundBrackets > $this->totalOpenRoundBrackets) {
+            throw new RuntimeException('Wrong number of parentheses');
```
jenshaase, line 93: Invalid input shouldn't be handled here. The error messages of the parser are much better.

Chemaclass: RIght. I remove all exceptions from here. Instead, I am checking that the total closed are >= than the opened ones:
```php
$this->totalCloseParentheses >= $this->totalOpenParentheses
```
in such a case, it will be ready to be parsed, and in case there is any parser error, then we will display the proper parser error.

src/php/Command/ReplCommand.php
```diff
@@ -21,102 +20,123 @@ final class ReplCommand
...
-                }
+                $this->addLineFromPromptToBuffer();
+                $this->checkExitInputBuffer();
+                $this->analyzeInputBuffer();
+            } catch (ExitException $e) {
+                break;
             } catch (Throwable $e) {
-                $this->io->output($e->getTraceAsString() . PHP_EOL);
+                $this->inputBuffer = [];
+                $this->io->output($this->style->red($e->getMessage() . PHP_EOL));
```
jenshaase, line 70: Why are we back to `->getMessage()` only. Shouldn't it be `->getTraceAsString()`?

src/php/Command/ReplCommand.php
```diff
@@ -21,102 +20,123 @@ final class ReplCommand
...
+        }
+
+        $prompt = empty($this->inputBuffer) ? self::INITIAL_PROMPT : self::OPEN_PROMPT;
+        $input = $this->io->readline($prompt);
+
+        if ($this->io->isBracketedPasteSupported()) {
+            $this->io->output(self::DISABLE_BRACKETED_PASTE);
+        }
+
+        if ($input === null) {
```
jenshaase, line 90: If `$input === null` the user pressed `CTRL+D`.
* If I'm in multiline mode I would expect to exit the multiline mode and start with a new prompt `>>> `
* If I'm not in multiline mode the REPL should exit.

Maybe this is implemented like this. But I can't see it.

The `exit` command is only handle in non multiline mode. 
* `>>> exit` will stop the REPL
* `... exit` will not stop the REPL. `exit` is then interpreted as a Phel symbol.

src/php/Command/ReplCommand.php
```diff
@@ -21,102 +20,123 @@ final class ReplCommand
...
-        if ('' === $input) {
+    private function analyzeInputBuffer(): void
+    {
+        if ('' === end($this->inputBuffer)) {
             return;
         }
 
-        $this->io->addHistory($input);
+        $inputAsString = implode(PHP_EOL, $this->inputBuffer);
+        $this->io->addHistory($inputAsString);
```
jenshaase, line 116: For the example:
```
>>> (
... +
... 1
... 1)
```

The history contains following entries:
* `(`
* `(\n+`
* `(\n+\n1`
* `(\n+\n1\n1)`

Is that behavior correct? (Sorry not tested)

jenshaase (changes requested):

src/php/Command/ReplCommand.php
Chemaclass: Right, it shouldn't be like that. I will change it to add to the history ONLY after being successfully parsed.
In your example:
```
>>> (
... +
... 1
... 1)
```
it should store in the history only one line:
```
(\n+\n1\n1)
```

src/php/Command/ReplCommand.php
Chemaclass: Exactly. This is the intended behavior.

src/php/Command/ReplCommand.php
Chemaclass: Ups, right, I think I missed that. I will put the trace back

src/php/Command/ReplCommand.php
Chemaclass: Nope, with is correct. We store the plain string as 
```php
$input = implode(PHP_EOL, $this->inputBuffer); // -> (\n+\n1\n1)
```

src/php/Command/ReplCommand.php
jenshaase: I tested it now. But it is not working. My input was this:
```
>>> (+
... 1
...
```
The I pressed `CTRL+D` and this happens:
```
>>> (+
... 1
... ...
```

After pressing `CTRL+D` multiple times:
```
>>> (+
... 1
... ... ... ... ... ... ... ... ... ... ... ... ...
```

If i insert this (last insert was `2`, no `enter`)
```
>>> (+
... 1
... 2
```
and press `CTRL+D` nothing happens

jenshaase (changes requested):

src/php/Command/ReplCommand.php
Chemaclass: Right! I was using Ctrl+C all the time, so I didn't properly test this one 🙈 
It should be fixed now. Thanks for checking it.

Chemaclass: Actually, regarding your last scenario:
```
If i insert this (last insert was 2, no enter)

>>> (+
... 1
... 2
and press CTRL+D nothing happens
```
Yes, the control+d scenario is really tricky, because of how it's implemented by design.
`readline()` returns false (and we cast it to null, see ReplCommandSystemIo:28) when "there is no more data to read, then FALSE is returned." (https://php.net/manual/en/function.readline.php)

I think it's ok like it is right now. Ctrl+C will exit the execution for sure always :)
Anyway, if you have another idea or suggestion on how to do that, please, don't hesitate. Anyway, we can always do a follow up PR with some other improvements later on.

jenshaase (changes requested):
Found some other miss behavior but I think it is almost done.

src/php/Command/ReplCommand.php
jenshaase: Ah that is ok for me.

src/php/Command/ReplCommand.php
```diff
@@ -21,102 +20,125 @@ final class ReplCommand
...
+
+        if (self::EXIT_REPL === $firstInput) {
+            throw new ExitException();
+        }
+    }
+
+    private function analyzeInputBuffer(): void
     {
+        if ('' === end($this->inputBuffer)) {
+            array_pop($this->inputBuffer);
```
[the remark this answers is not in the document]
Chemaclass: This is to remove the latest introduced input when it's an empty string. This means, to remove the empty lines that you write while being filling the "inputBuffer" thing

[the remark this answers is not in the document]
jenshaase: Ok, I see. But it is not required because the parser can still parse it (just whitespace). Why would you remove those empty lines? Maybe the user just wants to have empty lines.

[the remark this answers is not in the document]
Chemaclass: Because I didn't want to have them, actually 😁
Try it locally, and see it yourself, that you don't really want those empty lines in history. Because for the parser doesn't matter. They affect the history, and how it's saved. 
I took that decision based on the experience. Let me know what do you think. I could change it... but I think this is a good idea that improves the usability of the repl. 
I suggest keep it, and if we face some problems or we change our mind, we simply modify it later :)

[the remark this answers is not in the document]
jenshaase: Ok cool. Then I will merge it now.

jenshaase (approved):

jenshaase:
Merged now. Super cool to have this multiline input!

Chemaclass:
Now we can update the repl section in the docu mention it. Can you do that, @jenshaase?

jenshaase:
Yes I can. I will also add some more repl integration tests. Then we see if something breaks.

Issue #198, opened 2021-01-24
Improve `phel fmt _DIR_`


JesusValera:
Currently, when we run `php phel` we get the following output:
```
Usage: phel [command]

Commands:
    repl
        Start a Repl.

    run <filename-or-namespace>
        Runs a script.

    test <filename> <filename> ...
        Tests the given files. If no filenames are provided all tests in the
        test directory are executed.

    fmt <filename>
        Formats the given files.

    help
        Show this help message.
```

I suggest
- `phel fmt _DIR_` -> recursively format the `.phel` files inside the specified folder
- `phel fmt` -> recursively browse all dirs from the current folder (¿excluding dirs like `vendor`?)

jenshaase:
It should be possible to define multiple files and folders. For example
```
phel fmt src/phel tests/phel path/to/file.phel
```

jenshaase:
Fixed with #199

PR #199, opened 2021-01-24
Improve FormatCommand

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/198 

Recursively format the `.phel` files inside the specified folder.
It should be possible to define multiple files and folders.

## 🔖 Changes

- Add multiple paths (file or directory) for the `phel fmt` command
  - For example: `phel fmt src/file-1.phel src/file-2.phel`

- Recursive directory reading
  - For example: `phel fmt src/`

- The command will output the list of files which were formatted.
  - For example:
```
➜ php phel fmt src
Formatted files:
  1) src/phel/core.phel 
  2) src/phel/local.phel 
➜ php phel fmt src
No files were formatted.
```


src/php/Command/FormatCommand.php
```diff
@@ -4,23 +4,75 @@
...
+
+            if ($wasFormatted) {
+                $this->filePaths[] = $fileOrPath;
+            }
+        }
+    }
+
+    private function runFormatterInDirectory(string $directory): void
+    {
+        $paths = array_diff(scandir($directory), ['..', '.']);
```
jenshaase, line 54: This is not looking for files in subfolders. I would recommend this approach:
```
$iterator = new RecursiveIteratorIterator(
  new RecursiveDirectoryIterator($directory),
  RecursiveIteratorIterator::LEAVES_ONLY
);
```

src/php/Command/FormatCommand.php
```diff
@@ -4,23 +4,75 @@
...
+        FormatterInterface $formatter,
+        CommandIoInterface $io
+    ) {
         $this->formatter = $formatter;
+        $this->io = $io;
+    }
+
+    public function run(array $paths): void
+    {
+        foreach ($paths as $fileOrPath) {
```
jenshaase, line 32: Wouldn't it be better to first collect all files? Then you can filter out duplicates. For example if i run
```
phel fmt src/file-1.phel src/file-1.phel
```
This will format the file twice and not once.

jenshaase (changes requested):

src/php/Command/FormatCommand.php
Chemaclass: Thanks, I updated the code to use your suggestion.

src/php/Command/FormatCommand.php
Chemaclass: I improved this part filtering the inputs (by unique values) and also saved in a variable the already parsed files to avoid run the fmt twice on the same files at any chance.

src/php/Command/FormatCommand.php
```diff
@@ -4,23 +4,91 @@
...
         $this->formatter = $formatter;
+        $this->io = $io;
+    }
+
+    /**
+     * @param list<string> $paths
+     */
+    public function run(array $paths): void
+    {
+        foreach (array_unique($paths) as $path) {
```
jenshaase, line 40: Not fully of what I meant. Maybe it was a bad example. The next example will also run the same file twice:
```
phel fmt src/ src/test.phel
```
It would be better to have a function that takes `$paths` and returns a list of all files. On this list you can do `array_unique` and maybe also sort it, so the execution order is always the same.

src/php/Command/FormatCommand.php
```diff
@@ -4,23 +4,91 @@
...
+    }
+
+    private function runInPath(string $path): void
+    {
+        if (is_dir($path)) {
+            $this->runFormatterInDirectory($path);
+            return;
+        }
+
+        if (isset($this->alreadyTriedToFormatFilePath[$path])) {
```
jenshaase, line 54: With my approach above you don't need it

jenshaase (changes requested):

src/php/Command/FormatCommand.php
Chemaclass: Ah, got it. Ok. I am on it

src/php/Command/Format/PhelPathFilter.php
```diff
@@ -0,0 +1,59 @@
...
+
+        foreach ($paths as $path) {
+            if (is_file($path)) {
+                if ($this->hasPhelExtension($path)) {
+                    $returnPaths[] = $path;
+                }
+                continue;
+            }
+
+            foreach ($this->createRecursiveIterator($path) as $fileInfo) {
```
jenshaase, line 31: Is it save to say if it is not a file it is a directory?

* `is_file`: Returns true if the filename exists and is a regular file, false otherwise.
* `is_dir`: Returns true if the filename exists and is a directory, false otherwise. 

If I add a path that doesn't exists, what will happen?

jenshaase (changes requested):

src/php/Command/Format/PhelPathFilter.php
Chemaclass: I didn't test that case, you are right. I updated the condition to this logic:
```php
if (is_file($path)) {
    // ...
} elseif (is_dir($path)) {
    // ...
}
```

jenshaase (approved):
Looks good now.

PR #200, opened 2021-01-26
Statement number in REPL

Chemaclass:
## 🔖 Changes

- Added the statement number next to the REPL prompt
- Added multiline REPL integration tests
- Changed `output()` to `write()` and `writeln()` from `ReplCommandIoInterface`

## 🖼️  Screenshots

Screenshot 2021-01-26 at 17 26 36


jenshaase:
This is good. But the error messages of the parser, analyzer, etc. need to match the line numbers of the REPL than. Currently it looks like this:
```
phel(3)> ({}})
Unexpected token: {
in string:1

1| ({
    ^
```

But it should look like this:
```
phel(3)> ({}})
Unexpected token: {
in string:3

3| ({
    ^
```

I also would like to see the following prompt (colon instead of brackets). 
```
phel:1>
```
This is more consistent to the PHP and Phel errors messages. These are encoded like `<file>:<line>`.

jenshaase (approved):
Very cool!

jenshaase:
Thank you very much. I merged it now.

Issue #197, opened 2021-01-24
Set a color for the REPL output depending on the type


JesusValera:
## 📚  Description

I took this picture from elixir.

![image](https://user-images.githubusercontent.com/6381924/105629639-7940a980-5e44-11eb-9395-92ccd4a04add.png)

I like the idea of colour the output from the REPL console depending on the type.

What do you think? What colours should we use for each element type?

jenshaase:
Yes this can be added to the Printer. However, this should not be enabled by default, since the `print` function in the core library also uses the `Printer`.

First lets define, how many colors we need and then what colors there are. The colors should have high contrast on both white and black backgrounds.

jenshaase:
I would suggest the following colors:

* nil `\033[0;96m%s\033[0m`
* string `\033[0;95m%s\033[0m`
* boolean `\033[0;94m%s\033[0m`
* symbol `\033[0;91m%s\033[0m`
* number `\033[0;92m%s\033[0m`
* Keyword `\033[0;93m%s\033[0m`

Test script
```
<?php

echo sprintf("\033[0;96m%s\033[0m", "nil") . PHP_EOL;
echo sprintf("\033[0;95m%s\033[0m", "\"string\"") . PHP_EOL;
echo sprintf("\033[0;94m%s\033[0m", "true") . PHP_EOL;
echo sprintf("\033[0;91m%s\033[0m", "sym") . PHP_EOL;
echo sprintf("\033[0;92m%s\033[0m", "1234") . PHP_EOL;
echo sprintf("\033[0;93m%s\033[0m", ":keyword") . PHP_EOL;
```

PR #202, opened 2021-01-27
Basic colors to repl printers

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/197
Pass a printer with colors to the REPL printer, so it can handle the colors for each individual type printer.

## 🔖 Changes

- Add colors to REPL printer
- By default, all TypePrinters are without color
- Added `print-colorful` to `core/repl.phel`, because `println` wont render any color.

## 🖼️  Screenshot

Screenshot 2021-01-27 at 18 32 48

When using `(print ...)` function doesn't render the colors because print uses print-str, and there we are using a nonReadable printer without colors:  ```printer (php/:: Printer (nonReadable))```. The problem here is that `print-str`  is the "same as print. But instead of writing it to an output stream. The resulting string is returned."

```clojure
# see phel/core/print-operation:15
(defn print-str
  "Same as print. But instead of writing it to an output stream,
  The resulting string is returned."
  [& xs]
  (let [len (count xs)
        printer (php/:: Printer (nonReadable))
        pp |(php/-> printer (print $))]
    (case (count xs)
...
```
Screenshot 2021-01-27 at 18 41 29

For this reason, I created a new function `print-colorful` which will be loaded only for the REPL and it does the same as `println` but with color.
Screenshot 2021-01-27 at 19 47 08

### DIFF
--- a/src/phel/repl.phel
+++ b/src/phel/repl.phel
@@ -1,7 +1,8 @@
 (ns phel\repl
   (:use Phel\Lang\Symbol)
   (:use Phel\Runtime\RuntimeFactory)
-  (:use Phel\Compiler\Analyzer\Environment\NodeEnvironment))
+  (:use Phel\Compiler\Analyzer\Environment\NodeEnvironment)
+  (:use Phel\Printer\Printer))
 
 (defn- get-runtime []
   (php/:: RuntimeFactory (getInstance)))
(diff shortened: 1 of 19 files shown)


jenshaase (approved):
Very nice!

PR #203, opened 2021-01-28
fix composer require

aszenz:


Chemaclass (approved):
Yes, we have to update the getting started guide. Thanks for the fix!

Issue #204, opened 2021-01-28
The repl cannot print all values


aszenz:
The repl cannot print all kinds of values, for some values the internal class name is shown, for others an exception is thrown
Examples:

```
(php/explode "-" "hi-bye") # shows <PHP-Array>

(fn [x y] (+ x y)) # throws exception

(php/new \DateTime "now") # throws exception
```

jenshaase:
Closed with #206

PR #206, opened 2021-01-28
Fix repl cannot print all values

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/204
In order to avoid exceptions when printing anything from the REPL, I created 3 new printers.

## 🔖 Changes

- Created `WithColorTrait` for the Printers.

- Print one level of data from the `ArrayPrinter` instead of displaying only `<PHP-Array>`. 
- Created new printers:
  - `ToStringPrinter`: it calls the `__toString()` in case the object has it.
  - `AnonymousClassPrinter`: it notifies you that that class is an anon class, that's all.
  - `NonPrintableClassPrinter`: it tells you that the class that you are trying to print is not printable.

## 🖼️ Screenshots

Screenshot 2021-01-28 at 17 32 18

### DIFF
--- a/src/php/Printer/TypePrinter/ArrayPrinter.php
+++ b/src/php/Printer/TypePrinter/ArrayPrinter.php
@@ -4,12 +4,21 @@ declare(strict_types=1);
 
 namespace Phel\Printer\TypePrinter;
 
+use Phel\Printer\PrinterInterface;
+
 /**
  * @implements TypePrinterInterface<array>
  */
 final class ArrayPrinter implements TypePrinterInterface
 {
-    use WithColorTrait;
+    private PrinterInterface $printer;
+    private bool $withColor;
+
+    public function __construct(PrinterInterface $printer, bool $withColor = false)
+    {
+        $this->printer = $printer;
+        $this->withColor = $withColor;
+    }
 
     /**
      * @param array $form
@@ -54,7 +63,7 @@ final class ArrayPrinter implements TypePrinterInterface
             return sprintf('"%s"', $v);
         }
 
-        return (string)$v;
+        return $this->printer->print($v);
     }
 
     private function color(string $str): string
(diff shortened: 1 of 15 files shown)


src/php/Printer/TypePrinter/ArrayPrinter.php
```diff
@@ -4,13 +4,65 @@
...
+    /**
+     * @param mixed $v
+     */
+    private function formatValue($v): string
+    {
+        if (is_string($v)) {
+            return sprintf('"%s"', $v);
+        }
+
+        return (string)$v;
```
jenshaase, line 57: * If `$v` is `true` it will print `1` and not `true`.
* If `$v` is `null` it will print an empty string and not `null`.
* What is a PHP Array contains a Phel Array. Then this is not printed correctly.

I think this method must be recursively call the main Printer?

jenshaase (changes requested):

src/php/Printer/TypePrinter/ArrayPrinter.php
Chemaclass: Good point. I just changed 👌🏻

src/php/Printer/TypePrinter/ArrayPrinter.php
```diff
@@ -4,13 +4,74 @@
...
+            $form
+        );
+    }
+
+    /**
+     * @param mixed $v
+     */
+    private function formatValue($v): string
+    {
+        if (is_string($v)) {
```
jenshaase, line 62: Is this required? Can't we just use the String printer?

jenshaase (changes requested):

src/php/Printer/TypePrinter/ArrayPrinter.php
Chemaclass: of course we can 🙈 I will change it sofort!

jenshaase:
Thanks It is merged now!

Issue #205, opened 2021-01-28
Errored out expressions are not shown in repl history


aszenz:
Using the up arrow in repl only traverses expressions that evaluated successfully, it's hard to fix an errored out expressions since it's not recorded in repl history

Example:
```
Type "exit" or press Ctrl-D to exit.
phel:1> (first [1 2 3])
1
phel:2> ( firsts [2])
Can not resolve symbol 'firsts'
in string:2

2| ( firsts [2])
     ^^^^^^
phel:3> (first [1 2 3]) # The up arrow shows the last successfully executed expression instead of the last written expression
```

Chemaclass:
I did this intentionally. But now that you are reporting this issue, it makes me think that maybe it would be good as well to include in the repl-history all given inputs, and not only the successful ones 🤔 What do you think, @jenshaase?

jenshaase:
@aszenz is correct. We should add these to the history!

PR #207, opened 2021-01-28
Save in history incorrect REPL expressions

JesusValera:
## 📚 Description

Closes https://github.com/phel-lang/phel-lang/issues/205

When you press `<ENTER>` in the REPL, even if the statement is incorrect, it will be saved inside the `phel-repl-history` file.

## 🖼️  Screenshot

image

### DIFF
--- a/src/php/Command/ReplCommand.php
+++ b/src/php/Command/ReplCommand.php
@@ -105,9 +105,9 @@ final class ReplCommand
         }
     }
 
-    private function addHistory(?string $input): void
+    private function addHistory(string $input): void
     {
-        if (!empty($input)) {
+        if ('' !== $input) {
             $this->io->addHistory($input);
         }
     }
--- a/src/php/Command/ReplCommand.php
+++ b/src/php/Command/ReplCommand.php
@@ -101,14 +101,6 @@ final class ReplCommand
             $this->io->writeln();
         } else {
             $this->inputBuffer[] = $input;
-            $this->addHistory($input);
-        }
-    }
-
-    private function addHistory(string $input): void
-    {
-        if ('' !== $input) {
-            $this->io->addHistory($input);
         }
     }
 
@@ -132,6 +124,8 @@ final class ReplCommand
 
         $fullInput = implode(PHP_EOL, $this->inputBuffer);
 
+        $this->addHistory($fullInput);
+
         try {
             $result = $this->compiler->eval($fullInput, $this->lineNumber - count($this->inputBuffer));
             $this->io->writeln($this->printer->print($result));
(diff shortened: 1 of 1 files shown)


src/php/Command/ReplCommand.php
```diff
@@ -101,6 +101,14 @@ private function addLineFromPromptToBuffer(): void
             $this->io->writeln();
         } else {
             $this->inputBuffer[] = $input;
+            $this->addHistory($input);
+        }
+    }
+
+    private function addHistory(?string $input): void
+    {
+        if (!empty($input)) {
```
Chemaclass, line 110:
```suggestion
    private function addHistory(string $input): void
    {
        if ('' !== $input)) {
```

Chemaclass (approved):
Good one!

jenshaase (changes requested):
This is not working correctly in multi line mode. If I type this
```
phel:3> (
....:4> +
....:5> 1
....:6> 3
....:7> )
4
phel:8>
```
The history is `(`, `+`, `1`, `3`, `)` and not
```
(
+
1
3
)
```

JesusValera:
> This is not working correctly in multi line mode. If I type this
> 
> ```
> phel:3> (
> ....:4> +
> ....:5> 1
> ....:6> 3
> ....:7> )
> 4
> phel:8>
> ```
> 
> The history is `(`, `+`, `1`, `3`, `)` and not
> 
> ```
> (
> +
> 1
> 3
> )
> ```

Do you mean, if you press `up`, it should display all those lines in a single one?

jenshaase:
Not in a single row but like this
```
phel:3> (
....:4> +
....:5> 1
....:6> 3
....:7> )
4
phel:8> (
+
1
3
)
```

That's how it was before.

JesusValera:
Sorry for the delay.

Basically, I was storing into the `history` line-by-line, the solution was storing just after imploding all the input buffer. 😄 

@jenshaase @Chemaclass

Chemaclass:
You are storing the inputBuffer every time even when it was Unfinished.  Test this:
1) delete all content from the `.phel-repl-history`
2) exec in repl
```clojure
(defn adder [& xs]
  (apply + xs)
  ]
```
3) the `.phel-repl-history` should have only 1 line

JesusValera:
> You are storing the inputBuffer every time even when it was Unfinished. Test this:
> 
> 1. delete all content from the `.phel-repl-history`
> 2. exec in repl
> 
> ```clojure
> (defn adder [& xs]
>   (apply + xs)
>   ]
> ```
> 
> 1. the `.phel-repl-history` should have only 1 line

That's true!
Thank you for spotting this bug :P

FIXED!

Issue #211, opened 2021-01-29
Comparing arrays


mabasic:
This returns true, but I think that it should not

```
(= @["employee"] @[0])
true
```

jenshaase:
Fixed with #212

PR #212, opened 2021-01-29
Fix comparing arrays

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/211

## 🔖 Changes

- Add some extra logic to `PhelArray->equals()` in order to be able to compare in more detail the internal data from them.

### DIFF
--- a/src/php/Lang/AbstractType.php
+++ b/src/php/Lang/AbstractType.php
@@ -14,6 +14,14 @@ abstract class AbstractType implements MetaInterface, SourceLocationInterface
      */
     abstract public function hash(): string;
 
+    protected function equals1($a, $b): bool {
+        if ($a instanceof AbstractType) {
+            return $a->equals($b);
+        }
+
+        return $a === $b;
+    }
+
     /**
      * Check if $other is equals to $this.
      *
--- a/src/php/Lang/AbstractType.php
+++ b/src/php/Lang/AbstractType.php
@@ -14,7 +14,12 @@ abstract class AbstractType implements MetaInterface, SourceLocationInterface
      */
     abstract public function hash(): string;
 
-    protected function equals1($a, $b): bool {
+    /**
+     * @param mixed|null $a
+     * @param mixed|null $b
+     */
+    protected function equals1($a, $b): bool
+    {
         if ($a instanceof AbstractType) {
             return $a->equals($b);
         }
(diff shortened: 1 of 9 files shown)


src/php/Lang/PhelArray.php
```diff
@@ -138,7 +138,18 @@ public function hash(): string
 
     public function equals($other): bool
```
jenshaase, line 139: Not really sure about the last part of the method. In my opinion this would be enough:

```php
if (!($other instanceof PhelArray)) {
  return false;
}

return $this->data === $other->data;
```

Have you an example where this fails?

Chemaclass: I tried that first, sure, and here is what I got :)
Screenshot 2021-01-29 at 09 05 07

jenshaase: Ufff. Then I should have a look at that. Maybe we just need to call `reset` on both arrays before:

```php
if (!($other instanceof PhelArray)) {
  return false;
}

reset($this->data);
reset($other->data);

return $this->data === $other->data;
```

Chemaclass: Clever try, but still:

Screenshot 2021-01-29 at 09 19 05

jenshaase: Ok, I will have a look

jenshaase:
@Chemaclass Comparison was completely broken. I fixed it now for Arrays, Tuples and Tables. Sets are still missing since I haven't figured out a way to do that yet. Maybe you have an idea.

The main problem was that I wanted every Phel Type to compare with the `equals` Method in `AbstractType`. It only worked if I don't have to do nested comparison. Now I do the nested comparison manually but this is every cost intensive.

What we really need is somethings that is described in this RFC: https://wiki.php.net/rfc/object-comparison
It was rejected. :(

src/php/Lang/AbstractType.php
```diff
@@ -14,6 +14,14 @@ abstract class AbstractType implements MetaInterface, SourceLocationInterface
      */
     abstract public function hash(): string;
 
+    protected function equals1($a, $b): bool {
```
Chemaclass: This feels kind of weird... I will look at it 👀

jenshaase:
I investigated further into the sets. There are even more problems:
```
(set (set 1 2) (set 1 2))
```
will evaluate to
```
(set (set 1 2) (set 1 2))
```
and not to
```
(set (set 1 2))
```

What we need is a better hashing algorithm for Phel data structures. I have an idea for that.

jenshaase:
This is merged now!

Issue #213, opened 2021-01-29
Improve phel test api


Chemaclass:
Currently, we have only `deftest` &  `is` as part of the phel test API.

I would like to have a better API with more expressive functions such as:
- `(is-equals expected actual)`
- `(is-true actual)`
- `(is-false actual)`
- `(is-nil actual)`

Chemaclass:
Wont do.

PR #214, opened 2021-01-29
Improve phel test api

Chemaclass:
## 📚 Description
Related Issue: https://github.com/phel-lang/phel-lang/issues/213

Added some test functions and refactor all phel unit tests.

## 🔖 Changes

- Added `is-equals`, `is-true`, `is-false`, `is-nil` to phel/test module
- Reorganize all phel unit tests mirroring the same structure as we currently have in `src/phel/core`
- Use these new functions in the current phel tests


JesusValera (approved):
🔝 🔝 🔝 
Really love how you handle all the test files 🥇

jenshaase:
The error messages don't work this way. One example:
We had the test:
```
(is (= nil (first [])) "first of empty tuple")
```

If I change this text so if fails
```
(is (= nil (first [1 2 3 4])) "first of empty tuple")
```

The error message is:
```
Failure in 'first of empty tuple' (/home/jens/dev/php/phel-lang/tests/phel/test/core/basic-sequence-operation.phel:16)
          Test: (first [1 2 3 4])
  evaluated to: 1
    but is not: = to nil
```

With your changes the method error looks like this:
```
Failure in 'first of empty tuple' (/home/jens/dev/php/phel-lang/src/phel/test.phel:181)
          Test: actual <---------------- see here
  evaluated to: 1
    but is not: = to nil
```

This only prints `actual` and not `(first [1 2 3 4])`.

---

However, during the time I wrote the test lib I didn't had all the features it now has. What you already can do is to write a test like this:
```
(is (nil? (first [])) "first of empty tuple")
```

If we change it again to fail:
```
(is (nil? (first [1 2 3 4])) "first of empty tuple")
```

The output message is:
```
Failure in 'first of empty tuple' (/home/jens/dev/php/phel-lang/tests/phel/test/core/basic-sequence-operation.phel:16)
                 Test: (first [1 2 3 4])
         evaluated to: 1
  but doesn't satisfy: nil?
```
(After the bug I just discovered)

I think this is much better that having shortcuts for the tests. What do you think?

Chemaclass:
Absolutely right! I didn't see this as well about the `actual` feedback-error that the user receives 🙈
I will rollback and get rid of these is-equals, is-nil, etc, but keeping the dir/file structure (mirroring prod code) ♻️

Chemaclass:
I will close this PR and open a new one only with the splitting of the `phel/test/core.phel` file into diff files mirroring the prod code.

Issue #216, opened 2021-01-30
Display the origin of the potencial error when applying fmt command


Chemaclass:
## Current Behaviour

If there is a malformed file in a directory that you wanted to apply the formatted, the output of running `phel fmt dir` is simply an error like `Unterminated list (BRACKETS)` but it doesn't tell you which line or file got such error.

## Acceptance Criteria

If the fmt encounters an error it should output the origin of that error to be able to trace it back.

Chemaclass:
Done

PR #217, opened 2021-01-30
Use ExceptionPrinter in FormatCommand

Chemaclass:
## 📚 Description
Issue: https://github.com/phel-lang/phel-lang/issues/216
When running the "phel fmt" command, in case the phel file has any parsing error we didn't show much feedback.

## 🔖 Changes

- Include the ExceptionPrinter to the FormatCommand, so in case the formatted encounters any error, it will render the file and exact location where the error was found.

## 🖼️  Screenshots
Here you can see an example of the output when the phel code is malformed (I added the `]` for testing purpose):
Screenshot 2021-01-30 at 19 26 17

### DIFF
--- a/src/php/Command/FormatCommand.php
+++ b/src/php/Command/FormatCommand.php
@@ -39,27 +39,14 @@ final class FormatCommand
      */
     public function run(array $paths): void
     {
-        try {
-            $this->formatPaths($paths);
-        } catch (ParserException $e) {
-            $this->io->writeln(
-                $this->exceptionPrinter->getExceptionString($e, $e->getCodeSnippet())
-            );
-        }
-    }
-
-    /**
-     * @param list<string> $paths
-     *
-     * @throws ParserException
-     */
-    private function formatPaths(array $paths): void
-    {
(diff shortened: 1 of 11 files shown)


src/php/Command/FormatCommand.php
```diff
@@ -15,24 +17,43 @@ final class FormatCommand
...
     }
 
     /**
      * @param list<string> $paths
      */
     public function run(array $paths): void
+    {
+        try {
+            $this->formatPaths($paths);
+        } catch (ParserException $e) {
```
jenshaase, line 44: Isn't it better to handle the exception just around ` $this->formatter->formatFile($path);`. Then all other files can still be formatted?

jenshaase (changes requested):

src/php/Command/FormatCommand.php
Chemaclass: Oh, clever, I didn't see that. I will try it 👌🏻

jenshaase (approved):

PR #219, opened 2021-01-30
Restructure phel core tests

Chemaclass:
## 📚 Description

Reorganize all Phel unit tests mirroring the same structure as we currently have in `src/phel/core.phel`

## 🔖 Changes

- Created a new test file for each block, similar to what we did in `src/phel/core.phel`
- I replaced:
  - `(= true `-> `(true? `
  - `(= false `-> `(false? `
  - `(= nil `-> `(nil? `
- No logic was altered


JesusValera (approved):
nicee

jenshaase (approved):

jenshaase:
Merged now!

Issue #223, opened 2021-02-02
Reorganize Exceptions


Chemaclass:
Currently, we have all exceptions together in a directory named `Exceptions`.

We want to define the exceptions inside the module where they belong.
Also, get rid of the `\Exceptions` and `\RuntimeExceptions` and use a meaningful domain exception instead.

PR #224, opened 2021-02-02
Reorganize exceptions

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/223
Get rid of the global `Exception` directory and move them to their right modules.

## 🔖 Changes

- Reorganized the exceptions inside a `Exceptions` directory per Module.
- I created some extra domain exceptions in order to get rid of the `\Exception` or `\RuntimeException` inside the domain.
- Clean up how most of the exceptions are handled and added PHPDoc `@throws ...` for a better understanding of the context and potential exception throws per function.


jenshaase (approved):
Very cool. I like that all `Exception` and `RuntimeException` are now removed!

PR #226, opened 2021-02-03
Create ExportCommand

jenshaase:
## 📚 Description

Creating a new command (ExportCommand) that uses a new module (Interop) for the bidirectional interoperability between Phel and PHP.

## 🔖 Changes

- Added ExportCommand which uses the new Interop module, responsible to generate the compiled PHP Wrappers code.


jenshaase:
@mabasic @aszenz This is what could be used to call Phel functions from PHP (see ExampleWrapper). Would that be helpful and is this enough or do you need more?

mabasic:
I will need a few days to try this out, but looks good at first. ❤

aszenz:
> @mabasic @aszenz This is what could be used to call Phel functions from PHP (see ExampleWrapper). Would that be helpful and is this enough or do you need more?

Thanks, this looks nice and simple, a more complex example will provide further insight though which I hope to get around to soon.

It would be interesting to see how this plays our for functions transforming compound data types.

src/php/Interop/PhelCaller.php
```diff
@@ -0,0 +1,33 @@
...
+
+    /**
+     * @return mixed
+     */
+    private function getPhelDefinition(string $namespace, string $definitionName)
+    {
+        $rt = RuntimeFactory::getInstance();
+        $rt->loadNs($namespace);
+
+        return $GLOBALS['__phel'][$namespace][$definitionName];
```
Chemaclass, line 31: Don't we have a constant for `'__phel'`?🤔

jenshaase: No. But we should create one.

composer.json
```diff
@@ -64,6 +64,10 @@
             "loader-dev": {
                 "phel\\": "tests/phel/"
             },
+            "export" : {
+                "target-directory" : "src/PhelGenerated",
+                "prefix-namespace" : "PhelGenerated"
+            },
```
Chemaclass, line 70: What do you think about this, @jenshaase? We should add these config options in the getting started guide (+ scaffolding)

jenshaase: Can we rename `prefix-namespace` to `namespace-prefix`.

However, we don't needs these config options in here. They are just there so you can test?

Regarding documentation: It should not be added to getting started. That too much. It should be added to "PHP Interop" and "Configuration".

Chemaclass: > Can we rename prefix-namespace to namespace-prefix.

Done

> However, we don't need these config options here. They are just there so you can test?

Yes, I leave them there so you can actually easily test them. It doesn't affect the real usage of the end-user.

> Regarding documentation: It should not be added to getting started. That too much. It should be added to "PHP Interop" and "Configuration".

Done.

composer.json
```diff
@@ -66,7 +66,10 @@
             },
             "tests": [
                 "tests/phel/"
-            ]
+            ],
+            "export-directories": ["src/phel"],
```
jenshaase, line 70: I liked it more when it had it own `export` section:

```
"export": {
  "source-directories": ["src/phel"],
  "target-directory": "src/PhelGenerated",
  "namespace-prefix": "PhelGenerated"
}
```

Also the `source-directories` config is never used. Did I miss something?

doc/content/documentation/configuration.md
```diff
@@ -20,7 +20,11 @@ These are all Phel specific configuration options available.
         },
         "tests": [
             "tests/"
-        ]
+        ],
+        "export" : {
+            "target-directory" : "src/PhelGenerated",
```
jenshaase, line 25: `source-directories` is missing

doc/content/documentation/configuration.md
```diff
@@ -59,6 +63,12 @@ The `loader-dev` configuration section is equivalent to the `loader` configurati
 
 This configuration entry defines a list of folders where the test files of a project can be found.
 
+### `export`
+
+To be able to use the export command you have to define these two options:
+
+- "target-directory": the root directory where the PHP wrappers will be store.
```
jenshaase, line 70: `source-directories` is missing

doc/content/documentation/configuration.md
```diff
@@ -59,6 +63,12 @@ The `loader-dev` configuration section is equivalent to the `loader` configurati
 
 This configuration entry defines a list of folders where the test files of a project can be found.
 
+### `export`
+
+To be able to use the export command you have to define these two options:
```
jenshaase, line 68: Reference the `Interop` chapter to get more information about the `export` command

doc/content/documentation/php-interop.md
```diff
@@ -114,4 +114,25 @@ In Phel you can also use PHP Magic Methods `__DIR__` and `__FILE__`. These resol
 ```phel
 (println __DIR__) # Prints the directory name of the file
 (println __FILE__) # Prints the filename of the file
-```
\ No newline at end of file
+```
+
+## Calling Phel functions from PHP
+
+You can define your own classes using the `Phel\Runtime\RuntimeFactory\PhelCallerTrait`.
```
jenshaase, line 121: Split this in two section.
1. Manually
2. Using the `export` command

src/php/PhelConfig.php
```diff
@@ -0,0 +1,44 @@
...
+
+    private function __construct(array $config)
+    {
+        $this->config = $config;
+    }
+
+    /**
+     * @return mixed|null
+     */
+    public function get(string $key)
```
jenshaase, line 40: I would like to see a full Getters for every Config Entry. More like what you did in `ExportConfig`

src/php/Command/Export/ExportCommand.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if (empty($wrappers)) {
+            $this->io->writeln('No functions were found to be exported.');
+            return;
+        }
+
+        $this->writeGeneratedWrappers(...$wrappers);
+    }
+
+    private function writeGeneratedWrappers(Wrapper ...$wrappers): void
```
jenshaase, line 47: Why not just an array?

src/php/Interop/Generator/WrapperGenerator.php
```diff
@@ -0,0 +1,35 @@
...
+        string $destinationDir,
+        CompiledPhpClassBuilder $classBuilder,
+        WrapperRelativeFilenamePathBuilder $relativeFilenamePathBuilder
+    ) {
+        $this->destinationDir = $destinationDir;
+        $this->classBuilder = $classBuilder;
+        $this->relativeFilenamePathBuilder = $relativeFilenamePathBuilder;
+    }
+
+    public function generateCompiledPhp(string $phelNs, FunctionToExport ...$functionsToExport): Wrapper
```
jenshaase, line 28: Why not just an array?

src/php/PhelFacade.php
```diff
@@ -25,10 +26,14 @@ final class PhelFacade
 
     test <filename> <filename> ...
         Tests the given files. If no filenames are provided all tests in the
-        test directory are executed.
+        "tests" directory are executed.
 
     fmt <filename-or-directory> ...
         Formats the given files.
+    
+    export
```
jenshaase, line 34: I'm still confused. Do we need a argument now or not?

composer.json
Chemaclass: Do you mean `"export-directories"`? It's used in the CommandFactory:126:
https://github.com/phel-lang/phel-lang/pull/226/files#diff-fa00b187a2c56b8ee580791058fc96b508572e50c5da75b64d9b750e8658148eR126

src/php/PhelFacade.php
Chemaclass: Nope. Why? Every time we run this command we need to remove all other generated code, the whole directory. This means if you would be able to export only one file it will be harder to keep track of what to remove and what to keep. For simplicity, I decided to don't allow parameters and let generate all exported functions at once every time. It's a simple command, where you won't need to do anything special but some previous configuration in your composer options, and that's all.

composer.json
jenshaase: Oh, miss it.

src/php/PhelFacade.php
jenshaase: The documentation said:
```
Or you can use the `phel export <phel-path>`
```

But you are correct we just don't need the argument. I just got confused

jenshaase:
Found a bug. If I remove the `namespace-prefix` settings the generated namespace in the PHP file is
```php
namespace \Phel;
```
which is invaid. The `\` needs to be removed.
```php
namespace Phel;
```

jenshaase:
If I remove the `target-directory` option I got warnings:
```
PHP Warning:  mkdir(): Permission denied in
Warning: file_put_contents(/Phel/Core.php): failed to open stream: No such file or directory in src/php/Command/Export/ExportCommand.php on line 62
```

doc/content/documentation/php-interop.md
```diff
@@ -118,37 +118,52 @@ In Phel you can also use PHP Magic Methods `__DIR__` and `__FILE__`. These resol
...
+To mark a function as exported the following meta data needs to be added to the function:
+
+```phel
+(defn my-function
+  @{:export true}
+  [a b]
+  (+ a b))
+```
+
+Now the `phel export` command will generate a PHP wrapper class in `src/PhelGenerated`.
```
JesusValera:
```suggestion
Now the `phel export` command will generate a PHP wrapper class in the `target-directory` path (in this case in `src/PhelGenerated`).
```

Maybe? 😄

JesusValera (commented):
Some typos :)

jenshaase:
Found another bug:
1. It added created one exported function. 
2. The `phel export` command was working and the wrapper class was created. 
3. Then I removed the export meta data from the function and executed `phel export` again.

I was the expecting that the wrapper class was deleted. But it was still there.

Can we add tests for all the bugs I found, so they don't happen again?

jenshaase:
It is merged now!

PR #227, opened 2021-02-04
Add rule void_return to php_cs

JesusValera:
## 📚 Description

I saw a lot of methods that don't have the `void` return.

## 🔖 Changes

- Add the rule (`'void_return' => true`) to the `.php_cs.dist` file
- Execute the command `composer csfix` and check tests are green 🍏


Chemaclass (approved):

jenshaase (approved):
Yes that was needed. We had a few changes that only add `void` recently. Thank you!

PR #236, opened 2021-02-20
Handle exceptions inside commands

Chemaclass:
## 📚 Description

Don't throw up the logic exceptions from the commands, but handle them inside the commands themselves.

## 🔖 Changes

- Remove the `@throws ...` from the commands. They don't intentionally throw an exception up.
- Catch and handle the exceptions inside the `command->run()` logic.

### DIFF
--- a/src/php/Command/CommandFacade.php
+++ b/src/php/Command/CommandFacade.php
@@ -4,20 +4,11 @@ declare(strict_types=1);
 
 namespace Phel\Command;
 
-use Phel\Command\Run\Exceptions\CannotLoadNamespaceException;
-use Phel\Command\Shared\Exceptions\ExtractorException;
-use Phel\Command\Test\Exceptions\CannotFindAnyTestsException;
-use Phel\Compiler\Emitter\Exceptions\CompiledCodeIsMalformedException;
-use Phel\Compiler\Emitter\Exceptions\FileException;
-use Phel\Compiler\Exceptions\CompilerException;
 use Phel\Runtime\RuntimeInterface;
 use RuntimeException;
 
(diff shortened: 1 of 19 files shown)


JesusValera (approved):
👍🏼 💯

jenshaase:
Can we also write the Stack Traces? Currently only the messages of the exceptions are printed. May we should add `writeException` method to the IO class, since this is not the first time it was forgotten?

Chemaclass:
I will take a look at it later. Thanks for the idea 💡

src/php/Command/CommandFacade.php
```diff
@@ -51,21 +36,12 @@ public function executeRunCommand(string $fileOrPath): void
...
-     *
-     * @throws CompilerException
-     * @throws CompiledCodeIsMalformedException
-     * @throws FileException
-     * @throws CannotFindAnyTestsException
      */
     public function executeTestCommand(array $paths): void
     {
-        $result = $this->commandFactory
+        $this->commandFactory
```
jenshaase, line 42: As far as I can see this isn't sending no output codes (SUCCESS_CODE, FAILED_CODE) anymore? This is required for the CI Pipeline to indicate that the tests failed.

jenshaase (changes requested):

src/php/Command/CommandFacade.php
Chemaclass: Ups, I didn't know this. Ok, I will rollback this as how it was before 🪂

jenshaase (approved):

jenshaase:
Thanks. It is merged now!

PR #244, opened 2021-03-10
Persistent data structures

jenshaase:
## 📚 Description

This adds persistent (immutable) data structures to Phel.

The target of this PR is it to replace:
* the Parens-Tuple with a PersistentList
* the Bracket-Tuple and Array data structure with a PersistentVector
* the Table with a PersistentHashMap
* the Set with a PersistentHashSet

## Motivation

The current data structures of Phel have a few flaws:
* The Tuple data structure is use for two syntax options `()` and `[]`.
* Iteratively adding a lot of values to a Tuple is very slow because the inner array is copied every time.
* Array, Table and Set are mutable
* Hashing on Tables and Set is not working for all types. For example `(set (set 1 2) (set 1 2))` will not result in in `(set (set 1 2))` but in `(set (set 1 2) (set 1 2))`.

## Target of this PR

The target of this Pull request is to replace the existing data structures with new persistent data structures.
* The Tuple with parens (`()`) will be replaced with a persistent list
* The Tuple with brackets (`[]`) and the Array will be replaced with a persistent vector.
* The Table will be replace by a persistent hash map. Since the Hash map is mutable the syntax for defining a hash map will change to `{}` (e.g `{:a 1 :b 2}`)
* The set will be replace with a persistent hash set.
* The Table data structure and syntax will be marked deprecated in the next version and removed in the following version.
* The Array data structure and syntax will be marked deprecated in the next version and removed in the following version.

## What are persistent data structures?

> A persistent data structure is a data structure that always preserves the previous version of itself when it is modified. Such data structures are effectively immutable, as their operations do not (visibly) update the structure in-place, but instead always yield a new updated structure.
(Source: https://en.wikipedia.org/wiki/Persistent_data_structure)

This is effectively done by representing data structures as tree. When change the data structure a new tree is created that shares most of the nodes with the old tree and replaces only the changed nodes with new nodes.

At lot of work on these persistent data structures has been done in Clojure. The implementations of this PR use a lot of theses ideas.

If you want to learn more about persistent data structures you could have a look at the following blog posts. These explain how a persistent vector in Clojure works:

 * https://hypirion.com/musings/understanding-persistent-vector-pt-1
 * https://hypirion.com/musings/understanding-persistent-vector-pt-2
 * https://hypirion.com/musings/understanding-persistent-vector-pt-3
 * https://hypirion.com/musings/understanding-clojure-transients
 * https://hypirion.com/musings/persistent-vector-performance-summarised

## Persistent List

A persistent List is simply a linked list. Access or modifications on the first element is efficient, random access is not.

The implementation of a persistent list is quite simple. It basically is a class with two properties: The first value and rest of the list.
```
<?php

/**
 * @template T
 */
class List {
  /** @var T */
  private $first;
  /** @var ?List<T> $list */
  private ?List $rest;
}
```
## Persistent Vector

Persistent Vectors offer efficient random access (by index). It is also very efficient to append a new value at the end. Slicing a vector can also be done in O(1). 

The implementation of a persistent vector is more complex than the persistent list above. It is based on the idea of a n-ary tree. The inner nodes simply link to the sub node, while the leaf node contain the values. A detailed introduction can be found on the links mentioned above.

## Persistent Map

A persistent map contains key value pairs in random order. Key objects must implement the interfaces `HashableInterface` and `EqualsInterface`.

The implementation of a persistent map is based on Hash Array Mapped Tries (see [Ideal Hash Trees](http://lampwww.epfl.ch/papers/idealhashtrees.pdf) from Phil Bagwell) that has been improved by Rich Hickey in Clojure. The Trie requires three different type of nodes:

* *ArrayNode*: The Array is a node that only links to its child nodes. ArrayNode are only used as inner Nodes
* *IndexedNode*: A indexed node can contain key-value pairs and references to child nodes. It is used as inner node or as leaf node
* *HashCollisionNode*: A hash collision node contains key value pairs, where the keys map to the same hash. This node can only be a leaf node.

For maps with eight or less key a more simple and efficient map implementation is used. This implementation is called `PersistentArrayMap` and stores the key value pairs in a simple array. To look up a key the array is iterated to search for the key.

## Persistent Set

A persistent set contains unique values in random order. Values must implement the interfaces `HashableInterface` and `EqualsInterface`.

The implementation is just a wrapper around the persistent map. A added value `v` with be stored as the key value pair `<v, v>`.

## Transient Data Structures

Nearly all persistent data structures have a transient version (except for Persistent List). The transient version of each persistent data structure is a mutable version of them. It store the value in the same way as the persistent version but instead of returning a new persistent version with every modification it modifies the current version. Transient versions are a little bit faster and can be used as builders for new persistent collections. Since transients use the same underlying storage it is very fast to convert a persistent data structure to a transient and back.

For example, if we want to convert a PHP Array to a persistent map. This function can be used:
```
(defn php-array-to-map
  "Converts a PHP Array to a tables."
  [arr]
  (let [res (transient {})] ; Convert a persistent data to a transient
    (foreach [k v arr]
      (put res k v)) ; Fill the transient map (mutable)
    (persistent res))) ; Convert the transient map to a persistent map.
```

## Needs to be done

- [x] Add more unit test to Vector
- [x] Improve Hash Set
- [x] Integrate Hash Set in Phel core library
- [ ] Test Default Hasher implementation
- [ ] Test Default Equalizer implementation
- [ ] Use TypeFactory everywhere (also for Symbols and Keywords)
- [x] Map, Vector and Set as a function (implementation of `FnInterface`)
- [x] Struct should use Persistent Hash Maps instead of Table

## 🔖 Changes
* `()` is now a list on not a tuple. (`(list? '())` evaluates to true, `(tuple? '())` evaluates to false)
* `[]` is now a vector on not a tuple. (`(vector? [])` evaluates to true, `(tuple? [])` evaluates to false)
* Type of meta data changed from table to hash map
* `put-in` Will now add a hash-map instead to a table
* `distinct` Will now return a vector instead of a array
* `frequencies` Will now return a hash map instead of a table
* `group-by` Will now return a hash map instead of a table
* Request headers, query params, cookie params, server params, ... will now be hash maps instead of tables
* HTML templates support vectors and hash maps
* `first`, `next`, `rest` can not be called on sets anymore. To iterate over a set use `foreach`, `for` or convert it to a vector `(apply vector set)`


Chemaclass:
The "CI / Type Checker" is red because of the new psalm version "4.6.4".
Screenshot 2021-03-20 at 19 45 26

If you force the version to "4.6.3" in `composer.json` then it will be green. I had the same problem in another PR.

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/FnSymbol.php
```diff
@@ -63,21 +65,21 @@ private function analyzeBody(FnSymbolTuple $fnSymbolTuple, RecurFrame $recurFram
...
 
-    private function createDoTupleWithBody(array $body): Tuple
+    private function createDoTupleWithBody(array $body): PersistentListInterface
     {
-        return Tuple::create(
+        return TypeFactory::getInstance()->persistentListFromArray([
             (Symbol::create(Symbol::NAME_DO))->copyLocationFrom($body),
-            ...$body
-        )->copyLocationFrom($body);
+            ...$body,
```
Chemaclass, line 72: > ERROR: DuplicateArrayKey - src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/FnSymbol.php:72:16 - String keys are not supported in unpacked arrays (see https://psalm.dev/151)
            ...$body,
            
Maybe you might fix these psalm errors by using the values from the array? 🤔 

```php
...array_values($body)
```

jenshaase: No it is just enough to change type annotation from `MyType[]` to `array<int, MyType>`. This came up with Psalm 4.7. It fixed this already in master but was to tired to fix this here.

Chemaclass: I can help you this evening with this :)

jenshaase: Fixed it.

tests/php/Benchmark/Lang/Collections/Vector/VectorA.php
```diff
@@ -50,4 +52,115 @@ public function pop(): PersistentVectorInterface
     {
         return $this;
     }
+
+    public function cdr(): void
+    {
+        // TODO: Implement cdr() method.
```
JesusValera, line 58: Do we want to have those comments even for the test?
If this class is like a "mock", I would suggest cleaning the methods. 😃

Chemaclass: We might/could/should revome those comments, indeed. I did let the IDE generate the missing methods that were required for the code to simply run 🏃 . But I don't think it's urgent. I created a PR https://github.com/phel-lang/phel-lang/pull/263 with some improvements for the current Benchmark tests that were introduced here. 
I didn't want to commit and pushed them directly here because I wanted @jenshaase to double-check them before. But yeah, the intention is to include them in this branch/PR.

I am sure we could find some places/scenarios where PHPBench might be really useful especially for regressions (even between branches!).

jenshaase: This class was just to test an alternative implementation. I think this can be removed complete. The current implementation is the vector is very good.

mabasic:
Love seeing the immutable word ❤️

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/LoopSymbol.php
```diff
@@ -68,30 +71,34 @@ public function analyze(Tuple $tuple, NodeEnvironmentInterface $env): LetNode
...
         if (count($lets) > 0) {
-            $bodyExpr = [];
-            for ($i = 2; $i < $tupleCount; $i++) {
-                $bodyExpr[] = $tuple[$i];
-            }
-            $letSym = Symbol::create(Symbol::NAME_LET)->copyLocationFrom($tuple[0]);
-            $letExpr = (Tuple::create($letSym, new Tuple($lets, true), ...$bodyExpr))->copyLocationFrom($tuple);
-            $newExpr = (Tuple::create($tuple[0], new Tuple($preInits, true), $letExpr))->copyLocationFrom($tuple);
+            $bodyExpr = $list->rest()->rest()->toArray();
+            $letSym = Symbol::create(Symbol::NAME_LET)->copyLocationFrom($list->get(0));
```
Chemaclass, line 75: Minor: I saw that we have now `TypeFactory::getInstance()->symbol()` and it's a wrapper for `Symbol::create()`. Actually, it's not being used at all.
The same as `TypeFactory::getInstance()->symbolForNamespace()` also not used.

I still need to think about this concept deeper... 🤔

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/LetSymbol.php
```diff
@@ -0,0 +1,139 @@
...
+            ->persistentListFromArray($newListData)
+            ->copyLocationFrom($list)
+            ->withMeta($list->getMeta());
+
+        return $this->analyzeLetOrLoop($newList, $env);
+    }
+
+    private function analyzeLetOrLoop(PersistentListInterface $list, NodeEnvironmentInterface $env): LetNode
+    {
+        $exprs = $list->rest()->rest()->toArray();
```
Chemaclass, line 66: Minor observation: could we place this variable closer to its usage? I see it's being used only in line 92. Could we move this line 66 down to the 89, for example? Unless there is a reason why is this variable here, the first line in the function, I feel it's too alone and a bit "far away" from its usage...

src/php/Compiler/CompilerFactory.php
```diff
@@ -51,9 +51,9 @@ public function createFileCompiler(GlobalEnvironmentInterface $globalEnv): FileC
         );
     }
 
-    public function createLexer(): LexerInterface
+    public function createLexer(bool $withoutLocation = false): LexerInterface
```
Chemaclass, line 54: What about using a positive statement?
```php
public function createLexer(bool $withLocation = true): LexerInterface
```
> It applies to the interface as well.

src/php/Compiler/Emitter/OutputEmitter/LiteralEmitter.php
```diff
@@ -50,10 +52,18 @@ public function emitLiteral($x): void
...
+            $this->emitVector($x);
+        } elseif ($x instanceof PersistentListInterface) {
+            $this->emitList($x);
         } else {
-            throw new RuntimeException('literal not supported: ' . gettype($x));
+            $typeName = gettype($x);
+            if ($typeName === 'object') {
+                $typeName = get_class($x);
+            }
+            throw new RuntimeException('literal not supported: ' . $typeName);
```
Chemaclass, line 66: Good one 👍🏼

src/php/Compiler/Lexer/Lexer.php
```diff
@@ -36,10 +36,12 @@ final class Lexer implements LexerInterface
     private int $line = 1;
     private int $column = 0;
     private string $combinedRegex;
+    private bool $withoutLocation;
 
-    public function __construct()
+    public function __construct(bool $withoutLocation = false)
```
Chemaclass, line 41: What about using `bool $withLocation = true`?

Chemaclass (commented):
`108 / 296 files viewed`. I will continue tomorrow 🧭

tests/php/Unit/Compiler/Analyzer/AnalyzePersistentListTest.php
```diff
@@ -0,0 +1,249 @@
...
+        $list = TypeFactory::getInstance()->persistentListFromArray([
+            Symbol::create(Symbol::NAME_PHP_OBJECT_STATIC_CALL), '', Symbol::create(''),
+        ]);
+        self::assertInstanceOf(
+            PhpObjectCallNode::class,
+            $this->listAnalyzer->analyze($list, NodeEnvironment::empty())
+        );
+    }
+
+    /*public function testSymbolWithNamePhpArrayGet(): void
```
Chemaclass, line 148: What's the status of these commented tests? Do we need to update them? Or are they obsolete and we should remove them?

Chemaclass (approved):
This PR is a huge step forward for the project itself. 
I found some minor things that we could work on in another iteration.
Overall: **amazing** job here 👏👌🏻

src/phel/core.phel
```diff
@@ -107,61 +117,70 @@ elements, returns nil."
...
+
+(def meta
+  "```phel\n(meta obj)\n```\nGets the meta of the give object"
+  (fn [obj]
+    (php/-> obj (getMeta))))
+
+(def set-meta!
+  "```phel\n(set-meta! obj)\n```\nSets the meta data to a given object"
+  (fn [obj meta]
+    (php/-> obj (withMeta meta))))
```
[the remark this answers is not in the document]
jenshaase: This was part of the public API before the PR. I will leave it in for now.

src/php/Lang/TypeFactory.php
```diff
@@ -0,0 +1,121 @@
...
+
+        return PersistentHashMap::fromArray($this->hasher, $this->equalizer, $kvs);
+    }
+
+    public function emptyPersistentHashSet(): PersistentHashSetInterface
+    {
+        return new PersistentHashSet($this->hasher, null, $this->emptyPersistentHashMap());
+    }
+
+    public function persistentHashSetFromArray(array $values): PersistentHashSetInterface
```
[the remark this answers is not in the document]
jenshaase: This is used in Phel's core library (used in `(set ...)`.)

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/LoopSymbol.php
jenshaase: Yes that is still missing. I want to initial new Symbols with the TypeFactory and not with `Symbol::create...`. But maybe we clean this up later.

src/php/Compiler/Reader/QuasiquoteTransformer.php
```diff
@@ -25,122 +29,159 @@ public function __construct(GlobalEnvironmentInterface $env)
...
+            return $form->get(1);
         }
 
         if ($this->isUnquoteSplicing($form)) {
             throw new SpliceNotInListException();
         }
 
-        if ($form instanceof Tuple && count($form) > 0) {
-            return $this->createTupleFromTuple($form);
+        if ($form instanceof PersistentList && count($form) > 0) {
```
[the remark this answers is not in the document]
jenshaase: I know what you mean. The problem is that not every from is `Countable` (e.g. string or int).

src/php/Compiler/Reader/QuasiquoteTransformer.php
```diff
@@ -25,122 +29,159 @@ public function __construct(GlobalEnvironmentInterface $env)
...
-            return $this->createTupleFromPhelArray($form);
+            return $this->createFromPhelArray($form);
         }
 
         if ($this->isLiteral($form)) {
             return $form;
         }
 
-        return $this->createTupleOtherwise($form);
+        return $this->createOtherwise($form);
```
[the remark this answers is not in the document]
jenshaase: I also don't have a better name yet.

src/php/Lang/Collections/HashMap/HashMapNodeInterface.php
```diff
@@ -0,0 +1,35 @@
...
+
+    /**
+     * @template TDefault
+     *
+     * @param TKey $key
+     * @param ?TDefault $notFound
+     *
+     * @return TValue|TDefault
+     */
+    public function find(int $shift, int $hash, $key, $notFound);
```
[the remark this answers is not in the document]
jenshaase: `null` is a not super good here, since the value of in a map can be `null` too (it is ok for the test). Additionally, this is only used internally as part of the HashMap.

src/php/Lang/Collections/HashMap/IndexedNode.php
```diff
@@ -0,0 +1,271 @@
...
+
+    /**
+     * @param K $key
+     * @param V $value
+     *
+     * @return HashMapNodeInterface<K, V>
+     */
+    private function splitNode(int $idx, int $shift, int $hash, $key, $value, Box $addedLeaf): HashMapNodeInterface
+    {
+        $nodes = []; //array_fill(0, 32, null);
```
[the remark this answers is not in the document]
jenshaase: no

src/php/Lang/Hasher.php
```diff
@@ -0,0 +1,67 @@
...
+            return $value->hash();
+        }
+
+        if ($value === true) {
+            return 1231;
+        }
+
+        if ($value === false) {
+            return 1237;
+        }
```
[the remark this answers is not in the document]
jenshaase: I took them from Clojure. I will add a comment.

src/php/Lang/Symbol.php
```diff
@@ -34,9 +36,10 @@ final class Symbol extends AbstractType implements IdenticalInterface
     public const NAME_TABLE = 'table';
     public const NAME_THROW = 'throw';
     public const NAME_TRY = 'try';
-    public const NAME_TUPLE = 'tuple';
-    public const NAME_TUPLE_BRACKETS = 'tuple-brackets';
     public const NAME_PHP_OBJECT_SET = 'php/oset';
+    public const NAME_LIST = 'list';
+    public const NAME_VECTOR = 'vector';
+    public const NAME_MAP = 'hash-map';
```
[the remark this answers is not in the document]
jenshaase: It's `hash-map` because we already have a `map` function.

src/php/Lang/Table.php
```diff
@@ -7,11 +7,16 @@
 use ArrayAccess;
 use Countable;
 use Iterator;
+use Phel\Lang\Collections\HashMap\PersistentHashMapInterface;
 use Phel\Printer\Printer;
 use RuntimeException;
 
+/**
+ * @template-implements SeqInterface<\Phel\Lang\Collections\Vector\PersistentVectorInterface, PhelArray>
```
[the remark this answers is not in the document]
jenshaase: If I use the `use` statement the PHP CS Fixer will remove the line, because it is not used in the code. It doesn't recognizes the comment.

src/php/Lang/TypeFactory.php
```diff
@@ -0,0 +1,121 @@
...
+
+    public function symbol(string $name): Symbol
+    {
+        return Symbol::create($name);
+    }
+
+    public function symbolForNamespace(?string $namespace, string $name): Symbol
+    {
+        return Symbol::createForNamespace($namespace, $name);
+    }
```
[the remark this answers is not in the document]
jenshaase: Yes, they should be used instead of `Symbol::create`

src/php/Compiler/Lexer/Lexer.php
jenshaase: Will change this.

src/php/Compiler/Analyzer/Ast/MapNode.php
```diff
@@ -0,0 +1,24 @@
...
+declare(strict_types=1);
+
+namespace Phel\Compiler\Analyzer\Ast;
+
+use Phel\Compiler\Analyzer\Environment\NodeEnvironmentInterface;
+use Phel\Lang\SourceLocation;
+
+final class MapNode extends AbstractNode
+{
+    private array $keyValues;
```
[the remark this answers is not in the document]
jenshaase: Will add some types.

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/LetSymbol.php
jenshaase: Will fix this.

src/php/Compiler/CompilerFactory.php
jenshaase: Will fix this.

src/php/Lang/Symbol.php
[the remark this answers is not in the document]
Chemaclass: Of course! 😂 I completely forgot about it 🤦‍♂️

src/php/Compiler/Reader/QuasiquoteTransformer.php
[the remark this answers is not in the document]
Chemaclass: Ahh, I understand. It's ok for now then.

src/php/Compiler/Analyzer/Analyzer.php
```diff
@@ -111,15 +110,23 @@ public function analyze($x, NodeEnvironmentInterface $env): AbstractNode
...
+        if ($x instanceof PersistentListInterface) {
+            return (new AnalyzePersistentList($this))->analyze($x, $env);
+        }
+
+        if ($x instanceof PersistentVectorInterface) {
+            return (new AnalyzePersistentVector($this))->analyze($x, $env);
+        }
+
+        if ($x instanceof PersistentHashMapInterface) {
+            return (new AnalyzeMap($this))->analyze($x, $env);
```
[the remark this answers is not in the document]
jenshaase: I improved naming of persistent maps.

- They are now call PersistentMap (before: PersistentHashMap)
- The name PersistentHashMap is now only used for the concrete implementation of the PersistentMapInterface
- I renamed the Analyzer to `AnalyzePersistentMap`

PR #245, opened 2021-03-10
Adding take-last

Chemaclass:
## 📚 Description

Adding `take-last` phel function

## 🔖 Changes

- Added a new phel function `take-last`

### DIFF
--- a/src/phel/core/sequence-functions.phel
+++ b/src/phel/core/sequence-functions.phel
@@ -96,6 +96,11 @@ collection in map"))
   [n xs]
   (slice xs 0 (if (php/< n 0) 0 n)))
 
+(defn take-last
+  "Takes the last `n` elements of `xs`."
+  [n xs]
+  (if (php/<= n 0) @[] (slice xs (php/* -1 n))))
+
 (defn take-while
   "Takes alle elements at the front of `xs` where `(pred x)` is true."
   [pred xs]
--- a/tests/phel/test/core/sequence-functions.phel
+++ b/tests/phel/test/core/sequence-functions.phel
@@ -61,7 +61,21 @@
   (is (= @[] (take 0 @["a" "b" "c"])) "take zero elements")
   (is (= @[] (take -1 @["a" "b" "c"])) "take with negative number")
   (is (= ["a" "b"] (take 2 ["a" "b" "c"])) "take on tuple")
-  (is (= (php/array "a" "b") (take 2 (php/array "a" "b" "c"))) "take on php array"))
+  (is (= [] (take -1 ["a" "b" "c"])) "take on tuple")
+  (is (= (php/array "a" "b") (take 2 (php/array "a" "b" "c"))) "take on php array")
+  (is (= (php/array) (take -1 (php/array "a" "b" "c"))) "take on php array"))
+
+(deftest test-take-last
+  (is (= @["c"] (take-last 1 @["a" "b" "c"])) "take-last element")
+  (is (= @["b" "c"] (take-last 2 @["a" "b" "c"])) "take-last two elements")
+  (is (= @["a" "b" "c"] (take-last 3 @["a" "b" "c"])) "take-last three elements")
+  (is (= @["a" "b" "c"] (take-last 4 @["a" "b" "c"])) "take-last four elements")
+  (is (= @[] (take-last 0 @["a" "b" "c"])) "take-last zero elements")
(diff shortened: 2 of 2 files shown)


src/phel/core/sequence-functions.phel
```diff
@@ -96,6 +96,11 @@ collection in map"))
   [n xs]
   (slice xs 0 (if (php/< n 0) 0 n)))
 
+(defn take-last
+  "Takes the last `n` elements of `xs`."
+  [n xs]
+  (if (php/<= n 0) @[] (slice xs (php/* -1 n))))
```
Chemaclass, line 102: I am not using `(- n)` (and instead `php/* -1`) because the math module where the function `-` is defined is not loaded yet. Here we are getting into the dependency order problem, where we are trying to use core functions that aren't loaded yet ... I am talking about these loaders inside the `core.phel`. We need to think about this problem and how to solve it in the future 🤔 
```lisp
(load (str __DIR__ "/core/basic-constructors.phel"))
(load (str __DIR__ "/core/basic-sequence-operation.phel"))
(load (str __DIR__ "/core/control-structures.phel"))
(load (str __DIR__ "/core/boolean-operation.phel"))
(load (str __DIR__ "/core/type-operation.phel"))
(load (str __DIR__ "/core/sequence-operation.phel"))
(load (str __DIR__ "/core/for-loop.phel"))
(load (str __DIR__ "/core/sequence-functions.phel"))
(load (str __DIR__ "/core/set-operation.phel"))
(load (str __DIR__ "/core/function-operation.phel"))
(load (str __DIR__ "/core/more-sequence-operation.phel"))
(load (str __DIR__ "/core/bitwise-operation.phel"))
(load (str __DIR__ "/core/math-operation.phel"))
(load (str __DIR__ "/core/print-operation.phel"))
(load (str __DIR__ "/core/threading-macros.phel"))
```

jenshaase: That is ok for me.

> We need to think about this problem and how to solve it in the future

What do you mean exactly? Phels code is compiled/executed in order. Therefore, you can only use function that are already defined before.

tests/phel/test/core/sequence-functions.phel
```diff
@@ -61,7 +61,21 @@
...
+  (is (= @["c"] (take-last 1 @["a" "b" "c"])) "take-last element")
+  (is (= @["b" "c"] (take-last 2 @["a" "b" "c"])) "take-last two elements")
+  (is (= @["a" "b" "c"] (take-last 3 @["a" "b" "c"])) "take-last three elements")
+  (is (= @["a" "b" "c"] (take-last 4 @["a" "b" "c"])) "take-last four elements")
+  (is (= @[] (take-last 0 @["a" "b" "c"])) "take-last zero elements")
+  (is (= @[] (take-last -1 @["a" "b" "c"])) "take-last with negative number")
+  (is (= ["b" "c"] (take-last 2 ["a" "b" "c"])) "take-last on tuple")
+  (is (= @[] (take-last -1 ["a" "b" "c"])) "take-last with negative number on tuple")
+  (is (= (php/array "b" "c") (take-last 2 (php/array "a" "b" "c"))) "take-last on php array")
+  (is (= @[] (take-last -1 (php/array "a" "b" "c"))) "take-last with negative number on php array"))
```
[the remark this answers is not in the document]
jenshaase: I don't have a good solution for it. For me this is ok.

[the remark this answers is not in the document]
jenshaase: Could you add some test for an empty array?

src/phel/core/sequence-functions.phel
Chemaclass: I mean, we need to be aware of this, that depending on the order of the functions in the core we can use them or not. This was a perfect example where we cannot use `(- n)` and instead we have to use `(php/* 1)`. Anyway, not a big deal, and yes, it's in the nature of the lang, so there is no solution for this right now. And I think is also fine (to be aware at least)

jenshaase: Ok. I also had the problem after we split the file. When it was in one file it was much easier to reason about the order.

tests/phel/test/core/sequence-functions.phel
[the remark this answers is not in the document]
Chemaclass: done

jenshaase (approved):

jenshaase:
Thanks. I merged it!

PR #246, opened 2021-03-12
Fixes partition function.

mabasic:
## 📚 Description

`partition` was omitting all items if the length of the array was lower than `n` and it omitted items which length of the array was lower than `n` inside a partition.

Examples:

```
(def items @[1 2 3 4 5])
(partition 4 items)
@[@[1 2 3 4]]
```

I expected to get @[@[1 2 3 4] @[5]]

```
(def items @[1 2 3 4 5])
(partition 6 items)
@[]
```

I expected to get @[@[1 2 3 4 5]]

## 🔖 Changes

- `partition` now returns all items if the length of the given array is lower than the given size `n`.
- `partition` now returns remaining items if the size of the remaining array is lower than given size `n`.


src/phel/core/sequence-functions.phel
```diff
@@ -303,14 +303,29 @@ collection in map"))
...
     (let [res @[]]
       (loop [xs xs]
         (let [[a b] (split-at n xs)]
           (push res a)
           (if (>= (count b) n)
             (recur b)
-            res))))))
+            (do (push res b) res)))))))
+            
+(defn partition
```
Chemaclass, line 318: Isn't this function already defined in line 305? I think this is duplicated, no?

Chemaclass (changes requested):

src/phel/core/sequence-functions.phel
mabasic: OMG. Yes, I forgot to remove it. Nice catch. I am using the Web interface for this PR...

Chemaclass (approved):
This is great. Thanks for the fix!

jenshaase:
I think I designed it the way it was, but currently I'm not sure why. I'm open to change this to the proposed solution. However, I will put some research into how this is done in other languages.

Chemaclass:
> @jenshaase  I think I designed it the way it was, but currently I'm not sure why.

I know why you did it in this way, I realized now that the clojure signature https://clojuredocs.org/clojure.core/partition is actually  3:
```clojure
A: (partition n coll)
B: (partition n step coll)
C: (partition n step pad coll)
```
The implementation that we had before was A.
The changes from that PR are altering the behavior from that one because it's using for granted the "pad value" by taking always all leftover values.

src/phel/core/sequence-functions.phel
```diff
@@ -301,16 +301,18 @@ collection in map"))
   "Returns a tuple of [(take-while pred coll) (drop-while pred coll)]."
   [f xs]
   [(take-while f xs) (drop-while f xs)])
-
+
```
JesusValera, line 304:
```suggestion

```
Removing empty space

mabasic:
I have modeled this function to match this: https://laravel.com/docs/8.x/collections#method-chunk

JesusValera (approved):

jenshaase:
I did a little review on `partition` functions in other languages. There is not common definition on how `partition` (or `chunk`) works in other languages. I saw languages that behave like the original implementation and languages that behave like the proposed implementation. However, I think the proposed implementation is much easier to understand and has more use cases. Therefore I will accept this PR.

I will just have a look at the implementation in the next days. Maybe I can find a easier implementation with the same outputs.

jenshaase:
@mabasic I just merged the Persistent Data Structures into master. This PR has now some merge conflicts. Do you want do fix them or should I do that?

mabasic:
I'll do it, but It will have to wait next week. I'll be afk from tomorrow.

mabasic:
@jenshaase I have fixed the conflict, but have maybe made a mess with git commits. I've also tried to replace arrays with vectors, but the functions behaved differently. Need to learn about new data structures.

src/phel/core/sequence-functions.phel
```diff
@@ -314,16 +314,17 @@ collection in map"))
...
   [(take-while f xs) (drop-while f xs)])
-
 (defn partition
-  "Partition an indexed data structure into vectors of size n. Returns a new array."
+  "Partition an indexed data structure into vectors of maximum size n. Returns a new vector."
   [n xs]
-  (if (< (count xs) n)
-    @[]
+  (if (<= (count xs) n)
+    (if (= (count xs) 0) @[] @[xs])
```
jenshaase, line 321: Change this to:
```
(if (= (count xs) 0) [] [xs])
```

src/phel/core/sequence-functions.phel
```diff
@@ -314,16 +314,17 @@ collection in map"))
...
     (let [res @[]]
```
jenshaase, line 322: change this to:
```
(let [res (transient [])]
```

src/phel/core/sequence-functions.phel
```diff
@@ -314,16 +314,17 @@ collection in map"))
...
       (loop [xs xs]
         (let [[a b] (split-at n xs)]
           (push res a)
           (if (>= (count b) n)
             (recur b)
-            res))))))
+            (if (= (count b) 0) 
+              res
```
jenshaase, line 329: Change this to
```
(persistent res)
```

jenshaase (changes requested):
I added a few hints on how it should work!

mabasic:
@jenshaase Thanks for helping. I got it working. Take a look at #271

src/phel/core/sequence-functions.phel
```diff
@@ -314,16 +314,17 @@ collection in map"))
...
-  (if (< (count xs) n)
-    @[]
-    (let [res @[]]
+  (if (<= (count xs) n)
+    (if (= (count xs) 0) [] [xs])
+    (let [res (transient [])]
       (loop [xs xs]
         (let [[a b] (split-at n xs)]
-          (push res a)
+          (persistent (push res a))
```
jenshaase, line 325: I think this is not necessary. It should only be
```
(push res a)
```

jenshaase (changes requested):

jenshaase (approved):
Thanks a lot! I will merge it now.

PR #253, opened 2021-03-20
Implementing Gacela Framework

Chemaclass:
## 📚 Description

I was working on a concept to unify and improve the Factories/Facades/Config for each module. 
The result came into a new module, which I decoupled from this project (phel-lang) and I published it as https://github.com/gacela-project/gacela

You can read more info with simple examples on http://gacela-project.com/

## 🔖 Changes

Implemented the Gacela Framework module design. Bullet points:
  - The Facades are the open entry point of the communication between different modules. They have access to the Factories.
  - The Factories creates/instantiates the domain services/models which will be used by the Facade. They have access to its Config's module.
  - The Config has access to the values defined in the root/config.php file.
  - The dependencies between the modules (using their Facades) are defined inside the DependencyProvier class of the module.

## 🛑 Plugin dependency

We need to merge https://github.com/phel-lang/phel-composer-plugin/pull/4 and update the version of it here before merging this PR.


src/php/Compiler/CompilerFactoryInterface.php
```diff
@@ -4,24 +4,28 @@
...
+use Phel\Compiler\Compiler\EvalCompilerInterface;
+use Phel\Compiler\Compiler\FileCompilerInterface;
 use Phel\Compiler\Emitter\EmitterInterface;
 use Phel\Compiler\Emitter\OutputEmitterInterface;
 use Phel\Compiler\Lexer\LexerInterface;
 use Phel\Compiler\Parser\ParserInterface;
 use Phel\Compiler\Reader\ReaderInterface;
 
+/**
+ * deprecated Remove me without replacement. CompilerFacade is the one which should have the Interface!
```
JesusValera, line 16: ?

.php_cs.dist
```diff
@@ -10,6 +10,7 @@ return PhpCsFixer\Config::create()
   ->setRules([
     '@PSR2' => true,
     'array_syntax' => ['syntax' => 'short'],
+    'binary_operator_spaces' => ['operators' => ['|' => 'no_space']],
```
JesusValera, line 13: Any reason for the `no_space`?
As far as I know, we should use spaces, isn't it?

image
_src: https://www.php-fig.org/psr/psr-12/_

src/php/Compiler/Emitter/OutputEmitter/SourceMap/VLQ.php
```diff
@@ -128,7 +128,7 @@ public function encodeInteger(int $num): string
         $result = '';
 
         if ($num < 0) {
-            $num = (-$num << 1) | 1;
+            $num = (-$num << 1)|1;
```
JesusValera, line 131: spaces? 😄

src/php/Printer/PrinterFacade.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Printer;
+
+final class PrinterFacade
+{
+    /** @var PrinterFactory */
```
JesusValera, line 9: Remove the comment?

src/php/Runtime/RuntimeFacade.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        return $this->getFactory()
+            ->createNamespaceExtractor()
+            ->getNamespacesFromDirectories($directories);
+    }
+
+    public function getNamespaceFromFile(string $path): string
+    {
+        return $this
+            ->getFactory()
```
JesusValera, line 29:
```suggestion
        return $this->getFactory()
```
?

tests/php/Integration/Command/Export/ExportCommandTest.php
```diff
@@ -4,61 +4,46 @@
...
 
         $this->expectOutputRegex('~Exported namespaces:~');
         $this->expectOutputRegex('~TestCmdExportMultiple/Adder~');
         $this->expectOutputRegex('~TestCmdExportMultiple/Multiplier~');
 
         $command->run();
 
-        self::assertFileExists("{$currentDir}/PhelGenerated/TestCmdExportMultiple/Adder.php");
-        self::assertFileExists("{$currentDir}/PhelGenerated/TestCmdExportMultiple/Multiplier.php");
+        self::assertFileExists(__DIR__ . '/PhelGenerated/TestCmdExportMultiple/Adder.php');
```
JesusValera, line 39: Not sure if this works on Windows, maybe replace all those `/` by `DIRECTORY_SEPARATOR`?

tests/php/Integration/Command/Export/vendor/PhelRuntime.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+use Phel\Compiler\Analyzer\Environment\GlobalEnvironment;
+use Phel\Runtime\RuntimeSingleton;
+
+$runtime = RuntimeSingleton::initializeNew(new GlobalEnvironment());
+$runtime->addPath('phel\\', [__DIR__ . '/../../../../../../src/phel']);
```
JesusValera, line 7: 🙃 (`/../../../../......`)

JesusValera (approved):
👏🏼 👏🏼  So far so good.

src/php/Compiler/CompilerFactoryInterface.php
Chemaclass: Jap, I forgot to remove this one. Thanks for the reminder

.php_cs.dist
Chemaclass: Totally right. Actually, this change doesn't belong to this PR at all. I will undo this change.

tests/php/Integration/Command/Export/ExportCommandTest.php
Chemaclass: Good question... but for now I would not bother about changing this. Feel free to create a follow-up PR with this idea, but keep in mind that there might be more places (not just this file).

src/php/PhelAbstractConfig.php
```diff
@@ -0,0 +1,32 @@
...
+namespace Phel;
+
+use Gacela\Framework\AbstractConfig;
+use Gacela\Framework\Config;
+use function json_decode;
+
+/**
+ * This is a layer on top of Gacela config, to get the values from the composer.json.
+ */
+abstract class PhelAbstractConfig extends AbstractConfig
```
jenshaase, line 14: I should be named `AbstractPhelConfig`. See: https://github.com/phel-lang/phel-lang/blob/master/adrs/php/0001_naming_conventions.md

src/php/PhelAbstractConfig.php
```diff
@@ -0,0 +1,32 @@
...
+     * @param null|mixed $default
+     */
+    protected function get(string $key, $default = null)
+    {
+        if (empty($this->composerJson)) {
+            $composerJsonPath = Config::getApplicationRootDir() . DIRECTORY_SEPARATOR . 'composer.json';
+            $this->composerJson = json_decode(file_get_contents($composerJsonPath), true);
+        }
+
+        return $this->composerJson[$key] ?? $default;
```
jenshaase, line 30: Since everything is in `extra.phel`, we could directly return this.

```php
return $this->composerJson['extra']['phel'][$key] ?? $default
```

src/php/Command/CommandConfig.php
```diff
@@ -4,29 +4,21 @@
...
+use Gacela\Framework\Config;
+use Phel\PhelAbstractConfig;
 
-final class CommandConfig extends AbstractPhelConfig implements CommandConfigInterface
+final class CommandConfig extends PhelAbstractConfig
 {
+    /**
+     * @return string[]
+     */
     public function getDefaultTestDirectories(): array
```
jenshaase, line 15: Why is this named `default`? `getTestDirectories` would be a better name I think.

src/php/Command/CommandDependencyProvider.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        $this->addFacadeCompiler($container);
+        $this->addFacadeFormatter($container);
+        $this->addFacadeInterop($container);
+        $this->addFacadeRuntime($container);
+    }
+
+    private function addFacadeCompiler(Container $container): void
+    {
+        $container->set(self::FACADE_COMPILER, function (Container $container): CompilerFacadeInterface {
```
jenshaase, line 35: Couldn't this class be be simplified:

```php
    private function addToContainer(Container $container, string $name, string $class): void
    {
        $container->set($name, function (Container $container) use ($class): CompilerFacadeInterface {
            return $container->getLocator()->get($class);
        });
    }

    public function provideModuleDependencies(Container $container): void
    {
        $this->addToContainer($container, self::FACADE_COMPILER, CompilerFacade::class);
        // ...
    }
```

jenshaase (changes requested):
I'm not fully done with my review, but I have some more feedback already.

src/php/PhelAbstractConfig.php
Chemaclass: Totally 👍

src/php/PhelAbstractConfig.php
Chemaclass: I like the idea. I will do that 💡

src/php/Command/CommandDependencyProvider.php
Chemaclass: I see your point, but in your example, it wouldn't be possible to define the return type for the anonymous function which is using the locator to resolve the dependency. Also, the scope of this is pretty straight forward: To see which dependencies are we defining (once per function):
```php
$this->addFacadeCompiler($container);
$this->addFacadeFormatter($container);
$this->addFacadeInterop($container);
$this->addFacadeRuntime($container);
```
And I think it wouldn’t be a problem to leave it the way it is (at least for now).

src/php/AbstractPhelConfig.php
```diff
@@ -4,36 +4,29 @@
...
+     *
+     * @param null|mixed $default
      */
-    public function get(string $key)
+    protected function get(string $key, $default = null)
     {
-        return $this->config[$key] ?? null;
+        if (empty($this->composerJson)) {
+            $composerJsonPath = Config::getApplicationRootDir() . DIRECTORY_SEPARATOR . 'composer.json';
+            $this->composerJson = json_decode(file_get_contents($composerJsonPath), true);
```
jenshaase, line 27: Should this handle error like,
* composer.json is not readable
* composer.json is not parsable
?

src/php/Formatter/Formatter/FormatterInterface.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Phel\Formatter;
+namespace Phel\Formatter\Formatter;
```
jenshaase, line 5: I don't like these duplicated namespaces like,
`Formatter\Formatter`
`Compiler\Compiler`

Do you have any better ideas?

src/php/Formatter/FormatterFacadeInterface.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Formatter;
+
+interface FormatterFacadeInterface
```
jenshaase, line 7: Do we really need interfaces for each Facade? Could it, in theory, be possible to have multiple implementations of this interface?

jenshaase (changes requested):

src/php/Formatter/FormatterFacadeInterface.php
Chemaclass: The goal of the interface for the Facade is not to have multiple implementations for it but to **invert the dependencies** between the modules, so one module can communicate with another module via their Facade signatures/abstraction (interface). 
In short: the interfaces for the Facades are inverting the dependencies between the ones that used them.

src/php/Formatter/Formatter/FormatterInterface.php
Chemaclass: Good point. Let's start defining what does this means for each "level" in the namespace:
- The first `Formatter` is actually the module. 
- The second `Formatter` is a directory inside the module (we could also call it a "submodule", but that's not relevant here).

Being aware of this, the directory "Formatter" inside the module "Formatter" might seems a bit boilerplate, but it's in favor of the pattern that every other module is also having in place. Because the main point here is that at the module-root-level we'll find only the module stuff (Facade, Factory, Config, DependencyProvider). And from here, we can simply organize the submodules/directories following a simply DDD pattern. This is why I named it that way "Formatter\Formatter". The domain language of the module will group (and name) the actual logic of it.

I would keep it for now as it is, but if we came with other ideas we can always apply them in another iteration. I am always open to suggestions :)

src/php/AbstractPhelConfig.php
Chemaclass: Good point. Yes, I will verify this and throw a RuntimeException giving some hints in case of error.

Chemaclass: Done: https://github.com/phel-lang/phel-lang/pull/253/commits/54a560bb95d0ca1cc7dbf500beb8fe2fea97df3a

src/php/Formatter/FormatterFacadeInterface.php
jenshaase: Ok. However, the PrinterFacade is not using an Interface. Did you missed it or is there another reason?

jenshaase (changes requested):
Probably the last question...

src/php/Formatter/FormatterFacadeInterface.php
Chemaclass: Ahh, yes, that's right. The thing is that currently the `PrinterFacade` is not being used at all.
We are currently still using the old concept behavior of the `Printer`, using the static `readable()` and `nonReadable()`. 

So, this is a really good point/catch. My first idea was using the `PrinterFacade` as well (similar philosophy as every other module) but, the `Printer` is "coupled" somehow to the "Lang module" in a way which I preferred to work on in a separate PR (in order to simplify this PR a bit).
This is the reason why there is no PrinterFacadeInterface, basically, because the PrinterFacade is not in used (at least currently). I plan to work on it in the next iteration. So we can do two things:
1) I remove the unused code (PrinterFacade and PrinterFactory), and I create them in the future PR
2) I leave these two classes and I will continue with them in the next PR

I would prefer to do the 1) option. Is this ok for you, @jenshaase?

jenshaase: I would suggest 2. and also add the interface. Then we have a complete Printer Module and can refactor the usage of the Printer in another PR.

Chemaclass: Ok, then we leave it the second option. Totally ok for me as well. Then there is nothing else to add here :)

PR #258, opened 2021-03-22
Lock psalm to 4.6.3 and update composer

Chemaclass:
## 📚 Description

The version of psalm 4.6.4 is breaking our type checker CI, there must be some incompatibilities, so I would rather lock the version of psalm to the previous latest one 4.6.3 in order to have a green CI. And we'll try to upgrade psalm in another future version.

## 🔖 Changes

- Lock psalm to 4.6.3
- Updated all other composer dev dependencies

### Extra
This PR fixes the CI on this one: https://github.com/phel-lang/phel-lang/pull/244

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "require-dev": {
         "ext-readline": "*",
         "phpunit/phpunit": "^9.5",
-        "vimeo/psalm": "^4.6",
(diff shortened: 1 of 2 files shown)


JesusValera (approved):
👍🏼  😄

PR #263, opened 2021-04-06
Improve benchmark test cases

Chemaclass:
## 🔖 Changes

I improved a little bit the "Bench scenarios" that we currently have.

They don't assert anything yet, but I think it's a starting point. I think we'll be able to use them properly soonish. Right now I am also learning this tool :)


jenshaase:
Would it make sense to add that to the CI pipeline? Found something similar here: https://github.com/doctrine/instantiator/blob/1.4.x/.github/workflows/phpbench.yml

We probably don't have any reporting on it, but we can be sure that the benchmarks are still working.

JesusValera (approved):

PR #265, opened 2021-04-26
Using composer install in the CI

Chemaclass:
## 🔖 Changes

- Changing the CI to use composer install instead of update

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           coverage: none
 
       - name: Install dependencies
-        run: composer update --no-interaction --no-ansi --no-progress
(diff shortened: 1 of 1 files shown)


JesusValera:
What is the difference @Chemaclass?

mabasic:
@JesusValera It is much faster because update does some dependency update logic which is memory and CPU intensive. Install checks the requirements and  downloads packages = much faster. 

Maybe the CI can then use a cache layer to avoid downloading the dependencies each time, this would be even faster. 

Not sure about the part how CI works.

jenshaase:
We mainly did this because with `update` we got some Psalm updates into each Pull Request that broke the CI (without changing anything). This will not happen with `install`

jenshaase (approved):

Issue #271, opened 2021-05-05
Can't print TransientVector


mabasic:
This is related to PR #246.

I needed to convert transient vector to a persistent vector for it to be able to print out.

```
Printer can not print this type: Phel\Lang\Collections\Vector\TransientVector
```

Does this feature need to be implemented on the printer or is this the expected behavior?

PR #272, opened 2021-05-06
Allow print transients data structures

Chemaclass:
## 📚 Description

Issue: https://github.com/phel-lang/phel-lang/issues/271 
Currently, it's not possible to print any transient data structures. 
In this PR I basically use the `transient->persistent()->getIterator()` logic to be able to iterate over the transient data and this way allows printing it in the easiest way I could think of.

## 🔖 Changes

- Make `TransientVector`, `TransientMap`, and `TransientHashSet` printable

#### Follow up

I will start thinking about a potential refactoring to simplify the transient/persistent interfaces, unifying the common functionality in another interface that will be shared between them both. I didn't want to apply such an idea in this PR to don't overcomplicate it already.

### DIFF
--- a/src/php/Lang/Collections/Vector/TransientVectorInterface.php
+++ b/src/php/Lang/Collections/Vector/TransientVectorInterface.php
@@ -6,11 +6,13 @@ namespace Phel\Lang\Collections\Vector;
 
 use ArrayAccess;
 use Countable;
+use IteratorAggregate;
 
 /**
  * @template T
+ * @extends IteratorAggregate
  */
-interface TransientVectorInterface extends Countable, ArrayAccess
+interface TransientVectorInterface extends Countable, ArrayAccess, IteratorAggregate
 {
     public const BRANCH_FACTOR = 32;
     public const INDEX_MASK = self::BRANCH_FACTOR - 1;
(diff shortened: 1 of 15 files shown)


src/php/Lang/Collections/Vector/TransientVectorInterface.php
```diff
@@ -6,11 +6,13 @@
 
 use ArrayAccess;
 use Countable;
+use IteratorAggregate;
 
 /**
  * @template T
+ * @extends IteratorAggregate
```
JesusValera, line 13: Why only `IteratorAggregate`?

JesusValera (approved):

src/php/Lang/Collections/Vector/TransientVectorInterface.php
Chemaclass: Actually, I think this line is not necessary at all here...

jenshaase:
I didn't like this solution. It will print the transient data structure the same way as it prints the persistent data structure. If you print both you can not see a difference.

Transients are more for usage in internal functions. They shouldn't be passed around or printed somewhere. If you want to print it for debugging purposes I would recommend to convert it to a persistent version first and then print the persistent version.

Chemaclass:
I see your point and it makes totally sense. You can basically print transient data by simply converting it to persistent. 
For example:
```clojure
(def transient-vector (transient [1 2 3 4]))
(println (persistent transient-vector))
; [1 2 3 4]

(def transient-map (transient {:key-1 100 :key-2 "bar"}))
(println (persistent transient-map)) 
; {:key-1 100 :key-2 bar}
```
Therefore I will close this PR.

PR #274, opened 2021-05-14
Split Emitter and Evaluator

jenshaase:
## 📚 Description

This refactoring splits the emitter and the evaluator in different namespaces.

## 🔖 Changes

- Remove Evaluator from Emitter namespace
- Create Evaluator namespace


Chemaclass (approved):
Yes! Good refactoring, indeed 👌🏻

PR #277, opened 2021-05-16
Adding filter option to phel test command

Chemaclass:
## 📚 Description

Adding a new concept to the phel commands: "options". With this idea, we can extend the functionality of the different commands that we already have in a simple way.

The command arguments get divided by "inputs" and "options". The difference is basically that the options start with `--`

## 🔖 Changes

- Added an option `--filter=...` to the Test command

### DIFF
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -173,7 +173,11 @@
   (let [loc (location form) e (gensym)]
     `(try ,(assert-expr form message)
           (catch Throwable ,e
-            (report @{:state :error :location ,loc :message ,message :form ',form :exception ,e})))))
+            (report @{:state :error
+                      :location ,loc
+                      :message ,message
+                      :form ',form
+                      :exception ,e})))))
 
 # --------------
 # Defining tests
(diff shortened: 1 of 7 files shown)


src/phel/test.phel
```diff
@@ -298,19 +302,21 @@
           :when (get meta :test)]
       (get-in php/$GLOBALS ["__phel" munge-ns fn-name]))))
 
-(defn- run-test [ns]
+(defn- run-test [options ns]
```
jenshaase, line 305: You can use the destructuring here `[{:filter filter-options} ns]`

src/phel/test.phel
```diff
@@ -298,19 +302,21 @@
...
   (let [rt (php/:: \Phel\Runtime\RuntimeSingleton (getInstance))
         _ (php/-> rt (loadNs ns))
-        tests (find-test-fns ns)]
-    (dofor [test :in tests]
+        tests (find-test-fns ns)
```
jenshaase, line 308: I would do the following here:

1. Change `deftest` to `...(defn ,name {:test true :test-name ,name} ...` (I added `:test-name`)
2. Pass the `filter-options` to `find-test-fns` and do the filtering there. You get the test name with `(get meta :test-name)`

The `BOUND` variable is currently used for internals only (It is used to print better exceptions).

src/php/Command/Test/TestCommandOptions.php
```diff
@@ -0,0 +1,38 @@
...
+    public static function empty(): self
+    {
+        return new self();
+    }
+
+    public static function fromArray(array $options): self
+    {
+        $self = new self();
+
+        foreach ($options as $option) {
```
jenshaase, line 22: I would recommend to use `getopt` (https://www.php.net/manual/de/function.getopt.php) or symfony/console  for parsing command line arguments. Everything else will probably fail (edge cases)

src/php/Command/Test/TestCommandOptions.php
```diff
@@ -0,0 +1,38 @@
...
+            if (false !== strpos($option, self::FILTER_OPTION)) {
+                $self->filter = str_replace('-', '_', substr($option, strlen(self::FILTER_OPTION)));
+            }
+        }
+
+        return $self;
+    }
+
+    public function asPhelHashMap(): string
```
jenshaase, line 31: This looks ugly and there are probably some edge case where it will fail. Why not make a new instance of the Map class (using TypeFactory) and then use the Printer to print the result?

jenshaase (changes requested):

src/php/Command/Test/TestCommandOptions.php
Chemaclass: I agree. Let's talk in private to agree on this. Let's sync first on the way of how do we want to handle inputs and options.

Chemaclass:
After some brainstorming, we (@jenshaase and I) agreed on closing this PR in favor of:
1st PR: using `symfony/console` command for all commands
2nd PR: expand the commands themselves (adding options or whatever)

PR #283, opened 2021-05-21
Replace can not by cannot

JesusValera:
## 📚 Description

Replace `can not` by `cannot` in the message errors.

This word is correct only when it's written as a single word, you can check some information in this [link](https://www.grammarly.com/blog/cannot-or-can-not/). 😄 

- No logic was altered


Chemaclass (approved):
Thanks!

jenshaase (approved):

jenshaase:
Thanks!

PR #286, opened 2021-05-23
Disallows unexpected keywords in ns

### LINKED ISSUE #254: Phel not throwing error when miss typing a word in `ns`
I had this code:

```
(:require app\config\mail)
  (:requre app\logger))

(logger/debug "something")
```

Notice the `:requre` word. When using `phel run` to run the file it would not complain about `:requre`, but it would say:

```
Phel\Compiler\Emitter\Exceptions\CompiledCodeIsMalformedException: Can not resolve symbol 'logger/debug'
```

If it complained about `:requre` I would not spend half an hour trying to figure out what is wrong. Could this be somehow handled?

p810:
## 📚 Description

Fixes #254 by disallowing any unexpected keywords in `ns`, i.e. anything besides `:use` or `:require`.

## 🔖 Changes

- Adds a condition to handle any keyword that isn't `:use` or `:require` by raising an exception


jenshaase (approved):
Thank. Looks very good! I will merge it.

PR #290, opened 2021-05-25
Fixes #289: Add a transient map wrapper

### LINKED ISSUE #289: Can't put more than 16 key/value pairs in a hash map
Steps to reproduce:

```
(let [x (transient {})] (for [i :range [0 20]] (put x i i)) (persistent x))
```

returns:

```
{0 0 1 1 2 2 3 3 4 4 5 5 6 6 7 7 8 8 9 9 10 10 11 11 12 12 13 13 14 14 15 15}
```

jenshaase:
## 📚 Description

Bug fix for #289 

## 🔖 Changes

-  Add a transient map wrapper thats wrap TransientArrayMap and TransientHashMap

### DIFF
--- a/src/php/Lang/Collections/Map/PersistentArrayMap.php
+++ b/src/php/Lang/Collections/Map/PersistentArrayMap.php
@@ -136,12 +136,14 @@ class PersistentArrayMap extends AbstractPersistentMap
         }
     }
 
-    public function asTransient(): TransientArrayMap
+    public function asTransient(): TransientMapWrapper
     {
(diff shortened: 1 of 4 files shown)


Chemaclass (approved):
The code seems good, but what was the actual problem? I mean, where was the bug?

jenshaase:
The problem was that the `TransientArrayMap` not only returns itself but also a `TransientHashMap` when it gets to big. However, the Phel code was not build to update the newly created reference. Therefore I created the wrapper so that the reference stays the same.

PR #291, opened 2021-05-26
Refactoring test.phel

Chemaclass:
## 🔖 Changes

- Extract some code from from `report` into 2 functions: `print-report-state`, `should-report-println`
- Use hash-maps instead of arrays everywhere but on the global private variable `stats`. I left `stats` as an array in order to be able to mutate its state and keep using `(update-in stats [:counts :total] inc)`. Maybe there is a better alternative, @jenshaase, in order to use a hash-map in this context as well?

### DIFF
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -10,23 +10,28 @@
 # ------
 
 (def- stats @{:failed []
-              :counts @{:failed 0
-                        :error 0
-                        :pass 0
-                        :total 0}})
+              :counts {:failed 0
+                       :error 0
+                       :pass 0
+                       :total 0}})
+
+(defn- print-report-state [state]
+  (print (case state
+           :failed "F"
+           :pass "."
+           :error "E")))
+
+(defn- should-report-println [total-columns]
+  (= (% (get-in stats [:counts :total]) total-columns) 0))
 
 (defn report [data]
-  (let [@{:state state :type type} data
-        ok (= state :pass)]
-    (print
-     (case state
-       :failed "F"
-       :pass "."
-       :error "E"))
+  (let [{:state state :type type} data
+        ok (= state :pass)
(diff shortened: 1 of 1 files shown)


Chemaclass:
Right, it's a good point, but we could still apply this minor refactoring, so the left part will be still open to be refactored once we have that new concept, right?

jenshaase:
Yes, I will review it.

src/phel/test.phel
```diff
@@ -10,23 +10,28 @@
...
+                       :pass 0
+                       :total 0}})
+
+(defn- print-report-state [state]
+  (print (case state
+           :failed "F"
+           :pass "."
+           :error "E")))
+
+(defn- should-report-println [total-columns]
```
jenshaase, line 24: Rename to `report-newline?`. Functions that return a boolean should end with `?`

src/phel/test.phel
```diff
@@ -209,38 +214,38 @@
 
 (defn- print-error [data]
   (print-error-headline data)
-  (let [@{:exception exception :form form} data
+  (let [{:exception exception :form form} data
         printer (php/:: TextExceptionPrinter (create))]
     (println "              Test:" form)
-    (println " throwed exception:" (php/get_class exception))
+    (println "  thrown exception:" (php/get_class exception))
```
jenshaase, line 220: This should be in the past form. Is `threw` the correct form?

jenshaase (changes requested):

src/phel/test.phel
JesusValera: Yes, https://en.bab.la/conjugation/english/throw

Chemaclass: I will use `threw` instead.

jenshaase:
Looks good. I merged it!

PR #292, opened 2021-05-29
Use the CompilerFacade form the IntegrationTests

Chemaclass:
## 📚 Description

Use directly the Facade in order to compile phel code into the expected PHP code.

## 🔖 Changes

- Extract to its usage level (on the `Runtime`) the `file_get_content()` from the `FileCompiler::compile()`, so this function will be agnostic from the source itself (former filename) but now rather it receives the "phelCode" to be compiled to PHP.
- Rename `FileCompiler` to `CodeCompiler`.
- Minor cleanup, changing the argument order from `CompilerFacade::lexString()`

## 💡 Follow up

`EvalCompiler` and `CodeCompiler` seem pretty similar. I will try to figure out a way how to combine them if possible in another iteration.


src/php/Runtime/Runtime.php
```diff
@@ -204,7 +204,8 @@ private function loadCachedFile(string $filename, string $ns): bool
      */
     protected function loadFile(string $filename, string $ns): void
     {
-        $code = $this->compilerFacade->compile($filename);
+        $content = file_get_contents($filename);
+        $code = $this->compilerFacade->compile($content, $filename);
```
JesusValera, line 208: Makes sense to pass the `filename` in both methods?

Chemaclass: Yes. The `compiler()' second argument is the source of the content, it's something required/needed by the lexer in order to identify the source of such, but it's independent of the content itself. This way, the source it's actually an optional argument (being `string` the default value).

tests/php/Integration/Compiler/Reader/ReaderTest.php
```diff
@@ -758,10 +759,10 @@ public function testShortFnRestArgumentMultipleTimes(): void
     /**
      * @return TypeInterface|string|float|int|bool|null
      */
-    private function read(string $string, bool $removeLoc = false)
+    private function read(string $string, bool $withLocation = false)
```
jenshaase, line 762: Shouldn't the the default argument of `$withLocation` also change to `true`

tests/php/Integration/Fixtures/Call/global-call.test
```diff
@@ -12,12 +12,12 @@ $GLOBALS["__phel"]["user"]["x"] = new class() extends \Phel\Lang\AbstractFn {
 };
 $GLOBALS["__phel_meta"]["user"]["x"] = \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
   new \Phel\Lang\Keyword("start-location"), \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
-    new \Phel\Lang\Keyword("file"), "Call/global-call.test",
+    new \Phel\Lang\Keyword("file"), "string",
```
jenshaase, line 15: Could we keep the filenames here and in every other test?

jenshaase (changes requested):

tests/php/Integration/Compiler/Reader/ReaderTest.php
Chemaclass: Yes, why not. That thing + changing the line below where this is used. I will do that.
```php
$tokenStream = $this->compilerFacade->lexString($string, Lexer::DEFAULT_SOURCE, $withLocation); # <------
```

tests/php/Integration/Fixtures/Call/global-call.test
Chemaclass: With the first implementation that I did, it wasn't possible, but I just realize that the second option (the one that it's pushed in this PR) is actually possible and super easy, of course! I will change this 💡

jenshaase (approved):

jenshaase:
Thanks. I merged it.

PR #293, opened 2021-05-29
Merged core lib into one file and removed the load function

jenshaase:
## 📚 Description

Merge core lib into one file and the remove the load function.

Reason: The load function is problematic for implementing the compile command, since the load function is a evaluated at runtime and to at compile time.

## 🔖 Changes

- Merge core lib into one file
- Removed `load` function


Chemaclass (approved):
Because the `load` function seems to be a bit problematic, let get rid of it for now. 
I am totally ok with it 👍

PR #294, opened 2021-05-29
Moved namespace extractor code to the compiler module

jenshaase:
## 📚 Description

Moved namespace extractor code to the compiler module

## 🔖 Changes

- NamespaceExtractor now return a NsNode and not just a string.

### DIFF
--- a/src/php/Command/CommandConfig.php
+++ b/src/php/Command/CommandConfig.php
@@ -14,7 +14,10 @@ final class CommandConfig extends AbstractPhelConfig
      */
     public function getTestDirectories(): array
     {
-        return $this->get('tests', []);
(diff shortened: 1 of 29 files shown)


Chemaclass (changes requested):
Looking at the result now, moving the NamespaceExtractor from the Runtime to the Compiler module, it seems that we are putting too many things inside the Compiler module... Therefore I created a PR (pointing to this PR) extracting the NamespaceExtractor to its own module: https://github.com/phel-lang/phel-lang/pull/295
I think I prefer having its own module ☝️. What do you think, @jenshaase?

Chemaclass (approved):

PR #300, opened 2021-06-02
Minor clean-up on Code|EvalCompiler

Chemaclass:
## 📚 Description

**CodeCompiler vs EvalCompiler**
- Does it makes sense to DRY the `EvalCompiler`?
- Can we use the `CodeCompiler` inside the `EvalCompiler` in order to avoid code duplication?

The answer is "nope". I think these two compilers have different responsibilities and they have different purposes and, actually, different logic and behavior inside. They look a bit similar (indeed; they share part of the namespace `Compiler`) but so far I wasn't able to find a way to combine them in a good way.

I like how it is for now. Maybe I will come back to this one in the future.

## 🔖 Changes

- Apply some minor clean-up on `CodeCompiler` & `EvalCompiler` in order to improve readability.


jenshaase (approved):
Looks good.

PR #304, opened 2021-06-04
Make NamespaceExtractor more independent

jenshaase:
## 📚 Description

This Refactoring makes the NamespaceExtractor more independent from the compiler module.

## 🔖 Changes

- Return a `NamespaceInformation` object instead of the `NsNode`
- Move NamespaceExtractor classes from the compiler module to the NamespaceExtractor module.

### DIFF
--- a/src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
+++ b/src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace Phel\NamespaceExtractor\Extractor;
 
-class NamespaceInformation
+final class NamespaceInformation
 {
     private string $file;
     private string $namespace;
--- a/src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
+++ b/src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\NamespaceExtractor\Extractor;
+
+class NamespaceInformation
+{
+    private string $file;
+    private string $namespace;
+    /** @var string[] */
+    private array $dependencies;
+
+    public function __construct(string $file, string $namespace, array $dependencies)
+    {
+        $this->file = $file;
+        $this->namespace = $namespace;
+        $this->dependencies = $dependencies;
+    }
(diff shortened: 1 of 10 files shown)


src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
```diff
@@ -0,0 +1,35 @@
...
+namespace Phel\NamespaceExtractor\Extractor;
+
+class NamespaceInformation
+{
+    private string $file;
+    private string $namespace;
+    /** @var string[] */
+    private array $dependencies;
+
+    public function __construct(string $file, string $namespace, array $dependencies)
```
Chemaclass, line 14: We are setting three different properties, but we are using only one "the dependencies" https://github.com/phel-lang/phel-lang/pull/304/files#diff-ca9649ee607c19610338b1cce1a1a09b9e2fa3545eddf22787c15d8dfe0c3dfbR134
Do we need the other properties? Or are we using them somewhere else?

src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
```diff
@@ -0,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\NamespaceExtractor\Extractor;
+
+class NamespaceInformation
```
Chemaclass, line 7: Please, remember doing new classes [final](https://chemaclass.medium.com/final-classes-in-php-9174e3e2747e) according to our standard of closing classes by default in order to encourage composition over inheritance :)

Chemaclass (commented):
I am wondering where are we using these three properties of this new `NamespaceInformation` object. But, in general, I like the idea.

src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
jenshaase: I will use them in the Compile Command. I haven't done enough to publish a draft PR for that. But I thought I put out this PR already so it will not get to big.

src/php/NamespaceExtractor/Extractor/NamespaceInformation.php
jenshaase: done

Chemaclass (approved):

Issue #305, opened 2021-06-05
"PHP Notice:  Only variables should be passed by reference" when push'ing using a raw vector


Chemaclass:
When passing a vector directly to `push` (not using a variable, but a raw vector) you get an ugly PHP Notice/Warning.
```clojure
(push [1 2 3] 4)
```

<img width="1014" alt="Screenshot 2021-06-05 at 13 30 05" src="https://user-images.githubusercontent.com/5256287/120890311-25b84680-c602-11eb-857a-47688ec74d00.png">
<img width="1022" alt="Screenshot 2021-06-05 at 13 30 18" src="https://user-images.githubusercontent.com/5256287/120890317-2cdf5480-c602-11eb-9e38-ef0c73cf68f4.png">

## Findings

The message comes from the push method itself. Because the first element is flagged as a reference 
(it will be compiled to `&$xs`). This is in general not needed for vectors or maps. It is only needed for PHP Arrays. 
We can either remove the reference flag or live with the Notice. Let's evaluate those options... 🤔 

```clojure

(defn push
  "Inserts `x` at the end of the sequence `xs`."
  [^:reference xs x]
  (cond ... etc
```

PR #306, opened 2021-06-05
Remove passing by ref the array to the push function

Chemaclass:
## 📚 Description

Related to https://github.com/phel-lang/phel-lang/issues/305 

You are currently not able to exec push with a non-variable as 1st argument. Well, you get a really annoying NOTICE 
`(push [] 1)`. The reason was that the `push` function was using the iterable argument as a reference.

## 🔖 Changes

- Remove passing by reference the 1st argument in the phel `push` function.

## 🖼️  Screenshots

Screenshot 2021-06-06 at 00 28 15

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 * Removed `load` function in `phel\core`
+* Pass by value the array (1st argument) to `push` (#306)
 
 ## 0.3.3 (2021-06-04)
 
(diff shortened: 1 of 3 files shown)


jenshaase (changes requested):
I rethought about it. I would just remove the `^:reference` tag from the argument. It still can support PHP array. The difference is than that you get back a copy of the array and not the same reference. This probably also makes more sense because it matches the behavior with the immutable data structures.

jenshaase (changes requested):
Can you add a notice to the Changelog.md file that the behavior for PHP arrays has changed.

jenshaase (approved):

jenshaase:
Thanks. I merged it now.

PR #307, opened 2021-06-06
Use a placeholder to read the last result from REPL

Chemaclass:
## 📚 Description

If the last result from the REPL is a scalar value, you will be able to use it in your terminal by simply using the `$_` char.

## 🔖 Changes

- Add an `InputResult` class that is able to read the input buffer from the terminal and replace the placeholder `$_` to the previous result

## 🖼️ Screenshots

Some examples using simple math operations.
As you can see, `println` function returns `nil`, so that's the latest produced value. 
Some other examples using string and bool types:

Screenshot 2021-06-06 at 13 31 52

### To be considered

Screenshot 2021-06-06 at 13 32 24


jenshaase:
This is good idea. However, the underscore is already used for unused variables in let expressions.

```
(let [a 10
        _ 20]
  a)
```

We should use another symbol for that. Can you evaluate how it is done in other languages?

Chemaclass:
In Scala is with the `.`
```scala
scala> List(1,2,3,4)
res0: List[Int] = List(1, 2, 3, 4)

scala> .sum
res1: Int = 10
```
In Python is `_` 
```python
>>> [1,2,3,4]
[1, 2, 3, 4]
>>> sum(_)
10
```

In Clojure  (https://clojure.org/reference/repl_and_main#_launching_a_repl)
*1, *2, *3 - hold the result of the last three expressions that were evaluated
```clojure
user=> (+ 1 1)
2
user=> *1
2
```

It seems that there is no common "solution" for this.  We currently use `$X` for [short anon-func](https://phel-lang.org/documentation/functions-and-recursion/#anonymous-function-fn) (`|(+ $1 $2)`), therefore I like this one pretty much: `$_`. Or do you have in mind some other suggestions that might worth considering?

jenshaase (changes requested):
I think it would be better if `$_` is a actual Phel variable and not an internal state of the REPL command. Then you don't need all this replacement logic. You can simply use `$_` as variable in the code. However, I'm currently not sure what is the best solution to make it happen.

Chemaclass:
Well, that makes sense somehow... but, we currently don't have such a concept (bindings) and I don't see an easier way to implement this for now (related to your points that you mention above). I think we could take this PR and have this feature, and in the future consider refactoring and adapt it to the newer code that we might have.
From what I see is that this PR brings already some good value at a low cost (I don't mind this newly introduced logic for what it actually does). What do you say?

jenshaase:
However, the problem is that `$_` can a be anywhere. It will also be replaced if it is inside a string. Example:
```
> [1 2]
[1 2]
> "test$_"
??? <-- What is the output here?
```

Chemaclass:
Yeah, that's right. That's why I chose that weird combination 😂
I see two possible solutions for this now:
1. either we simply merge this PR being aware of this, and I will work on a new PR solving this problem (using regex or something like that.
2. don't merge this PR until we have solved this problem.

I personally prefer the 1st option, because of the low risk of this tech debt that we are introducing, BUT I totally get it if you say to go with the 2nd option. I will work on it next week.

jenshaase:
I would not merge it now. We better do it right instead of half right.

Chemaclass:
Ups, sorry, I will close this PR because it is out of control (I honestly don't know how) 😂 
I will create a new one.

PR #310, opened 2021-06-21
Use `$_` to get latest result from the REPL

Chemaclass:
## 📚 Description

If the last result from the REPL is a scalar value, you will be able to use it in your terminal by using `$_`.

## 🔖 Changes

- Add an `InputResult` class that is able to read the input buffer from the terminal and replace the placeholder (dollar sign + underscores)`$_` to the previous result.

## 🖼️ Screenshots

Some examples using simple math operations.
As you can see, `println` function returns `nil`, so that's the latest produced value. 
Some other examples using string and bool types:

image

To be considered: you cannot use `$_` to use a function, only scalar values.

image

Edge case: avoid replacing the value inside double quotes (string)

image

### DIFF
--- a/src/php/Command/Repl/InputResult.php
+++ b/src/php/Command/Repl/InputResult.php
@@ -1,0 +1,69 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Command\Repl;
+
+final class InputResult
+{
+    private const NO_VALUE = 'no_value';
+    private const LAST_RESULT_PLACEHOLDER = '__';
+
+    /** @var ?mixed */
+    private $lastResult;
+
+    /**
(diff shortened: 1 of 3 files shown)


JesusValera (approved):
Changed to `$_`. Thanks to someone from [stackoverflow](https://stackoverflow.com/questions/68531232/unable-to-replace-by-something-else-in-preg-replace?noredirect=1#comment121114235_68531232) 😄

Chemaclass:
As we agreed on this previous PR https://github.com/phel-lang/phel-lang/pull/307 I will merge this one because it's fulfilling our expectations, @jenshaase.

jenshaase:
Ok!

PR #316, opened 2021-07-03
Adding phpbench tests

Chemaclass:
## 🔖 Changes

- Add some benchmark tests for Run & Test commands.
- Add `phpbench-base` and `phpbench-ref` as composer scripts, to be able to run phpbench across different branches. 

### How can you run the same PHPBench tests on different branches? 

Tagging PHPBench's result and referencing it later.

1.- create a reference on the baseline branch, for example, master:
```bash
git checkout master 
composer phpbench-base
```
2.- checkout to the another branch and run PHPBench referencing the previous baseline:
```bash
git checkout your-branch 
composer phpbench-ref
```

This is still a prof-of-concept. The goal could be to be able to run this comparison in the CI on a PR level to ensure the performance of the current logic won't be affected negatively.


JesusValera (approved):
Really interesting 👌🏼

One question, if a specific new branch takes longer than the `main`, the CI will notify you about that, but in the end, it is up to you to merge on master, isn't it?

Chemaclass:
@JesusValera right now we don't run phpbench in the CI. That thing that I wrote tagging the phpbench results I don't know how to do that now, but it is the aiming goal to go.
And yes, of course, merging a PR is the responsibility of the dev, but if the CI turns out red, then first of all I would worry and investigate why that occurred.

jenshaase:
Looks interesting. I have to try it.

PR #318, opened 2021-07-24
Add spaceship operator function

Chemaclass:
## 📚 Description

All native PHP sorting functions should return an int and not a bool value.

```
usort(): Returning bool from comparison function is deprecated, 
return an integer less than, equal to, or greater than zero. 
Aka: the spaceship operator instead.
```

Source: https://wiki.php.net/rfc/stable_sorting

Therefore, currently, when running the tests from the core using PHP 8.0+ you can see this warning notice:

Screenshot 2021-07-24 at 14 15 09

And this was the result of using the function `<` as a sorting function in one test -> `(sort @[3 2 1 2 3] <)`

## 🔖 Changes

- Add a new core function as an alias for the new PHP spaceship operator `<=>`
- Use this new function in that previous "sorting test" with that warning
- Add another function `>=<` for the inversing sorting operation like `<=>`

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -486,6 +486,16 @@ Calling the and function without arguments returns true."
       (recur (first more) (next more))
       false)))
 
+(defn <=>
+  "Alias for the spaceship PHP operator in ascending order. Returns an int."
+  [a b]
+  (php/<=> a b))
+
+(defn >=<
+  "Alias for the spaceship PHP operator in descending order. Returns an int."
+  [a b]
+  (php/<=> b a))
+
 (defn all?
   "Returns true if `(pred x)` is logical true for every `x` in `xs`, else false."
   [pred xs]
--- a/tests/phel/test/core/sequence-functions.phel
+++ b/tests/phel/test/core/sequence-functions.phel
@@ -162,7 +162,8 @@
 
 (deftest test-sort
   (is (= @[1 2 3] (sort @[3 2 1])) "sort")
-  (is (= @[3 3 2 2 1] (sort @[3 2 1 2 3] <)) "sort descending order")
+  (is (= @[1 2 2 3 3] (sort @[3 2 1 2 3] <=>)) "sort ascending order")
+  (is (= @[3 3 2 2 1] (sort @[3 2 1 2 3] >=<)) "sort descending order")
   (is (= @[1 2 3] (sort-by identity @[3 2 1])) "sort-by identity")
   (is (= @[3 2 1] (sort-by - @[3 2 1])) "sort-by reversed"))
 


JesusValera (approved):
🎲🦙

tests/phel/test/core/sequence-functions.phel
```diff
@@ -162,7 +162,8 @@
 
 (deftest test-sort
   (is (= @[1 2 3] (sort @[3 2 1])) "sort")
-  (is (= @[3 3 2 2 1] (sort @[3 2 1 2 3] <)) "sort descending order")
```
Chemaclass, line 165: The `<` here was triggering the `PHP Deprecated` warning.

src/phel/core.phel
```diff
@@ -486,6 +486,16 @@ Calling the and function without arguments returns true."
       (recur (first more) (next more))
       false)))
 
+(defn <=>
+  "Alias for the spaceship PHP operator in ascending order. Returns an int."
+  [a b]
+  (php/<=> a b))
+
+(defn >=<
```
jenshaase, line 494: Not sure about this function. The spaceship operator is already weird. But if you come from PHP land it may be familiar to you.

There is an easy alternative to do descending order:
```
(sort @[3 2 1 2 3] (comp - <=>))
```

What do you think. Should we keep this function or not?

jenshaase (changes requested):

src/phel/core.phel
JesusValera: That's true the space-ship operator is weird, and this new one is even more, but I find it like a 'short-cut'.
It's easier and faster to type (and remember) `>=<` than `(comp - <=>)`.

I would say to put the `>=<` in the documentation with a small explanation of what it really means.

Chemaclass: This is a really tricky question. I don't really like any; not the `(comp - ...)` nor the new symbol..., but I think this is something we should address in our docu (in the best way as possible) for the final user. Basically because it's a potential real use case as you can see in its tests.

Now the question is: do we create a new symbol that could help the dev to express this intention in a simplified way (aka: the reverse way of `<=> / >=<`), or do we use a composition of a function (aka `comp - <=>`)?

### Pros vs Const for the new function
*Pros*:
- faster, due to use native PHP code. I don't know now how much faster, but I would say a lot faster than using the other thing
- easier to write and remember (at least when used once)
*Const*:
- we are introducing a new phel core func (so it might affect the compiling time; at least until we have a compiler command)
- it's a weird function (the reverse from the normal spaceship operator)

I think I prefer to keep it, specially because after having the compiler command that won't be a problem. Also, because it will be in the wiki/docu page. Also, you can simply not use it and use the other version if you want.

Having these arguments on the table, what do you think, Jens? If you still think we better remove it (at least for now) I will remove it, no problem, but I these are my thoughts about it for now.

jenshaase: I will keep it.

PR #326, opened 2021-08-12
Create CommandFacade and use it as entry point instead of CommandFactory

Chemaclass:
## 📚 Description

The Command module doesn't have a Facade, so we are using the Factory directly in the `phel` script 
(and in the integration tests).

## 🔖 Changes

- Add CommandFacade to make clear what are the entry points from the Command module.

### DIFF
--- a/phel
+++ b/phel
@@ -1,26 +1,26 @@
 #!/usr/bin/env php
 <?php
 
-$slash = DIRECTORY_SEPARATOR;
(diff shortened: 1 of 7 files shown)


jenshaase (approved):

jenshaase:
Looks good. Thanks.

JesusValera (approved):

PR #330, opened 2021-08-22
Add config platform php 7.4 to composer json

JesusValera:
## 📚 Description

Add in `composer.json`
```json
"config": {
  "platform": {
    "php": "7.4"
  }
},
```

## 💡 Benefits

Now it's possible to keep with PHP 7.4 as the "official" PHP version on `Phel` BUT we can use locally a greater version like PHP 8 (which is actually like 4 times faster), in addition we won't break the CI when we run `composer update` command 😄


Chemaclass (approved):

jenshaase (approved):

jenshaase:
Thanks I merged it.

PR #331, opened 2021-08-22
Allow optional argv in RunCommand

Chemaclass:
## 📚 Description

Before we could access the `argv` global values from PHP using the function:
```clojure
(def argv
  "Array of arguments passed to script."
  (php/aget php/$GLOBALS "argv"))
  ```

But, since we are using the Symfony console command, we don't allow using additional arguments when running phel scripts. 

Screenshot 2021-08-23 at 00 04 57

## 🔖 Changes

- Add to the `RunCommand` one more argument as `InputArgument::IS_ARRAY` and `OPTIONAL` ([source](https://symfony.com/doc/2.3/components/console/introduction.html#using-command-arguments)).

### DIFF
--- a/src/php/Command/Run/RunCommand.php
+++ b/src/php/Command/Run/RunCommand.php
@@ -45,6 +45,11 @@ final class RunCommand extends Command
                 'path',
                 InputArgument::REQUIRED,
                 'The file path that you want to run.'
+            )->addArgument(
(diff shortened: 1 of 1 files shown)


JesusValera (approved):

jenshaase:
Thank you.

PR #332, opened 2021-08-24
Allow Gacela php config 

Chemaclass:
## 📚 Description

The idea is to allow defining the project-config values from a separated PHP file and not in composer. Maybe we could also think about deprecating the "composer way" for the config, but for now (for BC) I would focus on keeping both things.

## 🔖 Changes

- Allow using Gacela's default PHP config setup before looking inside the composer.json extra config values.
- Keep full backward compatibility (BC).

## 🤔 Interesting to consider 

This is not valid for the **loader/loader-dev** config values, because they are used for the `phel-composer-plugin` which reads the `composer.json` file and reads these values in order to generate the `vendor/PhelRuntime.php`. For this, we need to think a bit more. Maybe it makes sense (maybe not) but that's another story.

Actually, this current PR won't alter at all the user experience. They will still keep using the **loader/loader-dev** for their namespaces, and we (in phel-lang project) can define internal phel config values in `phel-config.php`, outside the `composer.json` file.


config/default.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+return [
```
jenshaase, line 5: If we do configuration in PHP I want it to be in an object orientated way:

```
return (new ProjectConfiguration())
  ->setTests('tests/phel/')
  ->setExport(
    (new ExportConfiguration())
      ->setDirectories(['src/phel'])
      ->setNamespacePrefix('PhelGenerated')
      ->setTargetDirectory('src/PhelGenerated')
   );
```

Is that possible?

jenshaase:
One question left: The configuration is now in a `config` directory. Why don't we have the config files in the main directory as all the others like `phpunit.xml.dist`, `php-cs-fixer.dist.php` or `phpbench.json`. I would recommend that we have a `phel.config.php` file or something similar?

src/php/Config/ProjectConfiguration.php
```diff
@@ -0,0 +1,48 @@
...
+    {
+        return [
+            CommandConfig::TESTS => $this->testConfig->getDirectories(),
+            InteropConfig::EXPORT => [
+                InteropConfig::EXPORT_DIRECTORIES => $this->exportConfig->getDirectories(),
+                InteropConfig::EXPORT_NAMESPACE_PREFIX => $this->exportConfig->getNamespacePrefix(),
+                InteropConfig::EXPORT_TARGET_DIRECTORY => $this->exportConfig->getTargetDirectory(),
+            ],
+        ];
+    }
```
Chemaclass, line 47: I would prefer to create a separate follow-up PR to apply this change because this will break the current BC:
```php
return [
    'tests_directories' => $this->testsDirectories,
    'export_directories' => $this->exportConfig->getDirectories(),
    'export_namespace_prefix' => $this->exportConfig->getNamespacePrefix(),
    'export_target_directory' => $this->exportConfig->getTargetDirectory(),
];
```
> TL;DR: I want to keep the BC in this PR.

jenshaase (approved):

jenshaase:
I merged it. Thanks.

Issue #341, opened 2021-09-06
Repl startup script is loaded with every command


jenshaase:
The REPL startup script (startup.phel) is loaded when a command is executed and not only when the REPL is executed.

The following code in `CommandFactory::createReplCommand()` does that:
```php
$this->getRuntimeFacade()
            ->getRuntime()
            ->loadFileIntoNamespace('user', $this->getConfig()->getReplStartupPhel());
```

The code needs to be moved inside the REPL command (e.g. execute method)

PR #342, opened 2021-09-09
Load replStartupFile on ReplCommand->execute()

Chemaclass:
Issue: https://github.com/phel-lang/phel-lang/issues/341

## 📚 Description

The REPL startup script (`startup.phel`) is loaded when any command is executed and not ONLY when the REPL command is executed.

## 🔖 Changes

Move the `loadFileIntoNamespace()` from the `CommandFactory` (which was being called for every command when creating all commands in the phel entry point script) to the `ReplCommand::execute()` function


jenshaase (approved):

jenshaase:
Thank for fixing this.

PR #343, opened 2021-09-11
Using arg unpacking in config dirs setters

Chemaclass:
## 📚 Description

Right now you can pass either a string or an array of strings. I think it is simpler to just allow `string ...$dirs` using the variadics function (native PHP). I wrote last year a post about this: [Typed arrays in PHP
](https://chemaclass.es/blog/typed-arrays-php/)

## 🔖 Changes

It unifies the argument to `setDirectories(string ...$directories)`. There is no need to pass an array itself when calling that method, but it will be received as an array of strings. For example:

```php
return (new ProjectConfiguration())
    ->setTestConfiguration((new TestConfiguration())
        ->setDirectories('path-1/'))
    ->setExportConfiguration((new ExportConfiguration())
        ->setDirectories('path-1/', 'path-2/', 'path-3/'));
```

### DIFF
--- a/phel-config.php
+++ b/phel-config.php
@@ -10,6 +10,6 @@ return (new ProjectConfiguration())
     ->setTestConfiguration((new TestConfiguration())
         ->setDirectories('tests/phel/'))
     ->setExportConfiguration((new ExportConfiguration())
-        ->setDirectories(['src/phel'])
+        ->setDirectories('src/phel')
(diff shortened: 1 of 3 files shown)


JesusValera (approved):
Much better this way 👌🏼

jenshaase (approved):

jenshaase:
Thanks. I merged it.

PR #344, opened 2021-09-14
Added build facade to compile files in correct order

jenshaase:
## 📚 Description

This PR is the first direction for a compile command. This PR changes the way how Phel code is executed. Before we use the function `loadNs` to load and compile files that are needed to execute the requested file. It was more of a bottom-up approach. Now we will first compile the code before it is executed. Files with no dependencies will now be compile before their dependent files, which is a top-down approach.


src/php/Command/Test/TestCommand.php
```diff
@@ -73,13 +71,31 @@ public function addRuntimePath(string $namespacePrefix, array $path): self
     public function execute(InputInterface $input, OutputInterface $output): int
     {
         try {
-            /** @var list<string> $paths */
+            /** @var array<string> $paths */
```
Chemaclass, line 74: Actually, I would prefer to keep using `list` for lists of things.
> list: represents continuous, integer-indexed arrays (always start from index zero) like: [“red”, “yellow”, “blue”] 

Check out this page to find more information about the [Modern PHPDoc Annotations](https://suckup.de/2020/02/modern-phpdoc-annotations/).

src/php/Runtime/Runtime.php
```diff
@@ -85,6 +95,9 @@ public function exceptionHandler(Throwable $exception): void
      */
     public function loadNs(string $ns): bool
     {
+        return true;
```
Chemaclass, line 98: 😱😜

jenshaase: I have to check a few edge cases first before removing this completely :)

src/php/Command/Test/TestCommand.php
jenshaase: Done

src/php/Runtime/Runtime.php
jenshaase: I removed all calls to `loadNs` is this project. However, the Phel Composer Plugin still call this method, so I can not fully remove it.

src/php/Build/BuildFacade.php
```diff
@@ -41,4 +42,127 @@ public function getNamespaceFromDirectories(array $directories): array
...
+     *
+     * @return NamespaceInformation[]
+     */
+    public function getDependenciesForNamespace(array $directories, array $ns): array
+    {
+        $namespaceInformation = $this->getNamespaceFromDirectories($directories);
+
+        $index = [];
+        $queue = [];
+        foreach ($namespaceInformation as $info) {
```
Chemaclass, line 62: We don't want to have any logic in the Facade. The Facade should be like a "proxy" delegating all logic to domain classes that the Factory creates for it. I created a [PR](https://github.com/phel-lang/phel-lang/pull/347) pointing to this one showing what do I mean by this.

The main reason is that you cannot easily write tests for the Facade. Actually, you cannot write unit tests for the Facade because it is part of the infrastructure layer (like the Factory or Config, for example).  For this reason, you want to have isolated classes with the logic of your domain, and using the Factory to create your domain classes injecting all collaborators/dependencies that your domain classes need. 

☝️ This way allows you to create independent classes with logic that is really easy to test.

Chemaclass (changes requested):

PR #345, opened 2021-09-15
Use raw arrays for config

Chemaclass:
## 📚 Description

We realized that the `ProjectConfiguration` classes are a huge boilerplate that doesn't bring much value. So we will use raw arrays for config instead.

The "end array" has the same structure as the previous one, so it's BC.

## 🔖 Changes

- Remove all Config classes in favor of simple raw key-values array for phel project configuration


JesusValera (approved):
Ok 🙂

jenshaase (approved):

PR #346, opened 2021-09-18
Create RuntimeCommand and remove phel-composer-plugin dependency

Chemaclass:
## 💡 Goal

Move all phel config to `phel-config.php` and deprecate the composer extra config fields in favor of the PHP phel config.

## 📚 Description

We want to get rid of the `phel-composer-plugin` dependency and move all project config to `phel-config.php`. And follow up we could start thinking about how could we define the config in a phel itself. Kind of `phel-config.phel`. But this is out of the scope of this PR.

## 🔖 Changes

- Create `RuntimeCommand`
- Remove `phel-composer-plugin` dependency from `composer.json`


src/php/Command/Runtime/ConfigNormalizer.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Command\Runtime;
+
+final class ConfigNormalizer
+{
+    /**
+     * @param array<string, list<string>> $result [ns => [path1, path2, ...]]
```
jenshaase, line 10: Could you describe what this function is doing?

What type of `array` is `$configLoader`?

src/php/Command/Runtime/RuntimeCommand.php
```diff
@@ -0,0 +1,89 @@
...
+
+        return $result;
+    }
+
+    /**
+     * @return array<string, list<string>>
+     */
+    private function loadVendorConfig(array $result): array
+    {
+        $pattern = $this->applicationRootDir . 'vendor/*/*/' . self::PHEL_CONFIG_FILE_NAME;
```
jenshaase, line 78: In composer it is possible to change the location of the `vendor` directory (https://getcomposer.org/doc/06-config.md#vendor-dir). Therefore, we should have an options to change the `vendor` directory too. I would suggest we have a config value for that. Default to `vendor`'

src/php/Command/Runtime/RuntimeCommand.php
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * @return array<string, list<string>>
+     */
+    private function loadRootConfig(array $result): array
+    {
+        $rootPhelConfigPath = $this->applicationRootDir . self::PHEL_CONFIG_FILE_NAME;
+
+        if (is_file($rootPhelConfigPath)) {
+            $pathPrefix = '/..';
```
jenshaase, line 64: Can you explain this to me. Why this value?

jenshaase (changes requested):

src/php/Command/Runtime/RuntimeCommand.php
Chemaclass: I took that value from here: https://github.com/phel-lang/phel-composer-plugin/blob/master/src/DumpRuntime.php#L63 when `isRootPackage === true`

```php
$pathPrefix = $isRootPackage ? '/..' : '/' . $package->getName();
```

src/php/Command/Runtime/ConfigNormalizer.php
Chemaclass: I will do that 👍

src/php/Command/Runtime/RuntimeCommand.php
```diff
@@ -0,0 +1,134 @@
...
+        $loaderConfig = [
+            $this->loadRootConfig(),
+            $this->loadVendorConfig(),
+        ];
+
+        /** @var array<string, list<string>> $flatConfig [ns => [path1, path2, ...]] */
+        $flatConfig = array_merge(...$loaderConfig);
+
+        file_put_contents(
+            $this->applicationRootDir . 'vendor/PhelRuntime.php',
```
jenshaase, line 60: We need to replace `vendor` here too.

jenshaase (changes requested):

src/php/Command/Runtime/RuntimeCommand.php
Chemaclass: Right, thanks! I will change it

jenshaase (approved):

PR #347, opened 2021-09-19
Refactor BuildFacade; extract to classes the logic from the facade

Chemaclass:
## 📚 Description

We don't want to have any logic in the Facade. The Facade should be like a "proxy" delegating all logic to domain classes that the Factory creates for it.

The main reason is that you cannot easily write tests for the Facade. Actually, you cannot write unit tests for the Facade because it is part of the infrastructure layer (like the Factory or Config, for example). That said, of course, it's always a good idea to have integration tests that test the Facade methods (at least with the happy paths), because these are the endpoints/usages of a module, from beginning to end.

For this reason, you want to have isolated classes with the logic of your domain, and using the Factory to create your domain classes injecting all collaborators/dependencies that your domain classes need. 

☝️ This way allows you to create independent classes with logic that is really easy to test.

## 🔖 Changes

- Apply "extract class refactoring" to extract the logic from the Facade to its own classes.

### DIFF
--- a/src/php/Build/Compile/ProjectCompiler.php
+++ b/src/php/Build/Compile/ProjectCompiler.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Build\Compile;
+
+use Phel\Build\Extractor\NamespaceExtractorInterface;
+
+final class ProjectCompiler
+{
+    private NamespaceExtractorInterface $namespaceExtractor;
+
+    private FileCompiler $fileCompiler;
+
+    public function __construct(
+        NamespaceExtractorInterface $namespaceExtractor,
+        FileCompiler $fileCompiler
+    ) {
+        $this->namespaceExtractor = $namespaceExtractor;
+        $this->fileCompiler = $fileCompiler;
+    }
+
+    /**
+     * @return list<CompiledFile>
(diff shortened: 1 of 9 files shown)


src/php/Build/Compile/ProjectCompiler.php
```diff
@@ -0,0 +1,48 @@
...
+    public function compileProject(array $srcDirectories, string $dest): array
+    {
+        $namespaceInformation = $this->namespaceExtractor->getNamespacesFromDirectories($srcDirectories);
+
+        $result = [];
+        foreach ($namespaceInformation as $info) {
+            $targetFile = $dest . '/' . $this->getTargetFileFromNamespace($info->getNamespace());
+            $targetDir = dirname($targetFile);
+            if (!file_exists($targetDir)) {
+                mkdir($targetDir, 0777, true);
```
Chemaclass, line 35: I am aware that you cannot write a unit test for this class because there is `IO` involved (creation of a directory). It would be necessary to extract this call to an interface and inject that via the constructor. But I didn't want to refactor to that level. In this PR I just wanted to apply the minimum "extract-class-refactoring" to show the basic pattern about our module's design and why.

JesusValera (approved):
💡👌🏼

PR #348, opened 2021-09-20
Remove Runtime::loadNs function

jenshaase:
## 📚 Description

This function removes the `Runtime::loadNs` function. It is not needed anymore.


Chemaclass (approved):
Woho! Removing code is always good! 🙌

Issue #349, opened 2021-09-21
Error `Runtime` could not be loaded


JesusValera:
I think since this PR https://github.com/phel-lang/phel-lang/pull/346, phel is not working anymore.
When you try to run the `composer install` command, the `Runtime` file is not generated and with the `phel test` command, everything explores 😄 

You can investigate just simply cloning a new project, installing the dependencies and executing `composer test-all` for example.

```bash
> ./phel test
Phel\Runtime\Exceptions\PhelRuntimeException: The Runtime could not be loaded from: /home/jesus/Desktop/phel-lang//vendor/PhelRuntime.php
in /home/jesus/Desktop/phel-lang/src/php/Runtime/Exceptions/PhelRuntimeException.php:13 (gen: /home/jesus/Desktop/phel-lang/src/php/Runtime/Exceptions/PhelRuntimeException.php:13)

#0 /home/jesus/Desktop/phel-lang/src/php/Runtime/RuntimeFactory.php(29): Phel\Runtime\Exceptions\PhelRuntimeException::couldNotBeLoadedFrom()
#1 /home/jesus/Desktop/phel-lang/src/php/Runtime/RuntimeFacade.php(16): Phel\Runtime\RuntimeFactory->getRuntime()
#2 /home/jesus/Desktop/phel-lang/src/php/Compiler/CompilerFactory.php(70): Phel\Runtime\RuntimeFacade->getRuntime()
```

Chemaclass:
Did you try to generate the `PhelRuntime.php` file with `phel runtime` command?

JesusValera:
Ok, I see!

I will close the issue then.

PR #350, opened 2021-09-21
Fix Runtime exception

JesusValera:
## 📚 Description

I think this PR solves properly the https://github.com/phel-lang/phel-lang/issues/349 issue.

Please, take a look in order to check everything is fine :)

### ⚠️ Some notes

There are two classes related to `Runtime`: 

1. `Phel\Runtime\Runtime`
2. `Phel\Runtime\RuntimeSingleton`

I used the Singleton with the help of the static `initialize()` method (without params, not sure what does `GlobalEnvironmentInterface` do).

Finally, I removed some useless classes like the configuration or the exception, as those classes won't be used anymore because this refactorization.


src/php/Runtime/RuntimeFactory.php
```diff
@@ -5,30 +5,15 @@
...
-        $runtimePath = $this->getConfig()->getApplicationRootDir()
-            . DIRECTORY_SEPARATOR . 'vendor'
-            . DIRECTORY_SEPARATOR . 'PhelRuntime.php';
-
-        if (!file_exists($runtimePath)) {
-            throw PhelRuntimeException::couldNotBeLoadedFrom($runtimePath);
-        }
-
-        return require $runtimePath;
+        return RuntimeSingleton::initialize();
```
Chemaclass, line 17: I am not quite sure this is what we want... 🤔 because this way you would ignore the configuration that knows which paths and namespaces your program is aware of. Previously that config was under composer extra fields but now is in your `phel-config.php` file at the root level.

Previously the composer plugin was generating this file (when you did an install or update), but we deprecated that thing and now we generate the PhelRuntime manually when the user wants (see [PR](https://github.com/phel-lang/phel-lang/pull/346)). This is your new friend: `phel runtime`

Chemaclass (commented):
Although the idea was nice, the PR is not solving your "problem" because the vendor/PhelRuntime is what you actually need. You basically missed running manually the new command: `phel runtime`.

PR #351, opened 2021-09-22
Load during runtime the config paths without needing a static PhelRuntime file

Chemaclass:
## 📚 Description

Load the Runtime on the fly. 

## 🔖 Changes

- Remove `phel runtime` command.
- Create `RuntimeLoader` which add the paths from the "project & vendor" config (using the `loader` key)


Chemaclass:
After talking to @jenshaase, we decided that the goal is to get rid of the Runtime module completely. The reason is that the runtime is something that is existing when the code is executed. And we need these directories (src, tests & vendor) only for the compilation and not during execution.

I close this PR in favor of this other one: https://github.com/phel-lang/phel-lang/pull/353

PR #352, opened 2021-09-23
Use phel-test as a prefix for tests/phel directory

Chemaclass:
## 📚 Description

Use phel-test as a prefix for tests/phel directory, because just "phel" is already in use for the production phel library itself.

Also, I think we can end up with some collision of naming when unifying all src and test keys when doing ns extraction per file. But anyhow, I think we get some benefits already by having diff namespaces for different contexts. And in this case, for test context, it makes totally sense.

## 🔖 Changes

- Change `phel\\` to `phel-test\\` the `loader-dev` phel test namespace

### DIFF
--- a/phel-config.php
+++ b/phel-config.php
@@ -8,7 +8,7 @@ return [
         'phel\\' => 'src/phel/',
     ],
     'loader-dev' => [
-        'phel\\' => 'tests/phel/',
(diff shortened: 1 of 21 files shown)


JesusValera (approved):
👍🏼

jenshaase:
Yes, that is good. I will merge it.

PR #353, opened 2021-09-25
EPIC: Remove Runtime and refactoring commands

jenshaase:
## 📚 Description

- Remove `Command/Runtime`
- Create a `Config` module which has access to the `src-dirs`, `test-dirs`, and `vendor-dir`. Not the export config values.
- `Interop module` uses all export config, and we could start thinking about doing something similar for other commands.

#### 🔖  Tasks

- [x] `Interop` module will contain the `ExportCommand` https://github.com/phel-lang/phel-lang/pull/354
- [x] `Formatter` module will contain the `FormatCommand` https://github.com/phel-lang/phel-lang/pull/355
- [x] `Run` module will contain the `ReplCommand`, `RunCommand`, and `TestCommand` https://github.com/phel-lang/phel-lang/pull/356
- [x] Remove Runtime module https://github.com/phel-lang/phel-lang/pull/358

Inside the module, we can create a directory named `Command`, and inside that directory, we can find the commands related to that module.

Pros: each module will handle its config key-values their own way. We don't need a `Config` module, but each module is responsible for its own configuration.


Chemaclass (approved):
From a "monolith module" with too many responsibilities to splitting up that "Command" module into other modules which have their own console commands as part of their responsibilities.

Also the deprecation of the `composer.json key-values` and now using our own `phel-config.php` files

And not just that! But also now there is no need to have a PhelRuntime anymore.

I already have some follow-up refactoring ideas to do but it can be postponed as it's not blocking this EPIC idea anymore.

Amazing outcome. This is indeed a nice milestone for the project!

PR #354, opened 2021-09-26
Move ExportCommand inside Interop module

Chemaclass:
## 📚 Description

This branch is from `feature/config-facade` in order to split the scope of the PR and make the review process easier.

## Goal

Move the commands closer to their "proper" modules and scope

## 🔖 Changes

- Move ExportCommand inside Interop module

---

@jenshaase, if you like this approach, you can merge this one and I can work on a follow-up branch to do something similar for the other commands.


jenshaase (approved):
I like the idea.

PR #364, opened 2021-10-03
Move the src Finder from the Run to the Command module

Chemaclass:
## 📚 Description

This way we can access `src-dirs` + `vendor` dependencies from other modules (like Interop, so we can use it in the `ExportCommand` as well).

I think the `src-dirs`, `test-dirs`, and vendor would be interesting information for all general commands. For example, the `ExportCommand` (from the Interop module) needs to know about the `src-dirs` in order to compile the export code properly.

## 🔖 Changes

- Move the Finder folder from the `Run` to the `Command` module.

## 🖼️ Screenshots

Some screenshots from the phel-scaffolding project.

#### BEFORE (current master)
Screenshot 2021-10-03 at 11 43 08

#### AFTER (using this branch)
Screenshot 2021-10-03 at 11 43 37

### DIFF
--- a/src/php/Interop/ExportFinder/FunctionsToExportFinder.php
+++ b/src/php/Interop/ExportFinder/FunctionsToExportFinder.php
@@ -7,6 +7,7 @@ namespace Phel\Interop\ExportFinder;
 use Phel\Build\BuildFacadeInterface;
 use Phel\Build\Extractor\ExtractorException;
 use Phel\Build\Extractor\NamespaceInformation;
+use Phel\Command\CommandFacadeInterface;
 use Phel\Compiler\Evaluator\Exceptions\CompiledCodeIsMalformedException;
 use Phel\Compiler\Evaluator\Exceptions\FileException;
 use Phel\Compiler\Exceptions\CompilerException;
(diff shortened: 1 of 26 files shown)


src/php/Interop/ExportFinder/FunctionsToExportFinder.php
```diff
@@ -54,17 +55,30 @@ public function findInPaths(): array
...
+        ));
+
+        $dependenciesForNamespace = [];
+        foreach ($namespaces as $namespace) {
+            $dependenciesForNamespace[] = $this->buildFacade->getDependenciesForNamespace(
+                [
+                    ...$this->commandFacade->getSourceDirectories(),
+                    ...$this->commandFacade->getVendorSourceDirectories(),
+                ],
+                [$namespace, 'phel\\core']
```
jenshaase, line 75: I don't think, we need the foreach loop here. Instead we can just past `[...$namespaces, 'phel\\core']` as argument of the `getDependenciesForNamespace`. Otherwise `$allNamespace` will contain duplicates that can lead to unexpected behavior.

jenshaase (changes requested):

src/php/Interop/ExportFinder/FunctionsToExportFinder.php
Chemaclass: Right! Good catch. I will update it.

PR #365, opened 2021-10-04
Setup gacela config on bootstrap

Chemaclass:
## 📚 Description

We want to simplify the gacela custom config, so other projects don't need to create a custom `gacela.php` config file.

## 🔖 Changes

- Remove the `gacela.php` config file
- Setup the custom config on `Gacela::bootstrap()` in the `phel` entry point


jenshaase:
Cool!

PR #368, opened 2021-10-14
Compile command

jenshaase:
## 📚 Description

Create the new compile command to compile the code of a project to PHP code.

## 🔖 Changes

- Added `out-dir` config parameter (default out)
- Added command in build module
- `CompilerFacadeInterface::compile` now returns a `EmitterResult` instead of a `string`
- FileEmitter emits different code at some points (compared to StatementEmitter).
  * Emit a `namespace` statement at the beginning of each file
  * Do not a emit a `namespace` statement for the struct
  * Emit `require_once` statements to import other Phel namespaces
- Only recompile changed files
  * When writing a compiled PHP file, set the timestamp to the same value as form the PHP file
  * Before compiling again, check if the Phel file is newer than the PHP file.
  * This behavior can be disabled with the `--no-cache` option.

### DIFF
--- a/src/php/Build/BuildFactory.php
+++ b/src/php/Build/BuildFactory.php
@@ -8,6 +8,7 @@ use Gacela\Framework\AbstractFactory;
 use Phel\Build\Command\CompileCommand;
 use Phel\Build\Compile\DependenciesForNamespace;
 use Phel\Build\Compile\FileCompiler;
+use Phel\Build\Compile\FileCompilerInterface;
 use Phel\Build\Compile\FileEvaluator;
 use Phel\Build\Compile\ProjectCompiler;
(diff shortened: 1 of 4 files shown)


jenshaase:
I can't reproduce the `16` output in the tests. It is not coming on my machine. It is also not appearing in the output of the GitHub actions.

I fixed the other issues you proposed.

Chemaclass:
That number `16` is really strange. It's not reproducible in the GitHub actions, but I can reproduce it by simply running:
```bash
php phel compile --no-cache
```
Screenshot 2021-10-21 at 02 51 09

@JesusValera, please try it on your computer and tell us if you can reproduce it or not. Maybe it's just my local machine 🤔

Issue #370, opened 2021-10-16
Could not find config key "vendor-dir" in "Gacela\Framework\Config"


mabasic:
I just assumed that the default value `vendor` will be used and this it is not necessary to have it explicitly provided in the config. Am I wrong?

Issue #369, opened 2021-10-16
Could not find config key "export" in "Gacela\Framework\Config"


mabasic:
I have started updating my libraries and applications to Phel 4.0 and this is the first issue that I have come upon.

```
PHP Fatal error:  Uncaught Gacela\Framework\Exception\ConfigException: Could not find config key "export" in "Gacela\Framework\Config" in /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/gacela-project/gacela/src/Framework/Exception/ConfigException.php:21
Stack trace:
#0 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/gacela-project/gacela/src/Framework/Config.php(82): Gacela\Framework\Exception\ConfigException::keyNotFound()
#1 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/gacela-project/gacela/src/Framework/AbstractConfig.php(22): Gacela\Framework\Config->get()
#2 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/phel-lang/phel-lang/src/php/Interop/InteropConfig.php(25): Gacela\Framework\AbstractConfig->get()
#3 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/phel-lang/phel-lang/src/php/Interop/InteropFactory.php(43): Phel\Interop\InteropConfig->getExportTargetDirectory()
#4 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/phel-lang/phel-lang/src/php/Interop/InteropFactory.php(32): Phel\Interop\InteropFactory->createDirectoryRemover()
#5 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/phel-lang/phel-lang/src/php/Interop/InteropFacade.php(17): Phel\Interop\InteropFactory->createExportCommand()
#6 /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/phel-lang/phel-lang/phel(34): Phel\Interop\InteropFacade->getExportCommand()
#7 {main}
  thrown in /home/mabasic/code/src/github.com/mabasic/phel-json/vendor/gacela-project/gacela/src/Framework/Exception/ConfigException.php on line 21
```

Nowhere in the documentation says that `export` field is required.

mabasic:
I have managed to solve this by adding:

```
'export' => []
```

to the `phel-config.php` file.

But then I am seeing this warning:

```
[...]

PR #371, opened 2021-10-16
Enable config default values

Chemaclass:
## 📚 Description

Issues: 
https://github.com/phel-lang/phel-lang/issues/369: Could not find config key "export" in "Gacela\Framework\Config"
https://github.com/phel-lang/phel-lang/issues/370: Could not find config key "vendor-dir" in "Gacela\Framework\Config"

## 🔖 Changes

- Add default values for all current keys in `phel-config.php` file


jenshaase:
Should we catch the `Gacela\Framework\Exception\ConfigException` in our code a throw custom "Phel" exception?

jenshaase:
Ok sounds good. I will merge it now.

PR #375, opened 2021-10-30
Added internal compile mode definition

jenshaase:
## 📚 Description

This adds a `*compile-mode*` variable to Phel that will be set to true if a file is compiled. This flag can be used to prevent that side effects (like `print`) are executed while running the compile command.

## 🔖 Changes

- Added internal `*compile-mode*` definition
- Initial value is `false`
- Set the value to `true` in the `FileCompiler` class before the `compile` function is executed.
- Reset to `false` after the `compile` function is done.

### DIFF
--- a/src/php/Build/Compile/FileCompiler.php
+++ b/src/php/Build/Compile/FileCompiler.php
@@ -30,7 +30,9 @@ final class FileCompiler implements FileCompilerInterface
             ->setSource($src)
             ->setIsEnabledSourceMaps($enableSourceMaps);
 
+        $GLOBALS['__phel']['phel\core']['*compile-mode*'] = true;
         $result = $this->compilerFacade->compile($phelCode, $options);
+        $GLOBALS['__phel']['phel\core']['*compile-mode*'] = false;
 
         file_put_contents($dest, "<?php\n" . $result->getCode());
         file_put_contents(str_replace('.php', '.phel', $dest), $phelCode);
--- a/src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
+++ b/src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
@@ -36,7 +36,6 @@ final class GlobalEnvironment implements GlobalEnvironmentInterface
 
     public function __construct()
     {
-        // Add internal definition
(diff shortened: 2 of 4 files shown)


src/php/Build/Compile/FileCompiler.php
```diff
@@ -30,7 +30,9 @@ public function compileFile(string $src, string $dest, bool $enableSourceMaps):
             ->setSource($src)
             ->setIsEnabledSourceMaps($enableSourceMaps);
 
+        $GLOBALS['__phel']['phel\core']['*compile-mode*'] = true;
         $result = $this->compilerFacade->compile($phelCode, $options);
+        $GLOBALS['__phel']['phel\core']['*compile-mode*'] = false;
```
Chemaclass, line 35: I am a bit afraid of this `$GLOBALS` usage... I understand this is how we do it internally, but I think we could start thinking about encapsulating this usage somehow somewhere to have full control of this global variable and not usage it in different places.
What about creating a class on Phel's domain which will deal with this global PHP variable? this way it will be more explicit who usages this and how

src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
```diff
@@ -32,6 +34,12 @@ final class GlobalEnvironment implements GlobalEnvironmentInterface
 
     private bool $allowPrivateAccess = false;
 
+    public function __construct()
+    {
+        // Add internal definition
+        $this->addInternalDefinition('phel\core', Symbol::create('*compile-mode*'), false);
```
Chemaclass, line 40: Is this comment relevant? I mean, the function's name is already revealing that intention, no?

Chemaclass (approved):

src/php/Build/Compile/FileCompiler.php
jenshaase: Yes definitely, `$GLOBALS` should be removed in future.

src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
jenshaase: Removed the comment

PR #381, opened 2021-11-09
Support for namespaced keywords

jenshaase:
## 📚 Description

This PR adds a new language feature: namespaced keywords.

Currently keywords can be used like constants. For example `:foo` or `:bar` are two constants. A problem arise if multiple libraries use the same keywords for different purposes. Then the keywords name collide. In this PR I will introduce namespaced keywords, similar to keywords in Clojure.

This PR introduces the following syntax:

* `:foo` is a keyword without a namespace.
* `:bar/foo` is a keyword with the namespace `bar`.
* `::foo` is a keyword which uses the current namespaces as the namespace for the keyword (as defined by `ns`).
  ```
  (ns bar)
  ::foo # evals to :bar/foo
  ```
* `::bar/foo` Looks up the namespace alias `bar` in the current namespace and uses the this namespace for the keyword
  ```
  (ns foobar
    (:require abc\xyz :as bar))
  ::bar/foo # evaluates to :abc\xyz/bar
  ```

## 🔖 Changes

- Replaces all uses of `new Keyword` with `Keyword::create`
- Introduced `Keyword::createForNamespace`
- Updated the Parser to support namespaced Keywords using the `GlobalEnvironment`

### DIFF
--- a/src/php/Compiler/Parser/ExpressionParser/AtomParser.php
+++ b/src/php/Compiler/Parser/ExpressionParser/AtomParser.php
@@ -18,7 +18,7 @@ use Phel\Lang\Symbol;
 
 final class AtomParser
 {
-    private const KEYWORD_REGEX = '/:(:?)(([^\/]+)\/)?([^\/]+)/';
+    private const KEYWORD_REGEX = '/:(?<second_colon>:?)((?<namespace>[^\/]+)\/)?(?<keyword>[^\/]+)/';
 
     private GlobalEnvironmentInterface $globalEnvironment;
 
@@ -96,14 +96,14 @@ final class AtomParser
             throw new KeywordParserException('This is not a valid keyword');
         }
 
-        $isDualColon = $matches[1] === ':';
-        $hasNamespace = strlen($matches[3]) > 0;
+        $isDualColon = $matches['second_colon'] === ':';
+        $hasNamespace = $matches['namespace'] !== '';
 
         $namespace = null;
         if ($isDualColon && $hasNamespace) {
             // First case is a dual colon with a namespace alias
             // like ::foo/bar
-            $alias = $matches[3];
+            $alias = $matches['namespace'];
             $namespace = $this->globalEnvironment->resolveAlias($alias);
             if (!$namespace) {
                 throw new KeywordParserException("Can not resolve alias '$alias' in keyword: $word");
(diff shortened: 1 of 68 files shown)


src/php/Compiler/Parser/ExpressionParser/AtomParser.php
```diff
@@ -80,4 +86,47 @@ public function parse(Token $token): AbstractAtomNode
...
+    {
+        $word = $token->getCode();
+        $isValid = preg_match(self::KEYWORD_REGEX, $word, $matches);
+
+        if (!$isValid) {
+            throw new KeywordParserException('This is not a valid keyword');
+        }
+
+        $isDualColon = $matches[1] === ':';
+        $hasNamespace = strlen($matches[3]) > 0;
```
Chemaclass, line 100: Instead of `strlen() > 0`, you can check directly if it's an empty string ;)
```php
$hasNamespace = $matches[3] !== '';
```

src/php/Compiler/Parser/ExpressionParser/AtomParser.php
```diff
@@ -80,4 +86,47 @@ public function parse(Token $token): AbstractAtomNode
...
+
+        $namespace = null;
+        if ($isDualColon && $hasNamespace) {
+            // First case is a dual colon with a namespace alias
+            // like ::foo/bar
+            $alias = $matches[3];
```
Chemaclass, line 106: I see that we're using `$matches[x]` by position in the regex expression. Have you consider using aliases instead of indexes/numbers? Here you can see an example of how to use aliases: https://github.com/gacela-project/gacela/blob/master/src/Framework/ClassResolver/GlobalKey.php#L18 

I am not sure if it would be possible to use them in this context, but just as an idea that worth trying because an alias is more legible than a number.

src/php/Lang/Keyword.php
```diff
@@ -11,25 +11,32 @@ final class Keyword extends AbstractType implements IdenticalInterface, FnInterf
...
 
+    private ?string $namespace;
     private string $name;
     private int $hash;
 
     /** @var array<string, Keyword> */
     private static $refStore = [];
 
-    public function __construct(string $name)
+    public function __construct(?string $namespace, string $name)
```
Chemaclass, line 21: The constructor is not used anywhere else but in the named constructor inside this class. What about making it private (the constructor)?

src/php/Compiler/Parser/ExpressionParser/AtomParser.php
```diff
@@ -16,6 +18,15 @@
 
 final class AtomParser
```
Chemaclass, line 19: What about adding unit tests for this class? Especially the behaviour that you introduced on `parseKeyword()` is super interesting and complex, and it would be good to have a suite of unit tests covering those possible scenarios. 
This way we could think about refactoring the `KEYWORD_REGEX` using aliases or whatever else. It would be also great as documentation for the complexity here.

Chemaclass (approved):

PR #382, opened 2021-11-10
Support for variables

jenshaase:
## 📚 Description

This PR add a new language feature: Variables.

Sometime it is necessary to store a central value that is changing. Currently there was no support for this in Phel. Now you can define a variable to store a changing value. Example:

```
(def foo (var 10))

(deref foo) # Evaluates to 10
(set! foo 20) # Set foo to 20
(deref foo) # Evaluates to 20

(swap! foo + 2) # Evaluates to 22
(deref foo) # Evaluates to 22
```

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Support for variables
+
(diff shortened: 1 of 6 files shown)


Chemaclass (approved):
This is really cool! 😎

PR #384, opened 2021-11-13
Add deprecated note to old collection data structures

Chemaclass:
## 🔖 Changes

- Added `@deprected` to PhelArray, Set & Table classes in favor of the new persistent data structures (that were introduced here: https://github.com/phel-lang/phel-lang/pull/244)

### 🤔  Follow up

What are the next step we should do to get rid of these deprecated classes? I guess we should check that the core tests are using the persistent data structures, and the old ones aren't used anywhere else. Is there something else we need to keep in mind to do this, @jenshaase?

### DIFF
--- a/src/php/Lang/PhelArray.php
+++ b/src/php/Lang/PhelArray.php
@@ -11,6 +11,7 @@ use Iterator;
 use Phel\Printer\Printer;
 
 /**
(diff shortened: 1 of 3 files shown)


jenshaase (approved):

jenshaase:
I have no problem to remove them now completely. I don't think there is any usage for them currently.

PR #385, opened 2021-11-17
Define interface and let struct implement them

jenshaase:
## 📚 Description

This PR add a new language feature: interfaces.

It is now possible to define interfaces with `definterface`. It is possible to add zero or more function to a interface. Each function must have a least one parameter. This parameter is bound to `$this` in PHP. Examples:

```
(definterface MyInterfaceWithoutFunctions)

(definterface MyInterfaceWithOneFunction
  (foo [this]))

(definterface MyInterfaceWithTwoFunction
  (foo [this])
  (bar [this a b]))

(definterface MyInterfaceWithDocComments
  (foo [this] "my optional comment")
  (bar [this a b] "my other optional comment"))
```

Furthermore it is now possible to use these interface in structs. Structs can implement zero or more interfaces. Examples:

```
(definterface MyInterfaceWithoutMethods)

(defstruct* my-type-with-one-interface-without-methods [v]
  MyInterfaceWithoutMethods)
```

```
(definterface MyInterface
  (foo [this])
  (bar [this a b]))

(defstruct my-type-with-one-interface [v]
  MyInterface
  (foo [this] v)
  (bar [this a b] (+ a b v)))


(let [x (my-type-with-one-interface 1)]
  (foo)
  (bar 1 2))
```

```
(definterface MyFirstInterface
  (foo [this])
  (bar [this a b]))

(definterface MySecondInterface
  (foobar [this]))

(defstruct my-type-with-two-interfaces [v]
  MyFirstInterface
  (foo [this] v)
  (bar [this a b] (+ a b v))
  MySecondInterface
  (foobar [this] (php/-> this (foo))))

(let [x (my-type-with-two-interfaces 1)]
  (foo)
  (bar 1 2)
  (foobar))
```

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1743,7 +1743,7 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
                `(defn ,fn-name {:doc ,doc} ,args
                   (if (php/is_a ,(first args) ,class-name-str)
                     (php/-> ,(first args) (,munged-fn-symbol ,@(rest args)))
-                    (php/throw (php/new \InvalidArgumentException ,(str "Value doesn't implement interface " name)))
+                    (throw (php/new \InvalidArgumentException ,(str "Value doesn't implement interface " name)))
                     )))]
   `(do
      (definterface* ,name ,@fns)
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1694,6 +1694,10 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
           (finally
             ,@(map bind-value resets))))))
 
+# --------
+# Variable
+# --------
+
 (defn var
   "Creates a new variable with the give value"
   [value]
@@ -1721,3 +1725,26 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
         next (apply f current args)]
     (set! variable next)
     next))
+
+# ----------
+# Interfaces
+# ----------
+
+(defmacro definterface
+  "Defines a interface"
+  [name & fns]
+  (let [name-str (php/-> name (getName))
+        munge (php/new Munge)
+        class-name-str (php/. *ns* "\\" (php/-> munge (encode name-str)))
+        defs (for [[fn-name args doc] :in fns
+                   :let [fn-name-str (php/-> fn-name (getName))
+                         munged-fn-name (php/-> munge (encode fn-name-str))
+                         munged-fn-symbol (php/:: Symbol (create munged-fn-name))]]
+               `(defn ,fn-name {:doc ,doc} ,args
+                  (if (php/is_a ,(first args) ,class-name-str)
+                    (php/-> ,(first args) (,munged-fn-symbol ,@(rest args)))
+                    (php/throw (php/new \InvalidArgumentException ,(str "Value doesn't implement interface " name)))
+                    )))]
+  `(do
+     (definterface* ,name ,@fns)
+     ,@defs)))
--- a/src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
+++ b/src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
@@ -32,6 +32,9 @@ final class GlobalEnvironment implements GlobalEnvironmentInterface
     /** @var array<string, array<string, Symbol>> */
     private array $useAliases = [];
 
+    /** @var array<string, array<string, Symbol>> */
+    private array $interfaces = [];
+
     private bool $allowPrivateAccess = false;
 
     public function __construct()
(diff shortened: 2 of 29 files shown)


src/phel/core.phel
```diff
@@ -1721,3 +1725,26 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
...
+                         munged-fn-name (php/-> munge (encode fn-name-str))
+                         munged-fn-symbol (php/:: Symbol (create munged-fn-name))]]
+               `(defn ,fn-name {:doc ,doc} ,args
+                  (if (php/is_a ,(first args) ,class-name-str)
+                    (php/-> ,(first args) (,munged-fn-symbol ,@(rest args)))
+                    (throw (php/new \InvalidArgumentException ,(str "Value doesn't implement interface " name)))
+                    )))]
+  `(do
+     (definterface* ,name ,@fns)
+     ,@defs)))
```
Chemaclass, line 1750: This looks complex yet beautiful 👍

src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
```diff
@@ -223,21 +230,33 @@ private function resolveWithAlias(Symbol $name, NodeEnvironmentInterface $env):
...
             return new GlobalVarNode($env, $ns, $name, $def, $name->getStartLocation());
         }
 
         // Try to resolve in phel.core namespace
         $ns = 'phel\core';
+
+        // Try to resolve interfaces in phel.core
+        if (isset($this->interfaces[$ns][$name->getName()])) {
+            return new PhpClassNameNode($env, Symbol::createForNamespace($ns, $name->getName()), $name->getStartLocation());
+        }
```
Chemaclass, line 257: This hardcoded `$ns` 🤔 
I understand why we're doing this, and it's fine to leave it, but I wanted to write a comment here for a future iteration.

Some immediate thoughts about this: apparently the logic of the `$ns = $this->getNs();` and `$ns = 'phel\core';` are the same, the only difference is the value inside `$ns`. Also there are some comments explaining what the code does. So my first idea would be to create a private function to encapsulate this logic and pass the value of `$ns` instead. Maybe something like this: 

```php
resolveNamespace(Symbol $name, NodeEnvironmentInterface $env, string $ns): ?AbstractNode
```

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/DefInterfaceSymbol.php
```diff
@@ -0,0 +1,89 @@
...
+            return [];
+        }
+
+        $methods = [];
+        for ($forms = $list; $forms != null; $forms = $forms->cdr()) {
+            $method = $forms->first();
+
+            if (!$method instanceof PersistentListInterface) {
+                throw AnalyzerException::withLocation('Methods in definterface must be lists', $list);
+            }
```
Chemaclass, line 54: What about?
```php
    for ($forms = $list; $forms !== null; $forms = $forms->cdr()) {
        $first = $forms->first();

        if (!$first instanceof PersistentListInterface) {
            throw AnalyzerException::withLocation('Methods in definterface must be lists', $list);
        }

        $methods[] = $this->createDefInterfaceMethod($first);
    }

    return $methods;
}

private function createDefInterfaceMethod(PersistentListInterface $method): DefInterfaceMethod
{
    $name = $method->get(0);
    if (!$name instanceof Symbol) {
        throw AnalyzerException::withLocation('Method names must be symbols', $method);
    }

    $arguments = $method->get(1);
    if (!$arguments instanceof PersistentVectorInterface) {
        throw AnalyzerException::withLocation('Method arguments must be vectors', $method);
    }

    foreach ($arguments as $argument) {
        if (!$argument instanceof Symbol) {
            throw AnalyzerException::withLocation('A method argument must be symbol', $arguments);
        }
    }

    $comment = null;
    if (count($method) > 2) {
        $comment = $method->get(2);
        if (!is_string($comment)) {
            throw AnalyzerException::withLocation('Method comments must be strings', $method);
        }
    }

    return new DefInterfaceMethod(
        $name,
        $arguments->toArray(),
        $comment
    );
}
```

Chemaclass (approved):
I found some minor improvements that we can work on in a follow up PR. No real blockers, you can merge it, and I can work on the clean-follow-up PR later. Amazing work 👍

src/php/Compiler/Analyzer/Environment/GlobalEnvironment.php
Chemaclass: Done 👉🏼  https://github.com/phel-lang/phel-lang/pull/394

PR #400, opened 2021-12-11
Separate the CI tests on diff jobs

Chemaclass:
### 🤔 Background

We weren't using the real `matrix.operating-system` on the CI. We were using it only for the name, but not on the real field `runs-on: XXX`. Anyway, I was really wondering if we really need PHP for different OS... at least not for now. 

However, I was wondering if running the different type of tests on different jobs, so we could run them in parallel and check the timing as well for them independently. 

### 🔖 Changes

- Remove the `operating-system: [ 'ubuntu-latest', 'windows-latest', 'macos-latest' ]`, which we even didn't use properly 😂. Maybe we could reimplement this in the future if we find the need.
- Split the tests by types to run on different jobs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,6 +20,17 @@ jobs:
           php-version: 7.4
           coverage: none
 
+      - name: Get Composer Cache Directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: ${{ runner.os }}-composer-
+
       - name: Install dependencies
         run: composer install --no-interaction --no-ansi --no-progress
 
@@ -38,6 +49,17 @@ jobs:
         with:
           php-version: 7.4
           coverage: none
+
+      - name: Get Composer Cache Directory
+        id: composer-cache
+        run: echo "::set-output name=dir::$(composer config cache-files-dir)"
+
+      - name: Cache dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.composer-cache.outputs.dir }}
+          key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
+          restore-keys: ${{ runner.os }}-composer-
 
       - name: Install dependencies
         run: composer install --no-interaction --no-ansi --no-progress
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -49,14 +45,13 @@ jobs:
...
 
-  tests:
+  compiler-tests:
+    name: Compiler tests on PHP ${{ matrix.php }}
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        operating-system: [ 'ubuntu-latest', 'windows-latest', 'macos-latest' ]
         php: [ '7.4', '8.0' ]
```
JesusValera, line 54: Should we add also PHP 8.1 or let's wait until we migrate completely to PHP 8.0?

Chemaclass: Good idea, but I would first move to 8.0, and then you can add a follow up PR adding the 8.1, because we might have to fix some broken things (luckily not). But just in case, I would delay that for another iteration.

jenshaase: Not all our dependencies are 8.1 ready yet, I guess. So we can do that later in another PR

.github/workflows/ci.yml
```diff
@@ -76,8 +71,48 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php }}
+          coverage: none
+
+      - name: Install dependencies
```
jenshaase, line 91: Is it possible to install dependencies only once? I'm more aware with GitLab Pipelines than with GitHub. In GitLab you can cache the vendor dir and use it in other task again.

jenshaase (changes requested):

.github/workflows/ci.yml
Chemaclass: It might/should be possible, but I don't know it right now. I will investigate that and come back to this.

Chemaclass:
I found this docu: https://github.com/php-actions/composer#caching-dependencies-for-faster-builds and I updated our CI according what it says. The CI looks better now than before, but the "cache issue" doesn't seems to be working for some unknown reason 🤔 
Screenshot 2021-12-11 at 15 33 24

There is this "500 error" (see screenshot) and I cannot find the reason for it. Maybe could you help me out with this, @jenshaase?

jenshaase:
Looks like a problem with GitHub: https://github.com/actions/cache/issues/698

jenshaase:
Should we already merge this? The pipelines are still running even if the cache is not working currently.

jenshaase (approved):

Chemaclass:
Let's wait till tomorrow, and we can try one more time. And if the cache doesn't work, 
A) we can remove it from the CI or 
B) unify the tests again in one job
C) merge it as it is right now the PR

Chemaclass:
We still have the same error, but it's a non-blocker. I will merge it and we can keep an eye on this. Hopefully it will eventually work or it might need some minor adjustments :)

Issue #397, opened 2021-12-10
TextExceptionPrinter::create method is not available anymore


jenshaase:
In the test library the method `TextExceptionPrinter::create` is called:

https://github.com/phel-lang/phel-lang/blob/6fb71e33c8e8b114e38cb64f166092bfb1e166da/src/phel/test.phel#L217

 This method isn't available anymore.

I would recommend to make it accessible from the CommandFacade

Chemaclass:
There is no `TextExceptionPrinter::create()` anymore, that's right, but how is this even working? How is it possible that I don't see any PHP error when I run a failing test? 😮 How could you reproduce this bug, @jenshaase?
<img width="784" alt="Screenshot 2021-12-11 at 13 47 38" src="https://user-images.githubusercontent.com/5256287/145677016-2a003b4e-4f92-4aff-967b-60d509c9c048.png">

jenshaase:
The test must not just fail. It must throw an exception.

PR #402, opened 2021-12-11
Expose ExceptionPrinter from CommandFacade

Chemaclass:
### 🤔 Background

TextExceptionPrinter::create method is not available anymore #397

### 💡 Goal

Get rid of the PHP exception when the test has an exception on runtime

### 🔖 Changes

Use the CommandFacade to get the ExceptionPrinter in the `print-error` function.

I intentionally didn't add the `public function getExceptionPrinter(): ExceptionPrinterInterface` to the `CommandFacadeInterface` because I considered that we don't really want to expose that method to other modules, and the FacadeInterface is the contract that modules has between each other. If we end-up needing this function, we could add it, but till then I think it's ok keeping it like this.

### BEFORE
There is a PHP runtime exception due to calling an non-existing method.
Screenshot 2021-12-11 at 19 56 34

### AFTER
The Exception is being rendered properly.
Screenshot 2021-12-11 at 20 01 43

### DIFF
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -1,7 +1,7 @@
 (ns phel\test
   (:use Phel\Lang\Symbol)
   (:use Phel\Compiler\Emitter\OutputEmitter\Munge)
-  (:use Phel\Command\Shared\Exceptions\TextExceptionPrinter)
+  (:use Phel\Command\CommandFacade)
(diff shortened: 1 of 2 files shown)


jenshaase (approved):

jenshaase:
Thanks for the fix!

PR #404, opened 2021-12-18
Add CallablePrinter

Chemaclass:
### 🤔 Background

Right now when printing a function in the terminal we render the same as for anon-classes. 

### 💡 Goal

Render `<function>` when printing a function and not `<PHP-AnonymousClass>`

### 🔖 Changes

BEFORE
Screenshot 2021-12-18 at 19 47 20

AFTER
Screenshot 2021-12-18 at 19 46 39

### DIFF
--- a/src/php/Printer/Printer.php
+++ b/src/php/Printer/Printer.php
@@ -17,6 +17,7 @@ use Phel\Lang\Table;
 use Phel\Printer\TypePrinter\AnonymousClassPrinter;
 use Phel\Printer\TypePrinter\ArrayPrinter;
 use Phel\Printer\TypePrinter\BooleanPrinter;
+use Phel\Printer\TypePrinter\CallablePrinter;
 use Phel\Printer\TypePrinter\KeywordPrinter;
 use Phel\Printer\TypePrinter\NonPrintableClassPrinter;
 use Phel\Printer\TypePrinter\NullPrinter;
@@ -125,6 +126,9 @@ final class Printer implements PrinterInterface
         }
         if (method_exists($form, '__toString')) {
             return new ToStringPrinter();
+        }
+        if (is_callable($form)) {
+            return new CallablePrinter();
         }
         if ((new ReflectionClass($form))->isAnonymous()) {
             return new AnonymousClassPrinter();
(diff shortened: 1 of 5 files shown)


src/php/Printer/TypePrinter/CallablePrinter.php
```diff
@@ -0,0 +1,19 @@
...
+ * @implements TypePrinterInterface<object>
+ */
+final class CallablePrinter implements TypePrinterInterface
+{
+    /**
+     * @param object $form
+     */
+    public function print($form): string
+    {
+        return '<function>';
```
Chemaclass, line 17: I choose `<function>` because we already have these:
- `'<PHP-Object(...)>'`
- `'<PHP-AnonymousClass>'`

What do you say? I was also considering `:function`, because `(type fn)` says
```clojure
phel:2> (type (fn [] "Hello World"))
:function
```

jenshaase: `<function>` is better. `:function` is a keyword and that could be misleading. The following two lines would evaluate to the same result:

```
phel:1> (type (fn [] "Hello World"))
:function
phel:2> :function
:function
```

src/php/Printer/Printer.php
```diff
@@ -126,6 +127,9 @@ private function createObjectTypePrinter($form): TypePrinterInterface
         if (method_exists($form, '__toString')) {
             return new ToStringPrinter();
         }
+        if (is_callable($form)) {
```
jenshaase, line 130: I we only want to check for Phel function we should check if `$form` is an instance of `FnInterface`. Otherwise `php/substr` would also print `<function>`.

Don't know what's better.

src/php/Printer/TypePrinter/CallablePrinter.php
Chemaclass: I didn't think about it... now that you mention it, now I like to have `<function>`.

src/php/Printer/Printer.php
Chemaclass: I think it's a good idea to use `if ($form instanceof FnInterface) {...`.

jenshaase (approved):

PR #407, opened 2021-12-21
Removed deprecated table, array and set mutable data structures

jenshaase:
### 🤔 Background

In the very first release of Phel there were mutable data structure such as Table, Array and Set. These data structures have been deprecated with the introduction of the new persistent data structures.

### 💡 Goal

The goal of this PR is to remove all the deprecated data structures.

### 🔖 Changes

* Removed all PHP classes that were required for the old data structures.
* Removed all tests that used the old data structures
* Changed methods that still used the old data structures.


Chemaclass (approved):
wow! Awesome! 👏

tests/phel/test/html.phel
```diff
@@ -37,56 +36,38 @@
 
 (deftest arrays
   (is (= "<p>a</p><p>b</p>" (html [:p "a"] [:p "b"])))
-  (is (= "<body>foobar</body>" (html [:body @["foo" "bar"]])))
-  (is (= "<p>a</p><p>b</p>" (html @[[:p "a"] [:p "b"]]))))
+  # (is (= "<body>foobar</body>" (html [:body ["foo" "bar"]])))
+  # (is (= "<p>a</p><p>b</p>" (html [[:p "a"] [:p "b"]])))
```
JesusValera, line 40: Why is this commented? :)

JesusValera (approved):
Really amazing!

tests/phel/test/html.phel
Chemaclass: Well, so far I can see, this `(is (= "<body>foobar</body>" (html [:body ["foo" "bar"]])))` is not working :) 
Screenshot 2021-12-22 at 14 29 34

@jenshaase do we want to fix this before merging this PR? or could we do that in a follow up PR and that's why you just commented it out?

jenshaase: Ah thanks for the reminder. I wanted to check it later but forgot about it. I now fixed the tests.

PR #412, opened 2021-12-25
Keep Munge creation inside its module

Chemaclass:
### 🤔 Background

In the `Build` module, instead of creating a new instance of an object from the `Compiler` module, we should delegate that communication to the `CompilerFacade`.

### 💡 Goal

Don't create object from external modules directly inside a module.

### 🔖 Changes

- Add `encodeNs()` to the `CompilerFacade`
- Inject the `CompilerFacade` into the `ProjectCompiler`


JesusValera (approved):

jenshaase:
Great. Thank you!

PR #416, opened 2021-12-26
Fix formatter facade tests

Chemaclass:
### 🤔 Background

The test generator with name 'Map' was duplicated, like this:

```php
# line: 171
yield 'Map' => [
    [
        '{:x 1',
        ':y 2}',
    ],
    [
        '{:x 1',
        '  :y 2}',
    ],
];

# line: 221
yield 'Map' => [ 
    'actualLines' => ['{ :a 1 :b 2 }'],
    'expectedLines' => ['{:a 1 :b 2}'],
];
```

The funny thing was that the first one was a failing test (with a wrong expectation; it had two white spaces instead of one). While the second test was working as expected, and this was translated into having a non-running failing test on the fixtures.

Changing the name of any of these generators make the failed test explicit.

### 🔖 Changes

- Fix the failing `yield 'Map'` test, and renamed to `yield 'Indent map block'`.
- Remove unnecessary/duplicated test `yield 'Indent'`.
- Rename all yield names to something more descriptive.

### DIFF
--- a/tests/php/Integration/Formatter/FormatterFacadeTest.php
+++ b/tests/php/Integration/Formatter/FormatterFacadeTest.php
@@ -218,7 +218,7 @@ final class FormatterFacadeTest extends TestCase
             'expectedLines' => ['[1 2 3]'],
         ];
 
-        yield 'Map' => [
+        yield 'One liner Map' => [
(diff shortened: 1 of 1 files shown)


JesusValera (approved):
Nice catch and good renaming!

Issue #409, opened 2021-12-23
Create a way to load vendor/autoload.php from Phel code


jenshaase:
Currently there is no way to load the `vendor/autoload.php` file form Phel. The following two options are not working:

```
(ns hello-world\boot)
(php/require "vendor/autoload.php") # Too late: will be loaded after phel core is loaded
```

```
(php/require "vendor/autoload.php") # Not valid: First statement must be the `ns` statement
(ns hello-world\boot)
```

PR #421, opened 2022-01-09
Added feature to require php files in ns statement.

jenshaase:
This commit adds a new feature to load php files in the ns statement. This
can, for example, be used to load the composer autoload file. Example

```
(ns myapp
  (:require-file "vendor/autload.php"))
```

The :require-file import ins transformed into a require_once statement in PHP.
The file is loaded before any other :require import.

Fixes: #409

### DIFF
--- a/src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/NsSymbol.php
+++ b/src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/NsSymbol.php
@@ -50,6 +50,7 @@ final class NsSymbol implements SpecialFormAnalyzerInterface
         $this->analyzer->setNamespace($ns);
 
         $requireNs = [];
+        $requireFiles = [];
         for ($forms = $list->rest()->cdr(); $forms != null; $forms = $forms->cdr()) {
             $import = $forms->first();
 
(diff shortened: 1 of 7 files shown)


src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/NsSymbol.php
```diff
@@ -172,4 +175,14 @@ private function analyzeRequire(string $ns, PersistentListInterface $import): Sy
 
         return $requireSymbol;
     }
+
+    private function analyzeRequireFile(PersistentListInterface $import): string
+    {
+        $file = $import->get(1);
```
Chemaclass, line 181: What's the meaning of this `1`? Maybe would be better extracting that into a constant or writing a comment above giving some explanation about that number? 🤔

Chemaclass (approved):
Nice job! I got just one comment/question open.

mabasic:
Nice. Does this mean that we no longer need the php file in public for autoload and that we can move that to the boot file?

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/NsSymbol.php
jenshaase: `$import` is something like `(require-file "vendor/autoload.php")`. `$import->get(1)` is then `"vendor/autoload.php". Don't know if we can change this.

jenshaase:
@mabasic Yes you can import "vendor/autoload.php" now directly in Phel. But you need to be aware of the relative paths. The path of the compiled file may be different. You may write this:

```
# in src/boot.phel
(ns mycompany\myapp\boot
  (require-file "../vendor/autoload.php"))
```

However, the compiled file will be generated in `out/mycompany/myapp/boot.php`. The file contains then `require_once "../vendor/autoload.php"`. The autoload file will then be loaded from `out/mycomany/vendor/autoload.php` which is not correct. Currently I have no idea how to solve it. The best way would be to put the vendor directory into PHP's include path. Then you can just `(require-file "vendor/autoload.php")`.

PR #423, opened 2022-01-13
Introduce a Registry class to store the definitions.

jenshaase:
Currently all definitions have been stored in $GLOBALS. This seem to be a bit
of a bad design decision. This is solved with this PR. It introduces a new
Registry to store the definitions.

### 🔖 Changes

* The Registry class was added
* The Emitter have been changed to use the Registry
* All tests have been updated
* All code that used `$GLOBALS` changed to use the Registry


src/php/Lang/Registry.php
```diff
@@ -0,0 +1,77 @@
...
+
+    public function __construct()
+    {
+        $this->clear();
+    }
+
+    /**
+     * @return self
+     */
+    public static function getInstance()
```
JesusValera, line 28: Is there any reason this class is a Singleton?
It couldn't be possible to create from the factory? :)

JesusValera, line 28:
```php
public static function getInstance(): self
{
    //...
```
Is the return type `self` not working?

JesusValera (commented):
Awesome work! ✨

src/php/Lang/Registry.php
Chemaclass: There is no Factory inside the Lang module/directory. This Registry is the replacement for the `$GLOBALS` usage, so for now I think it's way better to use this Singleton instead.

Chemaclass (commented):
Nice 👌

PR #425, opened 2022-01-15
Override method getDefintionMetadata by getDefinitionMetadata

JesusValera:
### 📚 Description

Replace the string method name `getDefintionMetadata` to `getDefinitionMetadata` from PHP and Phel code.
(Notice the missing `i` on `Definition` word) 😄


Chemaclass (approved):
Nice catch 👁️

PR #426, opened 2022-01-17
Evaluate meta data in the special form def

jenshaase:
This commit evaluates the meta data that can be provided in def. This was
currently not a problem because all data stored in the map was literal data
that evaluates to itself.

Before this commit the code
```
(def x {:foo (+ 1 1)} 10)
```
was evaluated to
```
(def x {:foo '(+ 1 1)} 10) # The + operation is not evaluated
```
With the commit the code evaluates to
```
(def x {:foo 2} 10)
```
This required some changes to the GlobalEnvironment. The meta data
of a definition is now loaded from the Registry and is not stored in the
GlobalEnvironment itself. This has the drawback that the meta data
is only available after the definition was evaluated and not while analysing the
initial value of the definition. However, this only happens in recursive function
calls. In this case the meta data will be a empty persistent map.

### DIFF
--- a/src/php/Compiler/Emitter/OutputEmitter/NodeEmitter/MapEmitter.php
+++ b/src/php/Compiler/Emitter/OutputEmitter/NodeEmitter/MapEmitter.php
@@ -16,9 +16,39 @@ final class MapEmitter implements NodeEmitterInterface
     {
         assert($node instanceof MapNode);
 
-        $this->outputEmitter->emitContextPrefix($node->getEnv(), $node->getStartSourceLocation());
+        /*$this->outputEmitter->emitContextPrefix($node->getEnv(), $node->getStartSourceLocation());
         $this->outputEmitter->emitStr('\Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(', $node->getStartSourceLocation());
         $this->outputEmitter->emitArgList($node->getKeyValues(), $node->getStartSourceLocation());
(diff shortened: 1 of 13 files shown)


src/php/Compiler/Emitter/OutputEmitter/NodeEmitter/MapEmitter.php
```diff
@@ -16,10 +16,40 @@ public function emit(AbstractNode $node): void
     {
         assert($node instanceof MapNode);
 
-        $this->outputEmitter->emitContextPrefix($node->getEnv(), $node->getStartSourceLocation());
+        /*$this->outputEmitter->emitContextPrefix($node->getEnv(), $node->getStartSourceLocation());
         $this->outputEmitter->emitStr('\Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(', $node->getStartSourceLocation());
         $this->outputEmitter->emitArgList($node->getKeyValues(), $node->getStartSourceLocation());
         $this->outputEmitter->emitStr(')', $node->getStartSourceLocation());
+        $this->outputEmitter->emitContextSuffix($node->getEnv(), $node->getStartSourceLocation());*/
```
JesusValera, line 23: What about this?

Chemaclass: If this is not used, I would suggest removing it

Chemaclass (approved):

jenshaase:
I removed the lines.

PR #427, opened 2022-01-19
Added support for inline optimization.

jenshaase:
This commits adds the possibility to mark functions for inline optimization. Inline optimization is very useful for math and bit operation. It returns the number of instruction that need to be executed. For example before this PR

```
(+ 1 (* 2 3))
```

would lead two to function calls. First, Phel's *-function is called which itself calls the PHP's *-function. Then the same happens for the +-function. The compiled code looks like this:

```php
(\Phel\Lang\Registry::getInstance()->getDefinition("phel\\core", "+"))(
  1,
  (\Phel\Lang\Registry::getInstance()->getDefinition("phel\\core", "*"))(
    2,
    3));
```

The inline optimization would directly call PHP's function. The compiled PHP code looks exactly like expected:

```php
(1 + (2 * 3))
```

This can now be archived by adding `:inline` and optional `:inline-arity` meta data to a function. `:inline` is a macro function that returns new optimized code that should be analyzed instead of the normal function. Similar to macros. `:inline-arity` is an optional precondition that can be added. It is a one argument function that returns true or false. The value of the argument is amount of arguments that a analyzed call retrieves.

I also added a REPL helper function (`compile-str`) that prints the compiled php code.
```
phel:1> (repl/compile-str "(+ 1 1)")
(1 + 1);
```


src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/InvokeSymbol.php
```diff
@@ -27,6 +28,12 @@ public function analyze(PersistentListInterface $list, NodeEnvironmentInterface
             $env->withContext(NodeEnvironmentInterface::CONTEXT_EXPRESSION)->withDisallowRecurFrame()
         );
 
+        /*var_dump(gettype($f));
+        var_dump($this->isInline($f, count($list) - 1));*/
```
JesusValera, line 32: Missing comments 👀

JesusValera (commented):
Amazing ✨

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/InvokeSymbol.php
jenshaase: Fixed. :sweat:

Chemaclass: Btw instead of `var_dump()` try using `dump()` 😉

src/phel/core.phel
```diff
@@ -477,11 +487,13 @@ Calling the and function without arguments returns true."
 
 (defn <=>
   "Alias for the spaceship PHP operator in ascending order. Returns an int."
+  {:inline (fn [a b] `(php/<=> ,a ,b))}
   [a b]
   (php/<=> a b))
```
Chemaclass, line 492: What about these functions where the "phel fn" is obsolete by the "inline fn"?
If they aren't used, would it be wise to remove them and write a comment like:
```newlisp
(defn <=>
  "Alias for the spaceship PHP operator in ascending order. Returns an int."
  {:inline (fn [a b] `(php/<=> ,a ,b))}
  [a b]
  # fn body removed in favor of the `:inline` function
)
```
I guess this same idea applies for all `:inline` functions without `:inline-arity`?

src/php/Compiler/Analyzer/TypeAnalyzer/SpecialForm/InvokeSymbol.php
```diff
@@ -68,6 +110,17 @@ private function macroExpand(PersistentListInterface $list, GlobalVarNode $macro
...
 
+    /**
+     * @return TypeInterface|string|float|int|bool|null
+     */
+    private function callMacroFn(callable $fn, PersistentListInterface $list)
+    {
+        $arguments = $list->rest()->toArray();
+
+        $result = $fn(...$arguments);
+        return $this->enrichLocation($result, $list);
```
Chemaclass, line 121: This is nice.

tests/php/Integration/Fixtures/Inline/mult.test
```diff
@@ -0,0 +1,6 @@
+--PHEL--
+(*)
+(* 1)
+(* 1 2)
+--PHP--
+(\Phel\Lang\Registry::getInstance()->getDefinition("phel\\core", "*"))();(1);(1 * 2);
```
Chemaclass, line 6: I think it would be interesting to add one more test with 3 arguments for those who have variadic/undefined number of arguments, because by default they are 1 or 2, so it would be helpful to see `(* 1 2 3) => (1 * 2 * 3)` too. The same applies to `+` and `-`, for example.
Maybe it doesn't worth it, and it's ok as it is... but I wanted to share this idea.

tests/php/Integration/Fixtures/Inline/smaller.test
```diff
@@ -0,0 +1,6 @@
+--PHEL--
+(< 1)
+(< 1 2)
+(< 1 2 3)
+--PHP--
+(\Phel\Lang\Registry::getInstance()->getDefinition("phel\\core", "<"))(1);(1 < 2);(\Phel\Lang\Registry::getInstance()->getDefinition("phel\\core", "<"))(1, 2, 3);
```
Chemaclass, line 6: Although these tests are technically correct, the PHP assertion is really hard to understand because everything is in one line. Could we place them in different lines somehow? I know it might be not that easy, so we (I, for example)  could work on a follow up PR to try this idea out.

Chemaclass (approved):
We can merge this PR as it is, because it's indeed awesome! 
But I wanted to write down some thoughts and ideas for potential follow-up tasks.

src/phel/core.phel
jenshaase: The body of the function is still required, if you use the function as parameter in other function. For example

```
(apply <=> 1 [2])
(map inc [1 2 3])
# ...
```

tests/php/Integration/Fixtures/Inline/mult.test
jenshaase: I added these tests.

tests/php/Integration/Fixtures/Inline/smaller.test
jenshaase: Yes that can be improved. I just tried it but it is more work. Basically one have to change `OutputEmitter::emitContextSuffix` to do `emitLine` instead of `emitStr` and then remove all unnecessary `emitLine` calls in the other Emitters.

PR #432, opened 2022-01-24
Add final dot to symbol doc

JesusValera:
### 📚  Description

As we discuss, the purpose of this PR is to standardise the API documentation.
In order to achieve that, I add a final dot to the symbol documentation.

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -112,12 +112,12 @@ elements, returns nil."
 # ------------
 
 (def meta
-  "```phel\n(meta obj)\n```\nGets the meta of the give object"
+  "```phel\n(meta obj)\n```\nGets the meta of the give object."
   (fn [obj]
     (php/-> obj (getMeta))))
 
 (def set-meta!
-  "```phel\n(set-meta! obj)\n```\nSets the meta data to a given object"
+  "```phel\n(set-meta! obj)\n```\nSets the meta data to a given object."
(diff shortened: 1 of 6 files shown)


jenshaase:
See: https://github.com/phel-lang/phel-lang/blob/master/src/phel/core.phel#L189 and https://github.com/phel-lang/phel-lang/blob/master/src/phel/core.phel#L190

Is there a way to have a CI/CD Task to check if the period is available when new code is added or should we do this manually?

JesusValera:
> Is there a way to have a CI/CD Task to check if the period is available when new code is added or should we do this manually?

I think there is no way to do it (at least for now).
We can add some rules for PHP syntax as we were doing with [php_cs_fixer](https://github.com/phel-lang/phel-lang/blob/master/.php-cs-fixer.dist.php#L45), but our Phel code is not PHP, so, this tool doesn't know the syntax to check that.
By the way, I think manually is good enough, next time we should put more attention when creating new symbols, and if we forget some, we can fix it in a new PR.

Chemaclass (approved):
Cool. Just need to add the dot where Jens mentioned on lines: https://github.com/phel-lang/phel-lang/blob/master/src/phel/core.phel#L189 & 190

JesusValera:
Updated!

Chemaclass (approved):

PR #440, opened 2022-02-05
Docker quickstart container

jenshaase:
### 🤔 Background

I want to give new user a very easy way got try out the REPL.

### 💡 Goal

This PR adds a Dockerfile to build a Docker image that starts the REPL. A GitHub actions workflow is added to automatically publish a version to docker hub. 

New user can use this image to try out Phel simply by running
```
docker run -it --rm phellang/repl
```

I already a push a image to Docker Hub. So you can try it out yourself.

### 🔖 Changes

* Added a Dockerfile
* Added a new GitHub Actions workflow.

---
Next: Update the website with this command

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,24 @@
+name: Publish docker images
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  repl:
+    name: Docker quickstart repl
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: docker login
+      env:
+        DOCKER_USER: ${{secrets.DOCKER_USER}}
+        DOCKER_PASSWORD: ${{secrets.DOCKER_PASSWORD}}
+      run: |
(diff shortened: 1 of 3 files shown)


Chemaclass:
Minor tiny detail... but, considering that the org & project is named: "phel-lang", and the website phel-lang.org, and the twitter phel_lang, I see a clear space between "phel" and "lang", what about using the space as standar for the org also for this Docker image? I am thinking about:
Instead of: `docker run -it --rm phellang/repl`
Having this: `docker run -it --rm phel-lang/repl`

I know it might look a bit silly, but these details are way easier to change at the beginning than later. So I just wanted to put this thought on the table. I would prefer (for conistency) the `phel-lang/repl`, but if you prefer the other one, then it would be fine too. What do you think about it, @jenshaase & @JesusValera?

jenshaase:
* Docker doesn't allow a `phel-lang`. It only allows letters and digits :(
* I also couldn't register `phel-lang` on Twitter :(

Chemaclass (approved):

PR #441, opened 2022-02-10
Added two new merge function: merge-with and deep-merge

jenshaase:
### 🤔 Background

Merging maps or nested data structures is a common problem in programming. This PR will add new functions that may be very useful.

### 💡 Goal

The first function is call `merge-with`. `merge-with` takes a two-arity function `f` and a unlimited number of hash maps. It will merge these hash maps as in `merge` but in case two maps have the same key it will call `f` with the left and the right value. Example:

```
(merge-with + {:a 1 :b 2} {:b 3 :c 4}) # Evaluates to {:a 1 :b 5 :c 4}
```

The second function is called `deep-merge`. This function will recursively merge nested data structures. Hash maps will be merged as in `merge`, vectors will be concatenated, sets will be united. Example:

```
(deep-merge {:a {:b {:c [:d]}}} {:a {:b {:c [:e] :f :g}}} # Evaluates to {:a {:b {:c [:d :e] :f :g}}}
```

### 🔖 Changes

* Added: `merge-with` function
* Added: `deep-merge` function

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1395,6 +1395,43 @@ collection in map"))
    (complement indexed?)
    (rest (tree-seq indexed? identity xs))))
 
+(defn- merge-with-2 [f left right]
+  (for [[k v] :pairs right
+        :reduce [acc left]]
+    (if (contains? acc k)
+      (put acc k (f (get acc k) v))
+      (put acc k v))))
+
+(defn merge-with
+  "Merges multiple maps into one new map. If a key appears in more than one
+   collection, the result of `(f current-val next-val)` is used."
+  [f & hash-maps]
+  (case (count hash-maps)
+    0 {}
+    1 (first hash-maps)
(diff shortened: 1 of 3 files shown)


src/phel/core.phel
```diff
@@ -1395,6 +1395,43 @@ collection in map"))
...
+(declare deep-merge)
+
+(defn- deep-merge-2 [left right]
+  (cond
+    (nil? left) right
+    (nil? right) left
+    (and (hash-map? left) (hash-map? right)) (merge-with deep-merge left right)
+    (and (set? left) (set? right)) (union left right)
+    (and (vector? left) (vector? right)) (concat left right)
+    right))
```
Chemaclass, line 1424: I see, in case there are 2 different types, it will take the one on the right:
```lisp
(deep-merge {:a :b} [:c :d]) #=> [:c :d]
```
I just wanted to point it out. I don't see any other alternative than crashing 🤣

Chemaclass (approved):

PR #442, opened 2022-02-12
NamedInterface

jenshaase:
### 🤔 Background

`\Phel\Lang\Symbol` and `\Phel\Lang\Keyword` have both the methods `getName`, `getNamespace` and `getFullName`.

### 💡 Goal

Put the three methods into a unified Interface and let Symbol and Keyword implement these interface. Additionally, add functions in Phel's core library to call these methods. Use the new functions in the rest of the code.

The function `name` and `full-name` also support strings. This is quite useful to convert Hash Map to PHP arrays. Example:
```
(def data {:foo "value" "bar" "value2"})

(->> data
       (map-indexed (fn [k v] [(name k) v]))
       (flatten)
       (apply php-associative-array))
```

### 🔖 Changes

* Added: `NamedInterface` interface for Symbol and Keyword
* Added: `name` function
* Added: `namespace` function
* Added: `full-name` function
* Use the added function in the rest of the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ All notable changes to this project will be documented in this file.
 
 * Added: `merge-with` function
 * Added: `deep-merge` function
+* Added: `NamedInterface` interface for Symbol and Keyword
+* Added: `name` function
+* Added: `namespace` function
+* Added: `full-name` function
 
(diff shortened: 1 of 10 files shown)


Chemaclass (approved):
Nice 👌

PR #444, opened 2022-02-22
Improve phel\http

jenshaase:
### 💡 Goal

Improve `phel\http`

### 🔖 Changes

* Added: `http/uri-from-string` function
* Added: Uri structs implements `Stringable` interface
* Added: `http/response-from-map` function
* Deprecated: `http/create-response-from-map` in favor of `http/response-from-map`
* Added: `http/response-from-string` function
* Deprecated: `http/create-response-from-string` in favor of `http/response-from-string`
* Added: `attributes` field to `request` struct. Allows developers to enrich the request with custom data
* Added: `http/request-from-map` function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,14 @@ All notable changes to this project will be documented in this file.
 * Added: `name` function
 * Added: `namespace` function
 * Added: `full-name` function
+* Added: `http/uri-from-string` function
+* Added: Uri structs implements `Stringable` interface
+* Added: `http/response-from-map` function
+* Deprecated: `http/create-response-from-map` in favor of `http/response-from-map`
+* Added: `http/response-from-string` function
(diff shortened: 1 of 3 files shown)


Chemaclass (approved):
So many good stuff! 👍

mabasic:
This will probably be a breaking change for the http library?

I can't remember now if it was fixed but there was some confusion while building my website around accessing post payload or something like that where the data was not in the expected place.

I plan to do some work on my website and phel (pr for json library and documentation for developers) this week so I will review my code to remember what it was exactly.

JesusValera (approved):

jenshaase:
@mabasic There should be no breaking changes. I renamed two method, but also added aliases form the old names to the new names. The old names are marked as deprecated and will be removed in future releases.

PR #451, opened 2022-03-02
Emit PHP array literals

jenshaase:
With this commit it is possible to emit PHP array literals. This is useful if a macro wants to return a PHP array.

### 🤔 Background

Macros can currently only return Phel data structures. For some use cases it is useful to also allow PHP arrays as return type.

### 💡 Goal

This PR makes it possible to emit PHP array literals. Example:

```clojure
(defn build-some-php-array []
  (php/array 1 2))

(defmacro macro-returns-php-array []
  (build-some-php-array))
```

### DIFF
--- a/src/php/Compiler/Emitter/OutputEmitter/LiteralEmitter.php
+++ b/src/php/Compiler/Emitter/OutputEmitter/LiteralEmitter.php
@@ -28,7 +28,7 @@ final class LiteralEmitter
     }
 
     /**
-     * @param TypeInterface|string|float|int|bool|null $x The value
+     * @param TypeInterface|string|float|int|bool|null|array $x The value
      */
     public function emitLiteral($x): void
     {
@@ -52,6 +52,8 @@ final class LiteralEmitter
             $this->emitVector($x);
         } elseif ($x instanceof PersistentListInterface) {
             $this->emitList($x);
+        } elseif (is_array($x)) {
+            $this->emitArray($x);
         } else {
             $typeName = gettype($x);
             if ($typeName === 'object') {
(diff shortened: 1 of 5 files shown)


src/php/Compiler/Emitter/OutputEmitter/LiteralEmitter.php
```diff
@@ -186,4 +188,22 @@ private function emitVector(PersistentVector $x): void
...
+        $this->outputEmitter->emitStr('[');
+        $count = count($x);
+        $index = 0;
+        foreach ($x as $key => $value) {
+            $this->outputEmitter->emitLiteral($key);
+            $this->outputEmitter->emitStr(' => ');
+            $this->outputEmitter->emitLiteral($value);
+            if ($index < $count - 1) {
+                $this->outputEmitter->emitStr(', ');
+            }
```
Chemaclass, line 203: I wonder if this condition is needed? I mean, you can have a comma after the last element in an array in PHP. Or is this for another reason?

Chemaclass: Well, I don't think this is a big deal anyway. This is pure cosmetics to make the output PHP code nicer. And considering we have a compiler, there is no downtime performance or anything similar. This is totally fine.
Screenshot 2022-03-03 at 08 08 23

Chemaclass (approved):

Chemaclass:
Maybe should we add a line in the `CHANGELOG.md` with this addition, @jenshaase?

jenshaase:
I updated the Changelog file.

PR #453, opened 2022-03-18
Update gacela bootstrap in phel static helper

Chemaclass:
### 🤔 Background

When updating to gacela 0.14 we forgot to update the `Phel::run()` method, on which we are actually using `Gacela::bootstrap()`. This happened mainly because we didn't have any tests covering/using this code.

### 🔖 Changes

- Update the `Phel::run()` internal method to use the new gacela version code.
- Add a Bench test which is executing the `Phel::run()` so we wont have this problem in the future anymore.

### DIFF
--- a/src/php/Phel.php
+++ b/src/php/Phel.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Phel;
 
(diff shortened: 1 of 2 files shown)


JesusValera (approved):
🙉

PR #454, opened 2022-03-18
Printer module using php80

Chemaclass:
### 🔖 Changes

- Update Printer module to use some new PHP 8.0 features
  - use mixed keyword
  - use `array_is_list` function
- Fix a (really hidden) bug  inside the `createObjectTypePrinter()`
  - The condition checking the `AbstractPersistentStruct` was never reached

### DIFF
--- a/src/php/Printer/TypePrinter/ArrayPrinter.php
+++ b/src/php/Printer/TypePrinter/ArrayPrinter.php
@@ -6,6 +6,8 @@ namespace Phel\Printer\TypePrinter;
 
 use Phel\Printer\PrinterInterface;
 
+use function array_is_list;
+
 /**
  * @implements TypePrinterInterface<array>
  */
@@ -23,18 +25,13 @@ final class ArrayPrinter implements TypePrinterInterface
     /**
      * @param array $form
      */
-    public function print($form): string
+    public function print(mixed $form): string
     {
-        $arr = $this->isList($form)
+        $arr = array_is_list($form)
             ? $this->formatValuesFromList($form)
             : $this->formatKeyValuesFromDict($form);
 
         return sprintf('<PHP-Array [%s]>', $this->color(implode(', ', $arr)));
-    }
-
-    private function isList(array $form): bool
-    {
-        return array_keys($form) === range(0, count($form) - 1);
     }
 
     private function formatValuesFromList(array $form): array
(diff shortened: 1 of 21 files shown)


src/php/Printer/TypePrinter/ArrayPrinter.php
```diff
@@ -23,20 +25,15 @@ public function __construct(PrinterInterface $printer, bool $withColor = false)
     /**
      * @param array $form
      */
-    public function print($form): string
+    public function print(mixed $form): string
     {
-        $arr = $this->isList($form)
+        $arr = array_is_list($form)
```
jenshaase, line 30: This function is only available in PHP 8.1 not PHP 8.0. See: https://www.php.net/manual/de/function.array-is-list.php

There is probably no tests for this?

Chemaclass: Wow, good catch! Actually we do have tests for this, see `ArrayPrinterTest`. After some investigation, I found why this is working: because we have this in vendor
```php
# source: vendor/symfony/polyfill-php81/bootstrap.php
if (!function_exists('array_is_list')) {
    function array_is_list(array $array): bool { return p\Php81::array_is_list($array); }
}
```
And `Php81::array_is_list` has the same behaviour as the native function, but written in PHP:
```php
<?php
# source: vendor/symfony/polyfill-php81/Php81.php
namespace Symfony\Polyfill\Php81;

final class Php81
{
    public static function array_is_list(array $array): bool
    {
        if ([] === $array || $array === array_values($array)) {
            return true;
        }

        $nextKey = -1;

        foreach ($array as $k => $v) {
            if ($k !== ++$nextKey) {
                return false;
            }
        }

        return true;
    }
}
```

And why do we have that code in the vendor? because it's requires from `friendsofphp/php-cs-fixer  v3.7.0`:
```bash
phel-lang/ (refactor/printer-module-using-php80) $ composer why symfony/polyfill-php81
friendsofphp/php-cs-fixer  v3.7.0  requires  symfony/polyfill-php81 (^1.25)
```
---

I will simply rollback in order to have our custom `array_is_list` as we had before, in order to keep as less external dependencies as necessary.

jenshaase:
Looks good now. I merged it.

PR #455, opened 2022-03-19
Add more php cs rules and apply them

Chemaclass:
### 💡 Goal

The goal is to have more rules, so the code can be uniform, and optimized out of the box for every possible corner.

### 🔖 Changes

I went throw all PHP CS rules that could potentially bring some benefits to our codebase, so I added them. My favorite is the one that adds the `use function xxx` for native PHP functions that are optimized when declaring its usage. Also, the sorting of the methods, and the way conditions are written to avoid possible mistakes. 

I am fine with all of them so far, but let me know if you dislike something, in particular, to discuss it and change it if necessary.

For more details, you can see: https://mlocati.github.io/php-cs-fixer-configurator/#version:3.8


jenshaase:
Can you post a list of rules that have been added? I can't really see them in the diff.

Chemaclass:
I intended properly the `.php-cs-fixer.dist.php` (see [commit](https://github.com/phel-lang/phel-lang/commit/0b460c1eefb217f3f3aaccfcb89554b1e12da004)), and I rebased this branch from master. Now you can see the diff properly, @jenshaase.

JesusValera (commented):
I am not 100% sure with the Yoda style, it's a nice to have but should not be "mandatory", what do you think guys?

.php-cs-fixer.dist.php
```diff
@@ -17,21 +17,64 @@
...
+                'package',
+                'subpackage',
+                'version',
+            ],
+        ],
+        'global_namespace_import' => [
+            'import_functions' => true,
+        ],
+        'include' => true,
+        'increment_style' => true,
```
jenshaase, line 46: Shouldn't the value be "pre" or "post"? 
https://mlocati.github.io/php-cs-fixer-configurator/#version:3.8|fixer:increment_style

.php-cs-fixer.dist.php
```diff
@@ -17,21 +17,64 @@
...
+        'native_function_type_declaration_casing' => true,
+        'new_with_braces' => true,
+        'no_blank_lines_after_class_opening' => true,
+        'no_empty_comment' => true,
         'no_empty_phpdoc' => true,
         'no_empty_statement' => true,
+        'no_homoglyph_names' => true,
+        'no_leading_import_slash' => true,
         'no_leading_namespace_whitespace' => true,
+        'no_mixed_echo_print' => true,
```
jenshaase, line 60: Shouldn't the value be either "print" or "echo"?

https://mlocati.github.io/php-cs-fixer-configurator/#version:3.8|fixer:no_mixed_echo_print

jenshaase (commented):
I also would remove the yoda style rule.

.php-cs-fixer.dist.php
Chemaclass: By default is 'pre', but I can change it for clarity to
```php
'increment_style' => ['style' => 'pre'],
```

.php-cs-fixer.dist.php
```diff
@@ -119,5 +119,9 @@
         'types_spaces' => true,
         'visibility_required' => true,
         'void_return' => true,
-        'yoda_style' => false,
+        'yoda_style' => [
+            'equal' => true,
+            'identical' => false,
+            'less_and_greater' => null,
+        ],
```
Chemaclass, line 126: See Example2: https://mlocati.github.io/php-cs-fixer-configurator/#version:3.8|fixer:yoda_style
Screenshot 2022-03-26 at 13 03 40

It's only enabled for equality. I didn't know we could have this granular option, which is my favourite for now. What do you say, @jenshaase?

jenshaase:
I will merge this now. We can see how it works and then change thinks if we need to.

PR #459, opened 2022-04-03
Added read-string, eval and compile function to core library

jenshaase:
### 🤔 Background

Add some more function to core library

### 💡 Goal / Changes

The following functions have been added:

* `read-string`: Takes a string and returns the parsed expression
* `eval`: Takes a Phel form and evaluates it
* `compile`: Takes a Phel form and return the compiled PHP code.


Chemaclass (approved):
Cool 👍🏼

PR #460, opened 2022-04-03
Create a ConsoleBootstrap

Chemaclass:
### 🤔 Background

Currently all console commands are created inside the `phel` file, the entry point of phel. I would delegate the responsibility of the command creation in another place like a Console module.

### 🔖 Changes

- Created a `ConsoleBootstrap` which extends Application Symfony Command.
- Define all phel commands inside the `ConsoleDependencyProvider`.


JesusValera (approved):

jenshaase:
Looks very nice. Thanks!

PR #461, opened 2022-04-06
Remove constructor from CompileCommand

Chemaclass:
### 🤔 Background

Currently, the creation of the commands is done in the Factory of the module, on which we inject its dependencies. But then, the command it's not self-instanciable" and it needs the Factory or the Facade (which later uses its Factory) to create that command... it's a bit to ugly, because the command start depending on a lot of things, like other external facades or internal dependencies. 

But, this is odd, a Command (like a Controller) is part of the infrastructure code that should have access to the Facade of its module, and the Facade knows how to deal with other Facades and perform the logic that it's required.

This PR is a refactoring example of making the `CompileCommand` depending only on the `CompileFacade` using `getFacade()` method.

### 💡 Goal

Simplify the creation of the Commands in `src/php/Console/ConsoleDependencyProvider.php`.
In this PR I just focused on one: `CompileCommand`.

### 🔖 Changes

- Remove the constructor from `CompileCommand` and use `getFacade()` instead.
  - This is possible thanks to the new `FacadeResolverAwareTrait`.
- The command now can be created in the `ConsoleDependencyProvider` using `new CompileCommand()`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require": {
         "ext-json": "*",
         "php": ">=8.0",
-        "gacela-project/gacela": "^0.15",
+        "gacela-project/gacela": "dev-master",
         "symfony/console": "^5.4",
         "phpunit/php-timer": "^5.0"
     },
@@ -68,5 +68,6 @@
         "allow-plugins": {
             "composer/package-versions-deprecated": true
         }
-    }
+    },
+    "minimum-stability": "dev"
 }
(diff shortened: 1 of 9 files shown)


jenshaase:
Can you resolve the Merge Conflict?

Chemaclass:
@jenshaase done ✅

JesusValera (approved):

PR #462, opened 2022-04-08
Install PHPStan

Chemaclass:
### 🤔 Background

Similar like Psalm, PHPStan is another static analysis tool for PHP which catches a lot of other things that Psalm doesn't. Currently it's not possible to install PHPStan at any higher level but the minimum (level 1; the highest level in PHPStan is 9|max).

### 💡 Goal

Use PHPStan to spot potential bugs in the project, and so improve the quality overall. Starting with the min level and upgrading it to higher levels step by step.

### 🔖 Changes

- Install PHPStan with level 1 - the lowest level (for now).
- In the `phpstan.neon` you can see that I'm excluding a lot of paths. The reason: `Child process error (exit code 255): PHP Fatal error:  Allowed memory size of 541065216 bytes exhausted (tried to allocate 262144 bytes) in [...]`, so I thought a quick way to get rid of those "memory size" exceptions could be just ignoring those clases, and we can still get the benefits of PHPStan for the rest of the project.

Screenshot 2022-04-08 at 23 45 40

###  💡 Next steps

Change the [parameters.level from 1 to 2](https://github.com/phel-lang/phel-lang/compare/feature/install-phpstan?expand=1#diff-0361f0c81f363476ddc6f44ab36fcbe66ee685d5f4c2a46b054924591544b766R2) and run `composer phpstan`, you will get some feedback with ideas for potential bugs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,8 +70,11 @@ jobs:
       - name: Install dependencies
         run: composer install --no-interaction --no-ansi --no-progress
 
-      - name: Run vimeo/psalm on internal code
+      - name: Run Psalm on internal code
         run: ./vendor/bin/psalm --config=psalm.xml --no-progress --shepherd --show-info=false --stats --php-version=${{ matrix.php }}
+
+      - name: Run PHPStan on internal code
+        run: ./vendor/bin/phpstan --memory-limit=516M --debug -vvv
 
   compiler-tests:
     name: Compiler tests on PHP ${{ matrix.php }}
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,8 @@
         "symfony/var-dumper": "^5.4",
         "friendsofphp/php-cs-fixer": "^3.8",
         "phpmetrics/phpmetrics": "^2.8",
-        "phpbench/phpbench": "^1.2"
+        "phpbench/phpbench": "^1.2",
+        "phpstan/phpstan": "^1.5"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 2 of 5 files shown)


JesusValera (approved):

jenshaase:
Could you give me an overview of:

a) what causes the memory limit errors
b) what needs to be done to improve the levels?
c) can we use PHPStan and Psalm together without problems?

Chemaclass:
Absolutely, @jenshaase!  

### a) what causes the memory limit errors

I don't know yet. I assume it might be something related to the "@template" and/or some related recursivity problem...
You can reproduce it easily commenting out any line from the "ignorePaths" lines, and running
> `./vendor/bin/phpstan --memory-limit=516M --debug -vvv`

Screenshot 2022-04-09 at 20 15 26

### b) what needs to be done to improve the levels?
Easy: change the parameters.level to a higher number, and you'll directly see the errors that PHPStan reports.
Screenshot 2022-04-09 at 20 19 01
I would say, we can follow a similar approach as we did with Psalm last year. We didn't started with the higher/stricter level, but we were upgrading step by step. They have a different order for the levels. But, overall, they work in a really similar way.
- In Psalm "Level 1 is the most strict, level 8 is the most lenient.". See: https://psalm.dev/docs/running_psalm/error_levels/
- In PHPStan "Level 0 is the loosest and 9 is the strictest". See: https://phpstan.org/user-guide/rule-levels

### c) can we use PHPStan and Psalm together without problems?
Yes, we can use both totally fine. They are 100% compatible and independent. Gacela uses both tools, and I've never faced any problem by using both. In fact, there are some things that PHPStan catches that Psalm doesn't, and vice versa. So, I think they both gives you a really good feedback about hidden spots for potential quality improvements or possible hidden bugs.

jenshaase:
Thanks for the explanation. I will merge it now.

PR #465, opened 2022-04-20
Remove constructor for TestCommand using RunFacade instead

Chemaclass:
### 🤔 Background

Similar as it was done in https://github.com/phel-lang/phel-lang/pull/461 

### 💡 Goal

The goal is to allow the creation of the `TestCommand` directly, and for this we need to get rid of its dependencies.

### 🔖 Changes

Refactor the `TestCommand` using internally its RunFacade. This way, its communication will be via its facade; 
- accesible via `getFacade()`, a new feature of the latest version from Gacela with `use FacadeResolverAwareTrait;`
- see: https://gacela-project.com/docs/facade/#use-the-facade-from-your-infrastructure-layer


JesusValera (approved):

jenshaase:
Cool. Merged it.

PR #466, opened 2022-04-23
Refactor FormatCommand to use the Facade

Chemaclass:
### 🤔 Background

Similar as it was done in https://github.com/phel-lang/phel-lang/pull/461 & https://github.com/phel-lang/phel-lang/pull/465

### 💡 Goal

The goal is to allow the creation of the `FormatCommand` directly, and for this we need to get rid of its dependencies.

### 🔖 Changes

Refactor the `FormatCommand` using internally its `FormatterFacade`. This way, its communication will be via its facade;

- accesible via getFacade(), a new feature of the latest version from Gacela with use FacadeResolverAwareTrait;
- see: https://gacela-project.com/docs/facade/#use-the-facade-from-your-infrastructure-layer


jenshaase:
Looks good! Thank you!

Issue #467, opened 2022-04-27
Failed to run a REPL


mfaridw:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.6           |
| PHP version         | 8.1.2           |

#### Summary

I failed to run REPL from PowerShell. I have also tried to create a new project from the two methods provided but the result is still the same.

#### Current behavior

reproduce error messages :
```
stripos(): Argument #1 ($haystack) must be of type string, null given
#0 D:\farid\phel\hello-world\vendor\phel-lang\phel-lang\src\php\Run\Domain\Repl\ReplCommandSystemIo.php(71): stripos(NULL, 'editline')     
#1 D:\farid\phel\hello-world\vendor\phel-lang\phel-lang\src\php\Run\Command\ReplCommand.php(127): Phel\Run\Domain\Repl\ReplCommandSystemIo->isBracketedPasteSupported()
#2 D:\farid\phel\hello-world\vendor\phel-lang\phel-lang\src\php\Run\Command\ReplCommand.php(110): Phel\Run\Command\ReplCommand->addLineFromPromptToBuffer()
#3 D:\farid\phel\hello-world\vendor\phel-lang\phel-lang\src\php\Run\Command\ReplCommand.php(101): Phel\Run\Command\ReplCommand->loopReadLineAndAnalyze()
#4 D:\farid\phel\hello-world\vendor\symfony\console\Command\Command.php(298): Phel\Run\Command\ReplCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#5 D:\farid\phel\hello-world\vendor\symfony\console\Application.php(1015): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#6 D:\farid\phel\hello-world\vendor\symfony\console\Application.php(299): Symfony\Component\Console\Application->doRunCommand(Object(Phel\Run\Command\ReplCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))        
#7 D:\farid\phel\hello-world\vendor\symfony\console\Application.php(171): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
[...]

PR #468, opened 2022-04-27
Fix ReplCommandSystemIo when internal library_version is null

Chemaclass:
### 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/467

There seems to be a bug when `readline_info('library_version')` return `null`. And this is an error because `stripos()` expect a non-nullable string as first argument.

### 🔖 Changes

- Use an empty string when `readline_info('library_version')` returns null

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 * Added: `read-string` function
 * Added: `eval` function
 * Added: `compile` function
+* Bugfix: #467 Failed to run a REPL on Windows
 
 ## 0.6.0 (2022-02-02)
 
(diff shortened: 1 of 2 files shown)


jenshaase:
Can you update the Changelog.md?

mfaridw (commented):
Thanks, it's work

mfaridw (commented):
ok

mfaridw (commented):
nice

mfaridw (commented):
done.

mfaridw:
I'm still wondering, why every time I create a project, composer always gives me the old ReplCommandSystemIo.php?

jenshaase:
Probably, because composer uses the latest released version (0.6.0) and not the recent code from the master branch.

We will release a new version next week. Then your problem should be solved.

PR #470, opened 2022-04-29
Split domain and infrastructure code into folders inside each module

Chemaclass:
### 🤔 Background

Right now it's not easy to differentiate the layer of the logic/classes inside a module.

### 💡 Goal

Differentiate the code under domain or infrastructure layer.

### 🔖 Changes

Split each module into Domain (internal logic to the module) and Infrastructure (accesible to the "outside" [phel code, or commands, for example])


jenshaase:
Cool!

PR #489, opened 2022-06-11
Adds json library to phel.

mabasic:
### 🤔 Background

Moving my library mabasic/phel-json to Phel core as discussed previously many times :)
https://github.com/mabasic/phel-json/issues/2

### 💡 Goal

Move mabasic/phel-json library to Phel to have it in the language by default.

@Chemaclass @jenshaase @JesusValera I finally got the time to handle this :) Should we move the documentation from my package to Phel website also?


jenshaase:
@mabasic Regarding documentation: I think it would be enough if we add doc comments to the functions.

One other improvement: I would suggest to make all function private except `encode` and `decode`.

Let me know if you are short in time, then we can take over the PR and improve it before we merge it.

mabasic:
> One other improvement: I would suggest to make all function private except `encode` and `decode`.

@jenshaase It was like that before (private), but then I had a use case where I needed to use the private functions, so I set everything to be public:

https://github.com/mabasic/mariobasic-n7/blob/main/src/helpers.phel


I'll add the comments tonight or tomorrow. Will it be included in the docs api: https://phel-lang.org/documentation/api/ ?

jenshaase:
Can you tell me more about the use cases where you need the "private" function? This is probably easier than going through your code.

jenshaase:
Proposal: Let's make the private function private and merge this PR. We can make them public at a later point if necessary. Keeping them public and then making them private will always break backwards compatibility.

mabasic:
I needed it here:
https://github.com/mabasic/mariobasic-n7/blob/main/src/pages/start.phel#L172

This is the library that needed it:
https://github.com/mabasic/mariobasic-n7/blob/main/src/database.phel#L6

From what I can remember/gather I needed a way to convert this:
https://github.com/mabasic/mariobasic-n7/blob/main/src/pages/start.phel#L152

(helper here)
https://github.com/mabasic/mariobasic-n7/blob/main/src/helpers.phel#L12

to an associative php array which can then be passed to a library which would save it to the database.

## Use case

Create a PHP Array from a sequential data structure.

I have seen that Phel has `to-php-array` (https://phel-lang.org/documentation/api/#to-php-array) but it cannot convert things like keywords and symbols. That is why I have created `json/encode-value` function while creating the `json/encode` function. I found the method `json/encode-value` very useful in this case.

Maybe we can change the `to-php-array` to match the `json/encode-value (private function)`. In this way I would no longer need the `encode-value` and it could be erased?

## REPL example

```
# Taken from mabasic/json library.
(defn- valid-key? [v]
  (or (int? v) (float? v) (symbol? v) (keyword? v) (string? v)))

# Taken from mabasic/json library because it is private.
(defn encode-value [x]
  (cond
    (php/is_iterable x)
    (let [arr (php/array)]
      (foreach [k v x]
        (when-not (valid-key? k)
          (throw (php/new \Exception "Key can only be an integer, float, symbol, keyword or a string.")))
        (php/aset arr (encode-value k) (encode-value v)))
      arr)
    (symbol? x) (str (php/-> x (getName)))
    (keyword? x) (str (php/-> x (getName)))
    (float? x) (str x)
    true x))

(def name {:name "naziv"})

# Fail
(to-php-array name)
# Phel\Compiler\Domain\Evaluator\Exceptions\CompiledCodeIsMalformedException: Keys must be of type int|string during array unpacking

# Success
(encode-value name)
# <PHP-Array [name:naziv]>
```

src/phel/json.phel
```diff
@@ -0,0 +1,48 @@
+(ns phel\json
+  (:use \Exception))
+
+(defn valid-key? [v]
```
Chemaclass, line 4: Can we make this private?
```suggestion
(defn- valid-key? [v]
```

src/phel/json.phel
```diff
@@ -0,0 +1,48 @@
+(ns phel\json
+  (:use \Exception))
```
Chemaclass, line 2: What about using `\RuntimeException` to provide some more meaning, instead of `\Exception`?
In such a case, you need to update the line 14 (and others) to ` (throw (php/new RuntimeException ...`

mabasic: I have no problem with this.

src/phel/json.phel
mabasic: See my comment on this: https://github.com/phel-lang/phel-lang/pull/489#issuecomment-1259005195

Chemaclass: I see. This `valid-key?` is used only in the `encode-value`, I don't see why we want to keep it public? Or is there any other reason?

Chemaclass:
I see that you're using `helpers/encode-value` here: https://github.com/mabasic/mariobasic-n7/blob/main/src/pages/start.phel#L172 but, wouldn't work if you simply use `json/encode`?

I still believe that what @jenshaase said would be sufficient, which is to make all functions private except `encode` and `decode` as they are the public API for the json library.

```clojure
(defn- valid-key? [v]
  (or (int? v) (float? v) (symbol? v) (keyword? v) (string? v)))

(defn- encode-value [x]
  (cond
    ...
    (symbol? x) (str (php/-> x (getName)))
    (float? x) (str x)
    true x))

(defn encode [value & [{:flags flags :depth depth}]]
  (let [flags (or flags 0) depth (or depth 512)]
    (when-not ...)
    (php/json_encode (encode-value value) flags depth)))

(def name {:name "Mario"})

# Success
(encode-value name) # <PHP-Array [name:Mario]> ; <- why would we want to use this outside the json library?
(encode name)       # {"name":"Mario"}.        ; <- this is what we expect from this library
```

Shouldn't be the responsibility of the json library to decode and encode jsons? Therefore `encode` & `decode` would be enough as public functions.

mabasic:
@Chemaclass If you would need to convert a phel datastructure (with keywords, symbols, ...) to a "json compatible" php-array you would need to:
- either use `json/encode-value` or
- create/copy the exact same method (`json/encode-value`) yourself

`json/valid-key` can be used outside of this library (public) for purposes of detecting if the given variable is a valid json key which `json/encode-value` supports and depending on that one could convert/customize the value how he wants it.

`json/decode-value` converts from php json `{ "something": "value" }` to `{ :something "value" }`. Which I have found that makes sense for Phel.

I have proposed [here](https://github.com/phel-lang/phel-lang/pull/489#issuecomment-1259005195) that we replace `to-php-array` with this (json/encode-value) method. Maybe we can also add a `from-php-array` method (json/decode-value). This way `json/encode-value` and `json/decode-value` are no longer needed and people can use them when they need them outside the `json` library. I still see a use case for `json/valid-key` and would like it to be public. There is no need for people to read the source code to find what are valid keys and then implement the already existing method themselves.

Chemaclass:
@jenshaase and I were thinking about including this PR into the new release `Phel 0.8`, but we want to apply this suggested change first:

> Proposal: Let's make the private function private and merge this PR. We can make them public at a later point if necessary. Keeping them public and then making them private will always break backwards compatibility. ([link](https://github.com/phel-lang/phel-lang/pull/489#issuecomment-1258097310))

Can you make it, please, @mabasic? We will iterate over this in follow up ideas/PRs.

Chemaclass:
Because there is no news on this PR since long time, I will merge it and do the adaptations in master directly, so we'll be able to create the next release.

mabasic:
Sorry for not responding earlier, was busy with other things. Great that this is merged, but I still stand by my opinion that the functions in question should be public or we should make the changes mentioned above.

PR #490, opened 2022-06-18
Use PHP 8.0 syntax

JesusValera:
### Constructor promotion 🚢 

The code has been changed from this
```php
final class CompiledFile
{
    private string $sourceFile;
    private string $targetFile;
    private string $namespace;

    public function __construct(string $sourceFile, string $targetFile, string $namespace)
    {
        $this->sourceFile = $sourceFile;
        $this->targetFile = $targetFile;
        $this->namespace = $namespace;
    }

    // ...
```
to
```php
final class CompiledFile
{
    public function __construct(
        private string $sourceFile,
        private string $targetFile,
        private string $namespace,
    ) {
    }

    // ...
```

Classes with a single dependency, like `Box` or `WithAnalyzerTrait` have been updated to a single-line constructor as follows:
```php
trait WithAnalyzerTrait
{
    public function __construct(private AnalyzerInterface $analyzer)
    {
    }
}
```

However, some classes have not been promoted because could become hard to read/understand, like `RangeIterator`.
```php
class RangeIterator implements Iterator
{
    private PersistentVector $vector;
    private int $start;
    private int $end;
    private int $currentIndex;
    private int $base;
    private ?array $currentArray = null;

    public function __construct(PersistentVector $vector, int $start, int $end)
    {
        $this->vector = $vector;
        $this->start = $start;
        $this->end = $end;
        $this->currentIndex = $start;
        $this->base = $this->currentIndex - ($this->currentIndex % 32);

        if ($this->start < count($this->vector)) {
            $this->currentArray = $this->vector->getArrayForIndex($this->currentIndex);
        }
    }

    // ...
```

### Match operator ⚔️ 

Replace when possible the `switch` statement with `match`.

There are so many good improvements in various classes like `NodeEmitterFactory`, `ListNode` or `AnalyzePersistentList`.

```php
    public function getCodePostfix(): ?string
    {
        switch ($this->tokenType) {
            case Token::T_OPEN_PARENTHESIS:
            case Token::T_FN:
                return ')';
            case Token::T_OPEN_BRACKET:
                return ']';
            case Token::T_OPEN_BRACE:
                return '}';
            default:
                throw new RuntimeException('Cannot find code ...');
        }
    }

    # before and after 👌🏼

    public function getCodePostfix(): ?string
    {
        return match ($this->tokenType) {
            Token::T_OPEN_PARENTHESIS, Token::T_FN => ')',
            Token::T_OPEN_BRACKET => ']',
            Token::T_OPEN_BRACE => '}',
            default => throw new RuntimeException('Cannot find code ...'),
        };
    }
```

### Union types ➕ 

Instead of having
```php

    /**
     * @param TypeInterface|string|float|int|bool|null $x
     *
     * @throws AnalyzerException
     */
    public function analyze($x, NodeEnvironmentInterface $env): AbstractNode;
```
Now is possible to have the following code:
```php
    /**
     * @throws AnalyzerException
     */
    public function analyze(
        TypeInterface|string|float|int|bool|null $x,
        NodeEnvironmentInterface $env,
    ): AbstractNode;
```
So, from now on, the code is much clear, compact and less error-prone.

#### Incorrect PhpDocs 📝 

In some places is not possible to put the PhpDoc types or an error will be thrown, eg: `QuasiquoteTransformer` or the construct method from `ArrayNode`.

```diff
+    /**
+     * @param TypeInterface|string|float|int|bool|null $form The form to quasiqoute
+     *
+     * @return TypeInterface|string|float|int|bool|null
+     */
+    public function transform($form);
-    public function transform(TypeInterface|string|float|int|bool|null $form): TypeInterface|string|float|int|bool|null;
```
#### Generics problem 🌟 

It is not possible to define the type of some objects, like `/** @var T */`.
In those scenarios, I just keep the comment. I think it does not make sense to define them as `mixed`.

```php
final class Variable extends AbstractType
{
    /** @var T */
    private $value;

    /**
     * @param T $value
     */
    public function __construct(?PersistentMapInterface $meta, $value)
    {
        $this->meta = $meta;
        $this->value = $value;
    }
```

### Additionally 📗

This is not something related to PHP 8 itself, it was a small change I thought could be interesting.
Basically was about replacing all documentations as `Object[]` to `list<Object>` or `array<X, Object>`.
The option with braces is not recommended because the static analysis tools allow the array as a list and map at the same time.

I also drop `PhelArrayNode` and `SourceLocationTrait`. These classes were not used in the project anymore.

Finally, minor changes like removing unused variables or fixing some typos.


Chemaclass (approved):
Awesome cleanup/upgrade 👏

jenshaase:
Did you do this manually or did you use [Rector](https://getrector.org/) for this? I once tried to use Rector on this code base but I was not able to get it running.

Chemaclass:
I am pretty sure @JesusValera did all of this manually... but he can tell you better 😝

JesusValera:
> Did you do this manually or did you use [Rector](https://getrector.org/) for this? I once tried to use Rector on this code base but I was not able to get it running.

As far as I know, you cannot do this with Rector, but I am not sure.
I did it manually in order to have perfect control over the changes I had to make 😄

jenshaase (approved):

jenshaase:
Looks good. I will merge it.

PR #492, opened 2022-06-20
Support strings in `empty?`

Chemaclass:
### 🤔 Background

Currently, it's not possible to use the `count` function with strings. 
This is also affecting the `empty?` function, which uses the `count`, you get an error: `object is not countable`.

### 💡 Goal

Allow **checking emptiness** in strings.

### 🔖 Changes

- Allow `(empty? "")` and `(empty? "any string")`
- Display a more descriptive error when using an string with `(count "")` function

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+* Allow strings on `empty?` (#492)
+* Improved error message when using strings on `count` (#492)
+
 ## 0.7.0 (2022-05-05)
 
 * Added: `merge-with` function
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -316,9 +316,11 @@ implement the PHP Countable interface."
     (php/-> xs (count))
     (if (php/is_array xs)
       (php/count xs)
-      (if (php/=== xs nil)
-        0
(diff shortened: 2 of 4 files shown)


JesusValera (approved):
✅

jenshaase:
I don't know if this is a good idea. Counting strings only works as long as you use ascii chars. `strlen("ö")` for example results in `2` and not `1`. PHP also provides `mb_strlen` to fix this problem. However, using this as standard function is not that great because all the `mb_` functions are very slow.

I would recommend not to integrate strings into the count function. Maybe a new function could do that?

jenshaase:
> But the mb_strlen function would be called only if the variable given is a string.

Sure. I meant the performance between `mb_strlen` and `strlen`. To have a performant `count` function for strings `strlen` would be the best. But that comes with drawbacks. However, if we use `mb_strlen` we have the drawback that it would be slow and it is required to have the `mbstring` extension installed.

At the end of the day the user has to choose between `strlen` and `mb_strlen` anyway. We can't decide that for the user within the `count` function. Therefore I would suggest that we do not support string in the `count` function.

Chemaclass:
@jenshaase alright. In such a case, when using a string, I would display a clearer error message helping with the alternative that the user can use when dealing with `(count "string")`. However, we can easily check for `(empty? "string")`. I will apply these changes in this PR.

jenshaase (approved):

PR #496, opened 2022-06-29
Add .editorconfig & .gitattributes files

JesusValera:
### 📚 Description

This PR adds two config files to the `phel` project which I consider they are very important.

- What is [editorconfig](https://editorconfig.org/)?

**EditorConfig** helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs. The EditorConfig project consists of a **file format** for defining coding styles. EditorConfig files are easily readable and they work nicely with version control systems.

- What are [gitattributes](https://git-scm.com/book/en/v2/Customizing-Git-Git-Attributes)?

You can tell Git not to export certain files or directories when generating an archive. If there is a subdirectory or file that you don’t want to include in your archive file but that you do want checked into your project, you can determine those files via the export-ignore attribute.

In other words, basically, when you require from `composer` the `phel` dependency, you won't download the folders that are ignored.

You can see an example of `PHPUnit` dependency in any project:

image

If you take a look at the `phpunit/phpunit`, you can realise there are more files and they are ignored because of this file 😄 

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+charset = utf-8
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,15 @@
+/.github                export-ignore
+/adrs                   export-ignore
+/docker                 export-ignore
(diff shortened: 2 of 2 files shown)


Chemaclass (approved):
Great idea

PR #520, opened 2022-10-01
Add contains-value? function

Chemaclass:
## 🤔 Background

I think it would be practical if we have a function that checks if a value is inside a collection. Similar as the `contains?` 🤔 

## 🔖 Changes

- Add `contains-value?` function

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1191,6 +1191,14 @@ collection in map"))
       (put res k v))
     (persistent res)))
 
+(defn contains-value?
+  "Returns true if value is present in the given collection, otherwise returns false."
+  [coll val]
+  (let [php-array (to-php-array coll)
+        flipped-array (php/array_flip php-array)
+        flipped (php-array-to-map flipped-array)]
+    (contains? flipped val)))
+
 (defn sort
   "Returns a sorted vector. If no comparator is supplied compare is used."
   [xs & [comp]]
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1194,10 +1194,9 @@ collection in map"))
 (defn contains-value?
   "Returns true if value is present in the given collection, otherwise returns false."
   [coll val]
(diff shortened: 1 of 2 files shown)


JesusValera (approved):

jenshaase:
`array_flip` only works with string and integer values. It is not working for something like this:
```
(contains-value? [:a :b] :a)
```

Also the current implementation is not working for maps:
```
(contains-value? {:a 1 :b 2} 1) # => error
```

I would suggest to solve this problem with existing Phel functions only. Maybe the `for` loop is an option.

jenshaase:
Isn't the `find` function basically what you are looking for? Don't know how it behaves on maps, but you can combine it with the `values` function and it should work.

Chemaclass:
That's a good idea. Actually changing the function to this helps a lot:
```lisp
(defn contains-value?
  "Returns true if value is present in the given collection, otherwise returns false."
  [coll val]
  (if (empty? coll)
    false
    (not (nil? (find |(= $ val) coll)))))
```
And all the tests that I wrote pass, but not these two:
```lisp
(is (false? (contains-value? (set 1) 2)))
(is (true? (contains-value? (set 1) 1)))
```

Screenshot 2022-10-21 at 19 46 02

Any idea why are only these two tests failing? 🤔 
What do you think about the new implementation of the `contains-value?` using `find`, @jenshaase?

jenshaase:
I would suggest to wrap `coll` with the `values` function. This will then also support maps.

```
(defn contains-value?
  "Returns true if value is present in the given collection, otherwise returns false."
  [coll val]
  (if (empty? coll)
    false
    (not (nil? (find |(= $ val) (values coll))))))
```

Chemaclass:
@jenshaase, thanks! Code updated 👌🏻

jenshaase:
Could you add tests for maps?

Chemaclass:
Sure thing, @jenshaase. Done.

PR #522, opened 2022-10-03
Add argv argument to Phel::run()

Chemaclass:
## 🤔 Background

We have a function `Phel::run()` to run a concrete namespace from phel. However, there might be some need to have some concrete `argv` in that particular namespace. Right now, the only way is to setup manually from the outside like 
```php
$GLOBALS['argv'][] = 'additional arguments';
```

## 💡 Goal

To be able to add your additional phel arguments when running a concrete namespace with `Phel::run()` as an optional argument to that function.

## 🔖 Changes

- Add a new argument to the `Phel::run()` funtion ➡️  `array|string $argv`

### DIFF
--- a/src/php/Phel.php
+++ b/src/php/Phel.php
@@ -9,6 +9,8 @@ use Gacela\Framework\Bootstrap\GacelaConfig;
 use Gacela\Framework\Gacela;
 use Phel\Run\RunFacade;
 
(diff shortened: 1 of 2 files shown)


JesusValera (approved):

jenshaase (approved):
Looks good to me.

Issue #523, opened 2022-10-09
Add __name__ to identify the execution of a script


Chemaclass:
## 🤔 Problem

Right now all code get executed on run time. For example:
```lisp
  (loop []
    (play-hand)
    (println)
    (recur))
```
This is a problem when trying to compile the code with `phel compile`, because the code gets stuck on the I/O...

A workaround is to add ask for an additional argument when executing the script ([url](https://github.com/Chemaclass/hello-phel/blob/master/src/rock-paper-scissors.phel#L57)):
```lisp
(if (= "play" (php/aget argv 3))
  (loop []
    (play-hand)
    (println)
    (recur)))
```

This is fine, but tedious and repetitive work that could/should be simplified. A possible solution (the esiest that comes to my mind for now) is to use an approach similar to what python does using `the` __name__ guard like
```python
if __name__ == '__main__':
    do_foo()
    do_bar()
```

## 💡  Proposal

Introduce `__name__` to be able to identify the execution of a script, so we are able to separate the I/O user interaction from the compilation problem when running with multiple scripts.

### Idea 1
```lisp
(if __main__?
  (loop []
    (play-hand)
    (println)
    (recur)))
```
In this case `__main__?` is a function that calculates the value depending on the file which is been executed.

### Idea 2
```lisp
(if (= __name__ "__main__")
  (loop []
    (play-hand)
    (println)
    (recur)))
```
In this case `__name__` is a value that will vary and only the file which is been executed will received the value `__main__`.

Chemaclass:
I got another (potentially better) idea: introducing `__MAIN_NS__` as global constant defined in the `RunCommand`, so we can easily check for that value on any phel script, checking always the namespace. It would look like this:
```clojure
(ns phel\local
  (:use RuntimeException))

(if (= __MAIN_NS__ "phel\\local")
  (println "name A:" __MAIN_NS__ __FILE__))
```

I did already some drafting and it's working. I will create a PR soonish to have a better picture about it. We already have __FILE__ and __DIR__ ([see docs](https://phel-lang.org/documentation/php-interop/#dir-and-file)), so I used them as an inspiration to add the new constant.

What do you think, @jenshaase @JesusValera? Do you like this new idea `__MAIN_NS__`?

Chemaclass:
The idea of this feature was already solved by the existing `*compile-mode*`. See more https://github.com/phel-lang/phel-lang/pull/524#issuecomment-1286987738.
Therefore, there is no need for this anymore.

PR #524, opened 2022-10-11
Define __MAIN_NS__ in RunCommand

Chemaclass:
## 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/523

## 💡 Goal

To be able to ignore phel code when running tests or compiling code, especially code that involves I/O operations.

## 🔖 Changes

- Introduce a new global magic constant `__MAIN_NS__` that will return the value of the main namespace that it's being executed

### 🧪 For example

```clojure
(ns phel\game)

(if (= __MAIN_NS__ "phel\\game")
  (loop []
    (play-hand)
    (println)
    (recur)))
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 * Allow strings on `empty?` (#492)
 * Improved error message when using strings on `count` (#492)
+* Add __MAIN_NS__ global constant (#524)
 
 ## 0.7.0 (2022-05-05)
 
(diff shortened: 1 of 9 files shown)


JesusValera (approved):

Chemaclass:
Oh, gosh, of course it's the same"! I forgot about it, but it's exactly doing what I wanted. So I will close this PR as there is no benefit on it for now 😄 
```lisp
(when-not *compile-mode*
  (try
    (main)
    (catch RuntimeException e
      (println)
      (println (php/-> e (getMessage))))))
```

PR #526, opened 2022-10-24
Refactoring FileCompiler and making CompileOptions optional

Chemaclass:
### 🔖 Changes

- Decouple I/O from `FileCompiler`
- Rename `getCode()` to `getPhpCode()`
- Refactor StatementEmitter
- Make `CompileOptions` argument optional

### DIFF
--- a/src/php/Compiler/Domain/Emitter/StatementEmitter.php
+++ b/src/php/Compiler/Domain/Emitter/StatementEmitter.php
@@ -26,32 +26,25 @@ final class StatementEmitter implements StatementEmitterInterface
         $this->outputEmitter->resetSourceMapState();
 
         $sourceLocation = $node->getStartSourceLocation();
-        $file = $sourceLocation
-            ? $sourceLocation->getFile()
-            : 'string';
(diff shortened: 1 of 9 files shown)


src/php/Compiler/Domain/Emitter/StatementEmitter.php
```diff
@@ -25,33 +25,37 @@ public function emitNode(AbstractNode $node, bool $enableSourceMaps): EmitterRes
...
+        );
+    }
 
+    private function phpCode(AbstractNode $node): string
+    {
         ob_start();
         $this->outputEmitter->emitNode($node);
-        $code = ob_get_clean();
 
+        return ob_get_clean();
```
JesusValera, line 41: @Chemaclass do you think could make sense to move these PHP functions (`ob_start()` & `ob_get_clean()`) to an interface?

JesusValera (approved):

src/php/Compiler/Domain/Emitter/StatementEmitter.php
Chemaclass: That's a really interesting idea for a follow up PR 💡

PR #538, opened 2022-12-11
Update psalm 5.1

JesusValera:
## 📚 Description

Update `phel` to PHP 8.0.2

In order to upgrade Symfony dependencies to 6.x, we need to update the minimum PHP version from 8.0 to 8.0.2.

Also, Psalm was upgraded to 5.1. (reference https://github.com/phel-lang/phel-lang/commit/821457b50f5e6c00a19912bcbc0d33228ad627f3#r91961860)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,10 +17,10 @@
         }
     ],
     "require": {
-        "php": ">=8.0",
+        "php": ">=8.0.2",
         "gacela-project/gacela": "^0.32",
         "phpunit/php-timer": "^5.0",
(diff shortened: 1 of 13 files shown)


Chemaclass:
I fixed a bunch of displaying errors when upgrading psalm to the latest 5.1, but I also suppressed one error `MissingTemplateParam` which I didn't know how to solve and it didn't bring much value as feedback.

Chemaclass (approved):

PR #547, opened 2023-01-02
Refactor phel/json tests

Chemaclass:
### 📚 Context

This is a follow up PR from the original idea https://github.com/phel-lang/phel-lang/pull/489 by @mabasic 

### 🔖 Changes

- Add docs and normalize all `phel/json` functions
- Use `JsonException` instead of `Exception`


src/phel/json.phel
```diff
@@ -1,33 +1,45 @@
 (ns phel\json
-  (:use \Exception))
+  (:use \RuntimeException))
 
-(defn valid-key? [v]
+(defn valid-key?
+  "Check if v is a valid json key."
```
jenshaase, line 5:
```suggestion
  "Checks if `v` is a valid JSON key or can be converted to a JSON key."
```

src/phel/json.phel
```diff
@@ -1,33 +1,45 @@
...
+(defn- encode-value-iterable [x]
+  (let [arr (php/array)]
+    (foreach [k v x]
+      (when-not (valid-key? k)
+        (throw (php/new RuntimeException "Key can only be an integer, float, symbol, keyword or a string.")))
+      (php/aset arr (encode-value k) (encode-value v)))
+    arr))
+
+(defn- encode-value-with-name [x]
+  (str (php/-> x (getName))))
```
jenshaase, line 20: We now have the `name` function. This can be used instead of `getName` (https://phel-lang.org/documentation/api/#name)

jenshaase (changes requested):

src/phel/json.phel
```diff
@@ -1,33 +1,45 @@
...
+
 (defn encode-value [x]
+  "Convert a phel data structure to a 'json compatible' value."
```
jenshaase, line 23:
```suggestion
  "Convert a Phel data type to a 'json compatible' value."
```

src/phel/json.phel
Chemaclass: I forgot about that, thank you! I will use that instead of this new private function `encode-value-with-name`.

src/phel/json.phel
```diff
@@ -1,33 +1,42 @@
...
 
 (defn encode [value & [{:flags flags :depth depth}]]
+  "Returns the JSON representation of a value."
   (let [flags (or flags 0)
         depth (or depth 512)]
-    (when (php/is_resource value) (throw (php/new Exception "Value can be any type except a resource.")))
-    (when-not (int? flags) (throw (php/new Exception "Flags must be an integer.")))
-    (when-not (int? depth) (throw (php/new Exception "Depth must be an integer.")))
-    (when-not (> depth 0) (throw (php/new Exception "Depth must be greater than zero.")))
+    (when (php/is_resource value) (throw (php/new RuntimeException "Value can be any type except a resource.")))
```
JesusValera, line 32: What about instead of throwing a RuntimeException, return a [JsonException](https://www.php.net/manual/en/class.jsonexception.php) as PHP does?

Chemaclass: Good idea. I will change it 👍

JesusValera (approved):

PR #550, opened 2023-01-17
Fix docs for json functions

Chemaclass:
### 🤔 Background

I just noticed that the docs for the json functions are in the wrong place, because they are ignored in our API page

Screenshot 2023-01-17 at 22 00 21

I easily noticed this because `valid-key?` is the only function with docs in the API page.

### 💡 Goal

Fix the docs for the `phel\json` functions

### 🔖 Changes

- Move between the function name and the arguments the function docs

### DIFF
--- a/src/phel/json.phel
+++ b/src/phel/json.phel
@@ -16,8 +16,9 @@
       (php/aset arr (encode-value k) (encode-value v)))
     arr))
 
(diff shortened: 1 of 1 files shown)


JesusValera (approved):
Oh, nice 🙂

PR #551, opened 2023-01-18
Create Api module

Chemaclass:
### 🤔 Background

I originally created this repo https://github.com/phel-lang/phel-normalized-internal to be able to normalized all internal phel functions in a way that we could easily iterate over them from a PHP application; the main motivation to have this was/is to build the API page from the website (see: https://github.com/phel-lang/phel-lang.org/blob/master/build/src/php/FileGenerator/Domain/ApiMarkdownGenerator.php#L25)

### 💡 Goal

Simplify the logic and use the `phel-lang` main repo for this purpose.

### 🔖 Changes

- Moved the logic of the `phel-normalized-internal` repo inside the `phel-lang` repo in a module named `Api`
  - The idea behind the `Api` module is to expose information of the language itself in a PHP form

#### 🖼️ Example of usage ([link](https://github.com/phel-lang/phel-lang.org/pull/48))

```php
$groupedFns = (new ApiFacade)->getNormalizedGroupedFunctions([
    'phel\\core',
    'phel\\http',
    'phel\\html',
    'phel\\test',
    'phel\\json',
]);
```

> **Note**: Deprecate/remove the [phel-normalized-internal](https://github.com/phel-lang/phel-normalized-internal) repo in favor of using the logic from the [phel-lang](https://github.com/phel-lang/phel-lang).


src/php/Api/Infrastructure/PhelFnLoader.php
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    private function loadAllPhelFunctions(): void
+    {
+        if (self::$phelInternalDocLoaded) {
+            return;
+        }
+
+        $phelFile = __DIR__ . '/phel/doc.phel';
+        $this->runFacade = new RunFacade();
```
JesusValera, line 56: We are injecting by constructor this `RunFacade`, I would say we don't need to initialize it here.

src/php/Api/Transfer/NormalizedPhelFunction.php
```diff
@@ -0,0 +1,48 @@
...
+final class NormalizedPhelFunction
+{
+    public function __construct(
+        private string $fnName,
+        private string $doc,
+        private string $fnSignature,
+        private string $description,
+    ) {
+    }
+    /**
```
JesusValera, line 16:
```suggestion
    }

    /**
```

Missing space.

tests/php/Integration/Api/ApiFacadeTest.php
```diff
@@ -0,0 +1,30 @@
...
+final class ApiFacadeTest extends TestCase
+{
+    public static function setUpBeforeClass(): void
+    {
+        Gacela::bootstrap(__DIR__, GacelaConfig::withPhpConfigDefault());
+    }
+
+    public function test_number_of_grouped_functions(): void
+    {
+        // This test in isolation works, but when running with all other integration tests fails,
```
JesusValera, line 21: 🤔

tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
```diff
@@ -0,0 +1,310 @@
...
+
+        self::assertEquals($expected, $actual);
+    }
+
+    public function test_symbol_with_doc_and_desc(): void
+    {
+        $symbol = $this->createStub(PersistentMapInterface::class);
+        $symbol->method('offsetExists')->willReturn(true);
+        // false -> relates to `isPrivate`
+        // '...' -> relates to `doc`
```
JesusValera, line 217:
```suggestion
        // 'Constant for (...)' -> relates to `doc`
```
Maybe it's a bit clearer this way? 👀

JesusValera: Or even?:
```php
$symbol->method('offsetGet')->willReturnOnConsecutiveCalls(
    false,                                     // relates to 'isPrivate'
    'Constant for Not a Number (NAN) values.', // relates to 'doc'
);
```

tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
```diff
@@ -0,0 +1,310 @@
...
+        ];
+
+        self::assertEquals($expected, $actual);
+    }
+
+    public function test_symbol_with_doc_and_desc_and_signature(): void
+    {
+        $symbol = $this->createStub(PersistentMapInterface::class);
+        $symbol->method('offsetExists')->willReturn(true);
+        // false -> relates to `isPrivate`
```
JesusValera, line 249: Same here as before.

tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
```diff
@@ -0,0 +1,310 @@
...
+        self::assertEquals($expected, $actual);
+    }
+
+    public function test_symbol_with_desc_with_link(): void
+    {
+        $symbol = $this->createStub(PersistentMapInterface::class);
+        $symbol->method('offsetExists')->willReturn(true);
+        // false -> relates to `isPrivate`
+        // '...' -> relates to `doc`
+        $symbol->method('offsetGet')->willReturnOnConsecutiveCalls(
```
JesusValera, line 284: And finally here.

tests/php/Integration/Api/ApiFacadeTest.php
```diff
@@ -0,0 +1,30 @@
...
+        // because the IntegrationTest loads already the core and all internal code, which
+        // it crashes when it tries to load all phel funcs again. See: `PhelFnLoader::loadAllPhelFunctions()`
+        $this->markTestSkipped('Useful to debug the facade, but useless to keep it in the CI');
+        $facade = new ApiFacade();
+        $groupedFns = $facade->getNormalizedGroupedFunctions();
+
+        self::assertCount(212, $groupedFns);
```
JesusValera, line 28: As soon as we create a new function, this number will fail, isn't it?
I would suggest doing any other test for this to avoid updating this test each time we add a new one, it can become a bit tedious, what do you think? I cannot suggest any other possible test at the moment 😄

src/php/Api/Domain/PhelFnNormalizer.php
```diff
@@ -0,0 +1,80 @@
...
+
+        return strtolower(rtrim($key, '-'));
+    }
+
+    /**
+     * The $desc is in Markdown format, the regex transforms links `[printf](https://...)` into `<i>printf</i>`.
+     */
+    private function formatDescription(string $desc): string
+    {
+        return preg_replace(
```
JesusValera, line 67: We are doing this `markdown` to `HTML` translation here (in the `Domain` layer!), do we want to keep it here or maybe do this transformation somewhere else at a lower level (maybe defining the output in some extra params in the constructor of this class and/or even in `phel-lang.org` project while we output the data?)

src/php/Api/Infrastructure/PhelFnLoader.php
Chemaclass: Ups, I forgot to remove this line 😂

tests/php/Integration/Api/ApiFacadeTest.php
Chemaclass: I spent some time trying to avoid this comment... But I couldn't make it work after an hour. This is clearly an smell that should be improved, but I didn't see it as a blocker for the all concept/PR. I will try to make it better in a future iteration. Anyway, feel free to try it on your own if you feel brave enough hehe

tests/php/Integration/Api/ApiFacadeTest.php
Chemaclass: Exactly... That was another reason to skip the whole test. This is a useful test for integration, to verify that the whole thing works isolated, otherwise it's a fragile test that will break whenever we add new functions. 
This might be subjective at this point. I don't have a strong opinion on this at this moment 🤔

jenshaase:
One thing that was bothering me in the last release is the `doc.phel` file: I needed to change the content of the file to get the JSON documentation added. I think it would be better if we can provide a list of namespaces from the outside to this module. What do you think?

Chemaclass:
Yes, I like that idea a lot @jenshaase . That by default you get the funcs from phel/core and if you want more you simply pass the additional namespaces as argument, and the facade will resolve it.

This way, if we introduce a new nos in the future, we don't need to change this code (like what happened with the (doc) but simply update the client caller.

I will have a look at this idea later 🤘

src/php/Api/Domain/PhelFnNormalizer.php
Chemaclass: Yes, I will change it to keep the link as `<a>` instead of `<i>`. I think it's ok to keep this thing in the domain layer.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+* New Api module which exposes (via the `ApiFacade`) the internal functions of phel for PHP (#551)
```
JesusValera, line 7: Does it make sense to add a command at `phel` executable file (like: `php phel generate-api`) that output the API?

jenshaase, line 7: New Api module which exposes (via the `ApiFacade`) the functions documentation of Phel (#551)

tests/php/Integration/Api/ApiFacadeTest.php
jenshaase: Another thing: The test looks like it also contains the `phel\core` function. But the aren't in the provided list. Don't know if there is an easy fix for the problem.

Chemaclass:
> @jenshaase Another thing: The test looks like it also contains the phel\core function. But the aren't in the provided list. Don't know if there is an easy fix for the problem.

Actually, there is no need to mention the `phel\core` as these functions are always loaded by default. Maybe I should filter them out if we don't call for them? That might be an interesting idea 🤔

tests/php/Integration/Api/ApiFacadeTest.php
Chemaclass: Actually, there is no need to mention the `phel\core` as these functions are always loaded by default. Maybe I should filter them out if we don't call for them? That might be an interesting idea 🤔

CHANGELOG.md
Chemaclass: I don't see the need for now. But feel free to try it out in a future iteration if you want.

Chemaclass: Actually, I was thinking about this, and I do see the utility of a command that will render the doc of all known phel functions, something like this (which might accept an arg to define an specific ns, or even search by any chars)
Screenshot 2023-01-25 at 21 32 51

But this will be definitely in a follow up PR.

Chemaclass: PR: https://github.com/phel-lang/phel-lang/pull/552

PR #553, opened 2023-01-27
Remove generated temporal files

Chemaclass:
### 🤔 Background

Every time we run any **phel command** (like run, test, doc, etc...), this generates temporal files (with the executable PHP code, which later gets required and run inside the PHP interpreter.

However, these files are useless after the execution, and running again another command will generated a bunch of new files again and again. The temporal files will get drop after some point, but these files are not needed at all by default.

>**Note**: the inspiration for this PR was mainly after working on the `phel doc` see [PR](https://github.com/phel-lang/phel-lang/pull/552), in which I realised I am running that command a lot of times and I am generating everytime a lot of temporal files... which made me think, "why don't we remove those temporal files after the phel command execution if they are not really needed?" I don't see the need for those files, but for debugging. Especially now that we have the `phel compile` command.

### 💡 Goal

Remove the generated files after the phel command execution finished.
Allow the possibility of keeping those "generated temporal files", which are useful for debugging purposes. 

### 🔖 Changes

- Create a new module `Filesystem` in which we will keep an eye on the files that we generate, so we can clear them all later after the phel command finished
- The new config key `'keep-generated-temp-files'` will decide if we want to remove or keep the generated files (default: false). This will be useful for debugging.


phel
```diff
@@ -19,4 +20,9 @@ require $autoloadPath;
 Gacela::bootstrap($projectRootDir, Phel::configFn());
 
 $bootstrap = new ConsoleBootstrap();
-$bootstrap->run();
+$bootstrap->setAutoExit(false);
+$exitCode = $bootstrap->run();
+
+(new FilesystemFacade())->clearAll();
```
JesusValera, line 26: Can we place the `(new FilesystemFacade())->clearAll();` line inside the `$bootstrap->run()`? 🤔

tests/php/Integration/Filesystem/FilesystemFacadeTest.php
```diff
@@ -0,0 +1,63 @@
...
+        Gacela::bootstrap(__DIR__, static function (GacelaConfig $config): void {
+            $config->addAppConfigKeyValue(FilesystemConfig::KEEP_GENERATED_TEMP_FILES, true);
+        });
+
+        $filename = tempnam(sys_get_temp_dir(), '__test');
+        $this->filesystem->addFile($filename);
+
+        self::assertFileExists($filename);
+        $this->filesystem->clearAll();
+        self::assertFileExists($filename);
```
JesusValera, line 61: Should we unlink the temporal file once the test was done? (like preserving the original state and don't create "waste")

phel
Chemaclass: Yes, good idea! I will do that

src/php/Filesystem/Domain/FakeFilesystem.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Filesystem\Domain;
+
+final class FakeFilesystem implements FilesystemInterface
```
JesusValera, line 7: In my opinion, the word `fake` is more related to testing.

In fact, you are using the `Null Object Design Pattern`, I think maybe make more sense to rename the class from `FakeFileSystem` to  `NullObjectFileSystem` or `NullFileSystem`.

src/php/Filesystem/Infrastructure/RealFilesystem.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Phel\Filesystem\Infrastructure;
+
+use Phel\Filesystem\Domain\FilesystemInterface;
+
+final class RealFilesystem implements FilesystemInterface
```
JesusValera, line 9: What about renaming this class (and probably the interface?) to `RealFilesystemCollection`?

src/php/Filesystem/Infrastructure/RealFilesystem.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        self::$files[] = $file;
+    }
+
+    public function clearAll(): void
+    {
+        foreach (self::$files as $file) {
+            unlink($file);
+        }
+        self::$files = [];
```
JesusValera, line 29:
```suggestion
        $this->reset();
```

src/php/Filesystem/Infrastructure/RealFilesystem.php
JesusValera: To be honest, not sure if I really like the idea, but as a suggestion to take it into account. 😄

Chemaclass: I don't think I like the Collection suffix here, but thanks for the idea :)

src/php/Filesystem/Domain/FakeFilesystem.php
Chemaclass: I will rename it to `NullFileSystem`

JesusValera (approved):

mabasic:
This is great. Does this mean that after running phel code files generated in /tmp will be removed?

Chemaclass:
Exactly, @mabasic. I remembered your case when I was doing this task :)
An easy way to understand the different between **run** and **build** is:

The `run` command will:
1.- transpile phel to php creating temp files
2.- php will read those temp files
3.- the temp files will be removed

The `build` command will:
1.- transpile phel to php creating temp files
2.- copy the temp files to the out/ dir
3.- the temp files will be removed

This way with the build command you can run the transpiled php files directly avoiding the transpilation time from phel to php, so it will be more performance. Either way, the temp files will be always removed by default.

Issue #540, opened 2022-12-17
Create int function


JesusValera:
Currently, we have the `int?` function that checks if the current value is `int` value or not (returns a bool).
But we don't have an `int` function that casts a variable to an `int`.
I propose a new set of functions to cast as follows:
```lisp
(int "42") # output: 42 (integer)
```
Same for bool, float, string.

Does it make sense to add also for vector, list, map & set?

---

Additionally, maybe we could also add a super cast function with a match that can cast a value to another type, maybe something like:

```lisp
(cast int "42") # output: 42 (integer)
(cast string 3.5) # output "3.5" (string)
```

What do you think @jenshaase @Chemaclass ?

Chemaclass:
You can do this using native PHP: `(php/intval i)` (eg [I used it here](https://github.com/Chemaclass/phel-snake/blob/master/src/logic.phel#L74)), however, I think it might be handy to have this as an internal native phel function `(int i)`. What do you think, @jenshaase?

Chemaclass:
I created a PR with this idea https://github.com/phel-lang/phel-lang/pull/554 so I will close this issue and we can follow the conversation in the PR instead.

PR #554, opened 2023-01-28
Create parse-int function

Chemaclass:
## 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/540

## 🔖 Changes

- Added `parse-int` function

## 🖊️ Draft-PR

Although, I am not sure if this function (and similars) make 100% sense in the core, because it's just an alias for the native `intval` function that you can use simply with `(php/intval x)`. I am fixed feelings about this, but anyway, I wanted to do this PR to share the idea with you two @jenshaase & @JesusValera.

>**Note**: The bigger risk that I see is that the more functions we add to the core, the more it might take to run/compile because it needs to process more functions. And it wouldn't be clever to add ALL type of functions, especially if you can already use the native ones from PHP straight forward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+* Added `parse-int` function
 * New Api module which exposes (via the `ApiFacade`) the functions documentation of Phel (#551)
 
 ## 0.8.0 (2023-01-16)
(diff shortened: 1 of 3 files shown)


jenshaase:
I don't see any benefits in doing this. Calling the PHP function directly is even more efficient than wrapping it in another function.

JesusValera:
Yes, I agree.
I don't see much value at the moment in this situation. 👍🏼

PR #559, opened 2023-01-29
Replace switch by match in Parser->readExpression()

JesusValera:
## 📚 Description

Currently, we have the `Parser::readExpression()` method with a `D` grade in [Scrutinizer](https://scrutinizer-ci.com/g/phel-lang/phel-lang/code-structure/master/operation/Phel%5CCompiler%5CDomain%5CParser%5CParser::readExpression).

image

I think that is happening due to the multiple `return` statements the method has, hopefully  (but I am not 100% sure) with the `match` statement due to the single `return`, the qualification might improve.


Chemaclass (approved):
I like this idea 💡

PR #562, opened 2023-02-01
Create test-numbers-with-underscore

Chemaclass:
### 📚  Description

Adding phel tests for numbers with underscore as the docs mentions:
https://phel-lang.org/documentation/numbers-and-arithmetic/#numbers

### DIFF
--- a/tests/phel/test/core/math-operation.phel
+++ b/tests/phel/test/core/math-operation.phel
@@ -103,6 +103,6 @@
   (is (= 2 (mean [1 2 3])) "(mean [1 2 3])"))
 
 (deftest test-numbers-with-underscore
-  (is (= 1337 b101_0011_1001) "binary number with underscores")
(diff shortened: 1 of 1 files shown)


tests/phel/test/core/math-operation.phel
```diff
@@ -101,3 +101,8 @@
 
 (deftest test-mean
   (is (= 2 (mean [1 2 3])) "(mean [1 2 3])"))
+
+(deftest test-numbers-with-underscore
+  (is (= 1337 b101_0011_1001) "binary number with underscores")
```
jenshaase, line 106:
```suggestion
  (is (= 1337 0b101_0011_1001) "binary number with underscores")
```

jenshaase:
It should be as described here: https://www.php.net/manual/en/language.types.integer.php

PR #564, opened 2023-02-01
Allow underscores in decimal numbers

Chemaclass:
### 🤔 Background

 Idea: https://github.com/phel-lang/phel-lang/discussions/563
We can use underscore for binary, octal and hexa but not decimal.

### 💡 Goal

To be able to use underscore on decimal numbers. Supported from PHP 7.4

### 🔖 Changes

- Changed AtomParser to check for valid decimal numbers with underscores as well

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 * Rename command `phel compile` to `phel build` (#555)
 * Added config parameter `ignore-when-building` (#557)
 * Added config parameter `keep-generated-temp-files` (#553)
+* Allow underscores in decimal numbers (#564)
 
 ## 0.8.0 (2023-01-16)
 
--- a/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
+++ b/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
@@ -23,7 +23,6 @@ final class AtomParser
     private const REGEX_HEXADECIMAL_NUMBER = '/^([+-])?0[xX][0-9a-fA-F]+(_[0-9a-fA-F]+)*$/';
     private const REGEX_OCTAL_NUMBER = '/^([+-])?0[0-7]+(_[0-7]+)*$/';
     private const REGEX_DECIMAL_NUMBER = '/^([+-])?[0-9]+(_[0-9]+)*[\.(_[0-9]+]?|0$/';
-//    private const REGEX_DECIMAL_NUMBER = '/^([+-])?[0-9]+(_[0-9]+)*|0$/';
 
     public function __construct(private GlobalEnvironmentInterface $globalEnvironment)
     {
--- a/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
+++ b/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
@@ -22,7 +22,8 @@ final class AtomParser
     private const REGEX_BINARY_NUMBER = '/^([+-])?0[bB][01]+(_[01]+)*$/';
     private const REGEX_HEXADECIMAL_NUMBER = '/^([+-])?0[xX][0-9a-fA-F]+(_[0-9a-fA-F]+)*$/';
     private const REGEX_OCTAL_NUMBER = '/^([+-])?0[0-7]+(_[0-7]+)*$/';
-    private const REGEX_DECIMAL_NUMBER = '/^([+-])?[0-9]+(_[0-9]+)*$/';
+    private const REGEX_DECIMAL_NUMBER = '/^([+-])?[0-9]+(_[0-9]+)*[\.(_[0-9]+]?|0$/';
+//    private const REGEX_DECIMAL_NUMBER = '/^([+-])?[0-9]+(_[0-9]+)*|0$/';
 
     public function __construct(private GlobalEnvironmentInterface $globalEnvironment)
     {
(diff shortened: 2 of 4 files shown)


src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
```diff
@@ -44,50 +48,33 @@ public function parse(Token $token): AbstractAtomNode
...
         }
 
         if (is_numeric($word)) {
             return new NumberNode($word, $token->getStartLocation(), $token->getEndLocation(), $word + 0);
         }
 
+        if (preg_match(self::REGEX_DECIMAL_NUMBER, $word, $matches)) {
+            return $this->parseDecimalNumber($matches, $word, $token);
+        }
+
```
JesusValera, line 70: Not sure if I understood properly.
What is the difference between the two "numeric" methods (method from L63 & L67)?
Are not both of them returning a `new NumberNode()`? Maybe both conditions could be simplify into one?

I have done some tests with `is_numeric()` PHP method and it's interesting:
```php
$tests = array(
    123,     # is_numeric is TRUE
    1_23,    # is_numeric is TRUE
    ' 123 ', # is_numeric is TRUE
    -1_23,   # is_numeric is TRUE
    '1_23 ', # is_numeric is FALSE
);

foreach ($tests as $element) {
    if (is_numeric($element)) {
        echo var_export($element, true) . ' is numeric', PHP_EOL;
    } else {
        echo var_export($element, true) . ' is NOT numeric', PHP_EOL;
    }
}
```

JesusValera (commented):
Small question :)

src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
Chemaclass: You answered yourself: `is_numeric('1_23') # is_numeric is FALSE`, 
but it's a valid decimal number with underscore.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 * Rename command `phel compile` to `phel build` (#555)
```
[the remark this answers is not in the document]
Chemaclass: I am not sure, I don't mind either. Any preference, @jenshaase?

src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
JesusValera: What I mean is if it maybe makes sense to do something like this:
```php
public function parse(Token $token): AbstractAtomNode
{
    // ...

    if ($this->isNumeric($word)) {
        return $this->parseDecimalNumber($matches, $word, $token);
    }

    // ...
}

private function isNumeric(string $word): bool
{
    return is_numeric($word)
        || preg_match(self::REGEX_DECIMAL_NUMBER, $word, $matches);
}
```

Chemaclass: Oh, I didn't get it before. Good idea! Feel free to commit and push this yourself in this branch :)

CHANGELOG.md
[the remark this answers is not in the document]
jenshaase: I don't mind. For me it is enough.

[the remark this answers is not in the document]
Chemaclass: Then I will leave it as it is.

src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
```diff
@@ -60,11 +60,7 @@ public function parse(Token $token): AbstractAtomNode
             return $this->parseOctalNumber($matches, $word, $token);
         }
 
-        if (is_numeric($word)) {
-            return new NumberNode($word, $token->getStartLocation(), $token->getEndLocation(), $word + 0);
-        }
-
-        if (preg_match(self::REGEX_DECIMAL_NUMBER, $word, $matches)) {
+        if (is_numeric($word) || preg_match(self::REGEX_DECIMAL_NUMBER, $word)) {
```
Chemaclass, line 63: This change broke the tests, @JesusValera 😆

JesusValera: I am trying to understand why the test is failing and it seems the part of
```php
$parser->parse(
    new Token(Token::T_ATOM, '1.2', $start, $end),
),
```
in the test is not creating a "proper" `NumberNode` as the `value` is 1, instead of `1.2`, is that an error @jenshaase?


image

Chemaclass: But... the test shouldn’t fail, no? Maybe @jenshaase can help us understanding if the test is valid, or wrong and we need to update it 🤔

JesusValera: The problem is in the following method
```php
private function parseDecimalNumber(array $matches, string $word, Token $token): NumberNode
{
    $sign = (isset($matches[1]) && $matches[1] === '-') ? -1 : 1;

    return new NumberNode(
        $word,
        $token->getStartLocation(),
        $token->getEndLocation(),
        $sign * (int)str_replace('_', '', $word),
    );
}
```
In the last line, we are casting `$word` to integer, so, for decimal numbers like `1.2` are transformed into `1`. That's why some tests fail. I also tried to change the cast to `float` as follows
```php
$sign * (float)str_replace('_', '', $word),
```
but then, there are a lot of errors everywhere.

Changes were rollback, like this is ok 👌🏼

JesusValera (approved):

PR #565, opened 2023-02-03
ApiFacade returns a list of PhelFunctions

Chemaclass:
### 🤔 Background

Currently, the `ApiFacade` returns the phelFunctions groups by a normalized key that it's useful when rendering the API page in the webshop (and might be useful for other grouping reasons) but it's weird to have that grouping in the main method of the `ApiFacade`.

### 💡 Goal

Simplify the `ApiFacade` entry point.

### 🔖 Changes

- Instead of "grouping the phelFunctions" from the `ApiFacade`, rather move that `groupKey` as a property of the `PhelFunction` and simplify the signature of the `ApiFacadeInterface`.
- Added an ADR for the Api Module

### DIFF
--- a/tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
+++ b/tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
@@ -12,6 +12,17 @@ use PHPUnit\Framework\TestCase;
 
 final class PhelFnNormalizerTest extends TestCase
 {
+    public function test_no_functions_found(): void
+    {
+        $phelFnLoader = $this->createMock(PhelFnLoaderInterface::class);
+        $phelFnLoader->method('getNormalizedPhelFunctions')->willReturn([]);
+
+        $normalizer = new PhelFnNormalizer($phelFnLoader);
+        $actual = $normalizer->getPhelFunctions();
+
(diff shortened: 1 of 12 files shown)


tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
```diff
@@ -20,17 +31,16 @@ public function test_group_key_one_function(): void
...
-            'fn-name' => [
-                PhelFunction::fromArray([
-                    'fnName' => 'fn-name',
-                    'doc' => '',
-                    'fnSignature' => '',
-                    'desc' => '',
-                ]),
-            ],
+            PhelFunction::fromArray([
+                'fnName' => 'fn-name',
```
JesusValera, line 38: @Chemaclass do we have to update also the webpage project?

JesusValera (approved):

tests/php/Unit/Internal/Domain/PhelFnNormalizerTest.php
Chemaclass: Yes, I have already the code prepared. I will create a PR on the web repo once this PR is merged.

PR #567, opened 2023-02-07
Add testdox option

Chemaclass:
## 🤔 Background

You can use the optional option `--testdox` in PHPUnit, which renders a nicer output. For example:
Screenshot 2023-02-07 at 22 58 58
It would be handy to have such option when running the `phel test` command. 

## 🔖 Changes

I introduced a similar option for our `phel test`. It doesn't render the test-name (I didn't know how to do it), but I could render the test message. 

> **Note**: we can refactor it once we figure out how to use the test-name instead of the test message. At least when there is no test message.

### Here an example without and with the `--testdox` option
Screenshot 2023-02-13 at 10 24 43

> It's not as nice as the one from PHPUnit, but step by step 😄 

## 💭 Follow up ideas

IDEA 1: Add the running/execution time for each test next to the text itself (see the PHPUnit screenshot).
IDEA 2: Add color for the **fail** and **pass** checks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 * Added default format paths: 'src', 'tests' (#569)
 * Deprecate `*compile-mode*` in favor of `*build-mode*` (#570)
+* Added `--testdox` argument to `phel test` command (#567)
 
 ## 0.9.0 (2023-02-05)
 
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -8,17 +8,21 @@
 # Report
 # ------
 
+(def- testdox? (var false))
+
 (def- stats (var {:failed []
                   :counts {:failed 0
                            :error 0
                            :pass 0
                            :total 0}}))
 
+(defn- print-report-testdox-state [data state]
+  (let [s (case state :failed "✘" :pass "✔" :error "E")
+        message (or (:message data) "no test message found")]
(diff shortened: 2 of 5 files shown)


Chemaclass:
I didn't know how to get the `test-name` to be able to render it inside the `print-report-testdox-state`, because this is running from the `is` macro, and the test-name is inside the `deftest` macro. Could you help me out with this, @jenshaase? Either directly in this branch, or in a follow up PR, as you prefer.

jenshaase:
@Chemaclass I can take a look.

There might be a completely different solution. PHPUnit 10 introduced a event system. Maybe we can use this event system to rebuild our test library (same syntax as before, but PHPUnit as a base). Than we could use all PHPUnits outputs. I haven't look in the details and don't know if it is working but maybe we could give it a try.

jenshaase:
@Chemaclass See my recent commit for a solution using `binding`. You probably want to change how the message is displayed

Chemaclass:
I love that, @jenshaase! Just one question, is there a convention about naming-variables regarding adding the `*` prefix/suffix? eg see in `phel/test.phel` -> `*current-test-name*`, `testdox?`, `stats`

Chemaclass:
I would say, we can merge this now, and in a follow up PR we can add colors, and/or the timing for each test. If you're ok, @jenshaase @JesusValera feel free to approve the PR and merge it :)

JesusValera (approved):

Issue #568, opened 2023-02-08
Add default format paths


Chemaclass:
## 🤔 Current behaviour

The current `vendor/bin/phel format` expect a mandatory argument with the list of paths to run the format.

<img width="431" alt="Screenshot 2023-02-08 at 15 59 05" src="https://user-images.githubusercontent.com/5256287/217566453-31a9baac-8285-425e-b11e-53d1a5841a7b.png">

## 💡 Acceptance criteria

Introduce a new "phel config parameter" (in `phel-config.php`) named `'format-dirs'` which will be the default paths that the `format` command will use when there were no arguments provided.

<img width="401" alt="Screenshot 2023-02-08 at 16 02 17" src="https://user-images.githubusercontent.com/5256287/217567205-b3f224cd-12a9-4e5f-80bb-06000c37c034.png">

>**Note**: You can find the FormatCommand here: `src/php/Formatter/Infrastructure/Command/FormatCommand.php` 

> HINT: Make the `paths` argument OPTIONAL, and then if it's empty, then take the once from the config.

cgrabenstein:
I opened a PR for that issue. Let me know if there's anything I can improve!

PR #569, opened 2023-02-09
Add default format paths

cgrabenstein:
### 🤔 Background

This is the implementation for ticket #568

### 💡 Goal

The format command can be run without passing any paths. If no paths are being passed, the default paths "src" and "tests" should be used, as defined in the config.

### 🔖 Changes

In addition to the changes of the command, I updated the test cases to use the CommandTester from Symfony. This allows to run the command without having to mock the input and therefore allows to test the FormatterCommand without passing any paths.

### DIFF
--- a/tests/php/Integration/Formatter/Command/Format/FormatCommandTest.php
+++ b/tests/php/Integration/Formatter/Command/Format/FormatCommandTest.php
@@ -4,11 +4,12 @@ declare(strict_types=1);
 
 namespace PhelTest\Integration\Formatter\Command\Format;
 
+use Gacela\Framework\Bootstrap\GacelaConfig;
 use Gacela\Framework\Gacela;
+use Phel\Formatter\FormatterConfig;
 use Phel\Formatter\Infrastructure\Command\FormatCommand;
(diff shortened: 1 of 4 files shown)


tests/php/Integration/Formatter/Command/Format/FormatCommandTest.php
```diff
@@ -24,63 +25,69 @@ public function test_good_format(): void
...
     }
 
-    public function test_bad_format(): void
+    public function test_command_uses_default_paths(): void
     {
         $path = self::FIXTURES_DIR . 'bad-format.phel';
         $oldContent = file_get_contents($path);
 
-        $command = $this->createFormatCommand();
+        Gacela::bootstrap(__DIR__, static function (GacelaConfig $config) use ($path): void {
```
JesusValera, line 44: @Chemaclass does it make sense to bootstrap Gacela twice (it was already bootstrapped in the `setUpBeforeClass()`?

JesusValera (approved):
Thank you!

Chemaclass (approved):
Brilliant, great PR! Thank you 🙏

PR #570, opened 2023-02-11
Rename compile-mode to build-mode

Chemaclass:
### 🤔 Background

We can use the form `*compile-mode*` to identify if we are during building mode (former "compilation").
We renamed the command from `phel compile` to `phel build`, so it would make sense to use `*build-mode*` 

### 💡 Goal

Deprecate `*compile-mode*` in favor of `*build-mode*`

### 🔖 Changes

- Create a constant for `BuildConstants::BUILD_MODE = '*build-mode*'`
- Moved constant from `GlobalEnvironment::PHEL_CORE_NAMESPACE` to `CompilerConstants::PHEL_CORE_NAMESPACE`
- Create a new dir `Shared` to store these new constants interfaces (`BuildConstants`, `CompilerConstants`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 * Added default format paths: 'src', 'tests' (#569)
+* Renamed `*compile-mode*` to `*build-mode*` (#570)
 
 ## 0.9.0 (2023-02-05)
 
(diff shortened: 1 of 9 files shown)


JesusValera (commented):
The change makes sense.
I was wonder if this constant is only use internally or the client can define it manually.
If the second condition applies, maybe I would suggest to create an alias and deprecate the `*compile-mode*` key instead.

Chemaclass:
@JesusValera this feature allows you to execute the main functionality of your app only when it's NOT on compile/building mode. See for example: https://github.com/Chemaclass/phel-cli-gui/blob/master/src/phel/test-keyboard.phel#L19

I like your idea of don't removing it now, but mark it as deprecated, and remove it in the release afterwards.

JesusValera (approved):

Issue #494, opened 2022-06-25
Create constants for phel-config keys


Chemaclass:
### Current behaviour

Currently we have to use raw-strings when defining the keys for the phel-config file.

```php
# source: https://github.com/phel-lang/phel-lang/blob/master/phel-config.php
return [
    'src-dirs' => ['src/phel'],
    'test-dirs' => ['tests/phel'],
    'vendor-dir' => 'vendor',
    'export' => [
        'directories' => ['src/phel'],
        'namespace-prefix' => 'PhelGenerated',
        'target-directory' => 'src/PhelGenerated',
    ],
    'out-dir' => 'out',
];
```

### Desired behaviour

Define and use constants for them. Something like this:

```php
return [
    PhelConfig::SOURCE_DIRECTORIES => ['src/phel'],
    PhelConfig::TEST_DIRECTORIES=> ['tests/phel'],
    PhelConfig::VENDOR_DIRECTORY => 'vendor',
    PhelConfig::EXPORT=> [
        PhelConfig::EXPORT_DIRECTORIES => ['src/phel'],
        PhelConfig::EXPORT_NAMESPACE_PREFIX => 'PhelGenerated',
        PhelConfig::EXPORT_TARGET_DIRECTORY => 'src/PhelGenerated',
    ],
    PhelConfig::OUT_DIRECTORY=> 'out',
];
```

----

Actually, by doing this, we might even don't need the nested level of "export" and keep it like this:
```php
return [
    PhelConfig::SOURCE_DIRECTORIES => ['src/phel'],
    PhelConfig::TEST_DIRECTORIES=> ['tests/phel'],
    PhelConfig::VENDOR_DIRECTORY => 'vendor',
    PhelConfig::EXPORT_DIRECTORIES => ['src/phel'],
    PhelConfig::EXPORT_NAMESPACE_PREFIX => 'PhelGenerated',
    PhelConfig::EXPORT_TARGET_DIRECTORY => 'src/PhelGenerated',
    PhelConfig::OUT_DIRECTORY=> 'out',
];
```

What do you think about this, @jenshaase? Or do you see a simpler approach?

jenshaase:
Actually I would like to keep the nested structure.

An object orientated way builder class would probably be the better solution:

```php
return (new PhelConfigBuilder())
  ->setSrcDirs(['src/phel'])
  ->setTestDirs(['tests/phel'])
  ->setVendorDir('vendor')
  ->setExport(
[...]

PR #572, opened 2023-02-13
Added fluent config

cgrabenstein:
### 🤔 Background

This is an attempt to create a fluent config as described in #494 

### 💡 Goal

The goal is to have an easier way to configure phel, by providing a fluent config.

### 🔖 Changes

Added config classes and made the main config class implement the `JsonSerializable` interface.

This is just a draft since I did not add any tests yet. But I wasn't sure whether that is the right way to go. I tried to find the most gacela-way to implement this but I might as well have missed another, easier solution. If you agree on my implementation (or at least the idea behind it), I'll gladly finalize the PR.


cgrabenstein:
I just realized that the `...Config` classes could also implement the `JsonSerializable` Interface. That will work with the existing Reader and is an even simpler solution.

I will update the PR later today!

phel-config.php
```diff
@@ -2,17 +2,16 @@
...
-        'namespace-prefix' => 'PhelGenerated',
-        'target-directory' => 'src/PhelGenerated',
-    ],
-    'out-dir' => 'out',
-    'ignore-when-building' => ['src/phel/local.phel'],
-    'keep-generated-temp-files' => false,
-    'format-dirs' => ['src', 'tests'],
-];
+return (new \Phel\Config\PhelConfig())
+    ->setSrcDirs(['src/phel'])
```
JesusValera, line 6: Do we want to use the suffix `set` for setters? @Chemaclass @jenshaase

src/php/Config/PhelConfig.php
```diff
@@ -0,0 +1,143 @@
...
+use JsonSerializable;
+use Phel\Build\BuildConfig;
+use Phel\Command\CommandConfig;
+use Phel\Filesystem\FilesystemConfig;
+use Phel\Formatter\FormatterConfig;
+use Phel\Interop\InteropConfig;
+
+final class PhelConfig implements JsonSerializable
+{
+    private array $srcDirs = ['src/phel'];
```
JesusValera, line 16: Not sure if it makes sense to define the default value here (it was already defined in the `phel-config.php` file, isn't it? 😃

JesusValera (commented):
That's awesome, just a few comments.

Thank you so much!

phel-config.php
Chemaclass: I think it makes sense to have setters here, I think it makes the fluent interface more natural.

src/php/Config/PhelConfig.php
Chemaclass: I think it makes sense to use this as the source of truth for these values. Because the goal (as I understood) is to use this object instead of the raw arrays. 
> TODO: we need to update the docs page: https://phel-lang.org/documentation/configuration/

Chemaclass: Would you like to create a PR to update the [phel docs page](https://phel-lang.org/documentation/configuration/) regarding this topic, @cgrabenstein? That would be merged after creating the release, but it would be nice to have it ready sooner.

Chemaclass (approved):

src/php/Config/PhelConfig.php
cgrabenstein: Sure, will do!

Chemaclass: Awesome! One last thing, can you add a line in the `CHANGELOG.md` about this new feature?

JesusValera (approved):

src/php/Config/PhelConfig.php
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    public function jsonSerialize(): array
+    {
+        return [
+            CommandConfig::SRC_DIRS => $this->getSrcDirs(),
+            CommandConfig::TEST_DIRS => $this->getTestDirs(),
+            CommandConfig::VENDOR_DIR => $this->getVendorDir(),
+            CommandConfig::OUTPUT_DIR => $this->getOutDir(),
+            InteropConfig::EXPORT => [
```
jenshaase, line 133: Please let `PhelExportConfig` also implement the `JsonSerialize` interface. Then this could be simplified to `$this->getExport()`

src/php/Config/PhelConfig.php
```diff
@@ -0,0 +1,143 @@
...
+use Phel\Build\BuildConfig;
+use Phel\Command\CommandConfig;
+use Phel\Filesystem\FilesystemConfig;
+use Phel\Formatter\FormatterConfig;
+use Phel\Interop\InteropConfig;
+
+final class PhelConfig implements JsonSerializable
+{
+    private array $srcDirs = ['src/phel'];
+    private array $testDirs = ['tests/phel'];
```
jenshaase, line 17: I would be nice if you can add type annotations for all arrays, like `list<string>`

jenshaase (changes requested):

src/php/Config/PhelConfig.php
cgrabenstein: Done. It simplifies to `$this->getExport()->jsonSerialize()`. The config reader does not use the `json_encode` method (which would call `jsonSerialize()` on the export config), but calls `jsonSerialize` directly.

src/php/Config/PhelConfig.php
cgrabenstein: Done.

Chemaclass:
You can rebase your branch form master and resolve the easy conflict there, so we can merge the PR :)

Chemaclass (approved):
Awesome PR 👏

PR #573, opened 2023-02-16
PhelConfig not accessible as array

Chemaclass:
### 🤔 Background

After merging https://github.com/phel-lang/phel-lang/pull/572 I wanted to quickly try it with my [phel-snake](https://github.com/Chemaclass/phel-snake) project and I found this error:

Screenshot 2023-02-16 at 16 18 12

>**Note**: My bad, I forgot to add tests for this class, so this was a great opportunity to add some 💡 

### 💡 Goal

Fix the bug 🐛 `Error : Cannot use object of type Phel\Config\PhelConfig as array`

### 🔖 Changes

- When doing the "require file" check if it's an array or a PhelConfig object before trying to access to the `SRC_DIRS `
  - This way we keep total compatibility for both options, you can use the array version or the object version

### DIFF
--- a/src/php/Command/Domain/Finder/ComposerVendorDirectoriesFinder.php
+++ b/src/php/Command/Domain/Finder/ComposerVendorDirectoriesFinder.php
@@ -5,9 +5,12 @@ declare(strict_types=1);
 namespace Phel\Command\Domain\Finder;
 
 use Phel\Command\CommandConfig;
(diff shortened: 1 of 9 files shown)


Chemaclass:
There might be a possible code improvement afterwards, but I will merge this PR now, because master is broken without this.

PR #574, opened 2023-02-16
Inject PhpConfigReader

cgrabenstein:
### 🤔 Background

The `ComoposerVendorDirectoriesFinder` duplicated parts of the `PhpConfigReader` class. To avoid duplication of code, the reader should be injected into the finder.

### 💡 Goal

Reducing duplication improves maintainability and avoids problems down the road.

### 🔖 Changes

- The `PhpConfigReader` was added as dependency in the newly created `CommandDependencyProvider`
- A method was added to the `CommandFacade` that reads a config given its absolute path
- Said method is used in the `ComoposerVendorDirectoriesFinder` to read the configs. The old code has been removed.
- A no longer needed test has been removed.


tests/php/Integration/Command/Domain/DirectoryFinder/ComposerVendorDirectoriesFinderTest.php
```diff
@@ -4,18 +4,16 @@
...
 
+use Gacela\Framework\Gacela;
 use Phel\Command\Domain\Finder\ComposerVendorDirectoriesFinder;
-use Phel\Config\PhelConfigException;
+use Phel\Phel;
 use PHPUnit\Framework\TestCase;
 
 final class ComposerVendorDirectoriesFinderTest extends TestCase
 {
-    public function test_exception_when_wrong_type(): void
```
Chemaclass, line 13: Wouldn't be interesting to test what happen if you return a different type in the `phel-config.php`? I mean, I would expect some exception? Or maybe is just ignoring it? 🤔

cgrabenstein: A `RunetimeException` will be thrown, see https://github.com/gacela-project/gacela/blob/0.11.0/src/Framework/Config/ConfigReader/PhpConfigReader.php#L44

I was unsure whether or not to test this behaviour here, but I think we should not test framework code here. It is part of the `PhpConfigReader`'s conttract that it will throw that exception if it can't handle the config. Maybe you could add a custom exception in gacela and add it to the reader's interface, to make it more explicit.

cgrabenstein: However, another question is how the `ComposerVendorDirectoriesFinder` should handle that exception. What does it mean if a vendor dependency has an invalid config file? Should the whole process terminate? Or should that vendor just be ignored? 

Depending on the answer(s) to that question I can add a try-catch and silently ignore that exception to avoid the whole loop to blow up.

Chemaclass: Good point. I think I like the idea of throwing a custom exception at phel level if the file throw a runtimeException

Chemaclass (approved):

PR #576, opened 2023-02-19
Enable gacela cache by default

Chemaclass:
### 🤔 Background

Gacela has a [file cache system](https://gacela-project.com/docs/bootstrap/#gacela-file-cache) with the resolved class names for the internal php modules, and by enabling we can get some performance optimisations. 

The final user wont notice anything, as they should have the `*.cache` pattern in their `.gitignore` file.

This can be easily disabled, anyway, in any project using gacela by defining in their config key values the `'gacela-cache-enabled'` as `false`; you can find it in `Gacela\Framework\ClassResolver\Cache\GacelaFileCache::KEY_ENABLED`.

```php
# whenever you have define your app config files; eg `phel-config.php` or `phel-config-local.php`
return [
    # Using the raw string-key
    'gacela-cache-enabled' => false,

    # Or using the constant
    \Gacela\Framework\ClassResolver\Cache\GacelaFileCache::KEY_ENABLED => false,
];
```

### 🔖 Changes

- Enable gacela cache file system

### 🖼️ Screenshots

This is an example of how the gacela cache files are generated, so it is not necessary to compute this mapping all the time.

Screenshot 2023-02-19 at 12 55 27

We have benchmark tests to verify the performance improvement. 
See the without and with file cache enabled benchmark tests in [gacela's actions](https://github.com/gacela-project/gacela/actions/runs/4217718193/jobs/7321705420):

Screenshot 2023-02-20 at 10 29 30

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 .idea/
 .phpbench/
 .vscode/
-.gacela/
 data/
 out/
 PhelGenerated/
@@ -14,5 +13,5 @@ vendor/
 .phel-repl-history
 local.phel
 phel-config-local.php
-gacela.php
+gacela*.php
 src/php/Api/Infrastructure/phel/doc.phel
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea/
 .phpbench/
 .vscode/
+.gacela/
 data/
 out/
 PhelGenerated/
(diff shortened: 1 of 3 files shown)


JesusValera (approved):

JesusValera:
Should we update the `CHANGELOG.md` with this feature?

PR #577, opened 2023-02-20
Better notice testing

cgrabenstein:
### 🤔 Background

This PR contains my review suggestions for #575 

### 💡 Goal

The PR aims to improve the testing of php notices. By collecting them instead of throwing an Exception, we stay true to the original implementation where no exceptions are being thrown.

### 🔖 Changes

See diff


Chemaclass (approved):
Nice idea! I will merge it in my branch and fix the CI complains in the other branch :)

Issue #578, opened 2023-02-22
php/apush and php/aset not working as expected


cgrabenstein:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.9.0           |
| PHP version         | 8.1.15           |

#### Summary

`php/apush` and `php/aset` not working as expected.

#### Current behavior

When using `php/apush` or `php/aset` I expect to get the updated array returned. Instead I only get the newly added item. I have no possibility to access the updated array. Or at least I could not find any way to do so.

#### How to reproduce

```
phel:1> (def arr (php/array 1 2 3))
1
phel:2> arr
<PHP-Array [1, 2, 3]>
phel:3> (type arr)
:php/array
phel:4> (php/apush arr 4)
4
phel:5> arr
<PHP-Array [1, 2, 3]>
phel:6> (def new-arr (php/apush arr 4))
1
phel:7> new-arr
4
phel:8> (type new-arr)
:int
```

#### Expected behavior

I would expect that `new-arr` in the example above holds an array with the values 1, 2, 3 and 4.

Chemaclass:
Fixed here https://github.com/phel-lang/phel-lang/pull/579

PR #579, opened 2023-02-23
Fix apush, aset and aunset for global php arrays

Chemaclass:
### 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/578

### 💡 Goal

Fix `php/push`, `php/aset` and `php/aunset` on native php arrays when using global definitions.

### 🔖 Changes

- Allowing using `getDefinitionReference()` that will be used when dealing with native php arrays

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 * Added `--testdox` argument to `phel test` command (#567)
 * Added support for fluid configuration in `phel-config.php` (#494)
 * Enable gacela cache filesystem by default (#576)
+* Fix `php/apush`, `php/aset` and `php/aunset` for global php arrays (#579)
 
 ## 0.9.0 (2023-02-05)
 
--- a/src/php/Compiler/Domain/Analyzer/Ast/GlobalVarNode.php
+++ b/src/php/Compiler/Domain/Analyzer/Ast/GlobalVarNode.php
@@ -41,4 +41,9 @@ final class GlobalVarNode extends AbstractNode
     {
         return $this->meta[Keyword::create('macro')] === true;
     }
+
+    public function useReference(): bool
+    {
+        return $this->getEnv()->useGlobalReference();
+    }
 }
--- a/src/php/Compiler/Domain/Analyzer/Environment/NodeEnvironment.php
+++ b/src/php/Compiler/Domain/Analyzer/Environment/NodeEnvironment.php
@@ -16,6 +16,9 @@ final class NodeEnvironment implements NodeEnvironmentInterface
     /** Def inside of def should not work. This flag help us to keep track of this. */
     private bool $defAllowed = true;
 
+    /** Use Registery::getDefinitionReference() instead of Registry::getDefinition() */
+    private bool $globalReference = false;
+
     /**
      * @param array<int, Symbol> $locals A list of local symbols
      * @param string $context The current context (Expression, Statement or Return)
@@ -126,6 +129,14 @@ final class NodeEnvironment implements NodeEnvironmentInterface
         return $result;
     }
 
+    public function withUseGlobalReference(bool $useGlobalReference): NodeEnvironmentInterface
+    {
+        $result = clone $this;
+        $result->globalReference = $useGlobalReference;
+
+        return $result;
+    }
+
     public function withAddedRecurFrame(RecurFrame $frame): NodeEnvironmentInterface
     {
         $result = clone $this;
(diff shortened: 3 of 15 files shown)


jenshaase:
The generate code is something like this:

```php
(\Phel\Lang\Registry::getInstance()->getDefinition("namespace", "arr"))[] = 4
```

The returned array for the `getDefinition` function is a copy of the array and not a reference.

Don't have a good solution yet.

Chemaclass:
Jap, realised that after creating this integration test (not comited [yet], just for debugging purposes):
```php
--PHEL--
(def arr (php/array 1 2 3))
(def new-arr (php/apush arr 4))
(is new-arr [1 2 3 4])
--PHP--
\Phel\Lang\Registry::getInstance()->addDefinition(
  "user",
  "arr",
  array(1, 2, 3),
  \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
    \Phel\Lang\Keyword::create("start-location"), \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
      \Phel\Lang\Keyword::create("file"), "Call/php-def-push.test",
      \Phel\Lang\Keyword::create("line"), 1,
      \Phel\Lang\Keyword::create("column"), 0
    ),
    \Phel\Lang\Keyword::create("end-location"), \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
      \Phel\Lang\Keyword::create("file"), "Call/php-def-push.test",
      \Phel\Lang\Keyword::create("line"), 1,
      \Phel\Lang\Keyword::create("column"), 27
    )
  )
);
\Phel\Lang\Registry::getInstance()->addDefinition(
  "user",
  "new-arr",
  (\Phel\Lang\Registry::getInstance()->getDefinition("user", "arr"))[] = 4,
  \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
    \Phel\Lang\Keyword::create("start-location"), \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
      \Phel\Lang\Keyword::create("file"), "Call/php-def-push.test",
      \Phel\Lang\Keyword::create("line"), 2,
      \Phel\Lang\Keyword::create("column"), 0
    ),
    \Phel\Lang\Keyword::create("end-location"), \Phel\Lang\TypeFactory::getInstance()->persistentMapFromKVs(
      \Phel\Lang\Keyword::create("file"), "Call/php-def-push.test",
      \Phel\Lang\Keyword::create("line"), 2,
      \Phel\Lang\Keyword::create("column"), 31
    )
  )
);
\Phel\Lang\Registry::getInstance()->getDefinition("user", "new-arr");
```

Would it be possible to define in the `PhpArrayPushEmitter` (or maybe in an earlier stage) that we mutate the copy array and then we set the copy array into the old one (overriding the previous reference)?

jenshaase:
`addDefinition` should not update anything. Definition should only be set once.

However, there is a workaround using `var`. This is even better Phel code because it explicitly says that there is some mutation happening:

```
phel:1> (def arr (var (php/array 1 2 3)))
1
phel:2> (deref arr)
<PHP-Array [1, 2, 3]>
phel:3> (let [a (deref arr)] (php/apush a 4) (set! arr a))
nil
phel:4> (deref arr)
<PHP-Array [1, 2, 3, 4]>
phel:5> 
```

Chemaclass:
What do you think about this ☝🏼 about using `(var)`, @cgrabenstein (related to https://github.com/phel-lang/phel-lang/issues/578)

jenshaase:
Is there a way to have a `getDefinitionReference` method that returns the reference and not the copy?

Chemaclass:
@jenshaase there might be a way with https://www.php.net/manual/en/language.references.return.php 🤔

cgrabenstein:
@Chemaclass using `var` is a reasonable workaround. However _if_ there's an implementation that allows to use `apush` without `var`, I would prefer that. I see that this goes against the way how php handles arrays and will require some shenanigans though. 

@jenshaase it's totally possible. See https://3v4l.org/sYHa8

jenshaase:
I added a fix. @Chemaclass do you want to review it?

jenshaase:
I also fixed it for `php/aset` and `php/aunset`

Chemaclass:
Thanks @jenshaase for your help! I added some unit tests for the Registry logic, and renamed some variables. 
Can you review, approve and merge the PR? I cannot approve it because I am the author of the PR 😝

JesusValera (approved):

PR #603, opened 2023-05-30
Support str on contains?

Chemaclass:
### 🔖 Changes

Support strings on `contains?`, additional to the `str-contains?` function.

> Original idea from  for the idea @danirod

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -541,6 +541,7 @@ Calling the `and` function without arguments returns true."
   "Returns true if key is present in the given collection, otherwise returns false."
   [coll key]
   (cond
+    (php/is_string coll) (str-contains? coll key)
     (php/instanceof coll ContainsInterface) (php/-> coll (contains key))
     (php/is_array coll) (php/array_key_exists key coll)
     (throw (php/new InvalidArgumentException (php/. "cannot call 'contains?' on " (php/get_class coll))))))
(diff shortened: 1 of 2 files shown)


jenshaase (commented):
Mh, the documentation of `contains?` says:

```Returns true if key is present in the given collection, otherwise returns false.```

which is different on what `str_contains` does

```Determine if a string contains a given substring```

Also, we already have `str-contains?`.

**Comment by Chemaclass** [comment]
Jap, I know that we have the `str-contains?`, but what about combining it inside the `contains?`
This way we could deprecate the `str-contains?` in favor of `contains?`. Why? To reduce the number of functions that you need to remember in your head while you're working with phel.

> Extra: Maybe we could do something similar with the other function `contains-value?` 🤔

**Comment by jenshaase** [comment]
Currently `contains?` is more like `contains-key?`. For example, let's say we work on a vector. With `contains?` you can only check if the index is present but not that a value is present:

```
(contains? [:a :b :c] :b)  # false
(contains? [:a :b :c] 2)   # true
```

One could say that the `contains-value?` would be a better fit to support strings. But I also don't think this is a good idea. The idea of `str-contains?` is that it checks for a subsequences. The idea of `contains-value?` is that is only checks for one specific value and not for subsequences. That are two different things.

> To reduce the number of functions that you need to remember in your head while you're working with phel.

The philosophy of phel functions is that they should do only one thing and not multiple things. In my opinion it is better to have more functions that focus on one thing than less functions that do multiple things.

Chemaclass:
Thanks for the clear answer, @jenshaase! I agree with what you said. @danirod this is also for you ☝🏼 😄

Issue #593, opened 2023-04-19
Add phel config file automatic after requiring phel


Chemaclass:
## 📚 Description

Currently, in the [Getting Started Guide](https://phel-lang.org/documentation/getting-started/), you are told to manually create a file `phel-config.php` with some configuration that you would copy-paste from the docs. But this could be automated by adding a post-command hook on composer, which could create that file with all methods configured with the default values.

## 🖥️ Tech hint

https://github.com/gacela-project/gacela/blob/main/composer.json#L74

### IDEA 1

Copy the vendor/phel-lang/phel-lang/phel-config.php into the project root dir

### IDEA 2

Create a simpler template somewhere and copy-paste that template into the project root dir; the different with IDEA 1 it would be that the copy-paste is from a different template and not the config from the actual project when using it for development. But I think IDEA 1 would be simpler 🤔

PR #623, opened 2023-10-21
Copy config after install

Chemaclass:
### 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/593

### 💡 Goal

The goal is to copy automatically the phel-config.php from phel-lang to the project where it's been required.
However, I am not sure if this is possible like this, because I think the `post-package-install` (as well as other [composer events](https://getcomposer.org/doc/articles/scripts.md#event-names)) are linked to the project itself, and not projects requiring other projects.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -51,7 +51,9 @@
     },
     "scripts": {
         "post-install-cmd": [
-            "tools/git-hooks/init.sh",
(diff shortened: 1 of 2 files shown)


Chemaclass:
Not doable"

PR #625, opened 2023-10-21
Enable creating an executable phar

Chemaclass:
### 🤔 Background

Currently, phel can be used only as a dependency using composer. However, it would be better to offer the possibility to download and use phel as a unique executable file.

### 💡 Goal

Allow downloading phel as one executable file with phar.

### 🔖 Changes

Add settings to the project to create a phar using [box-project](https://box-project.github.io/box/).


JesusValera (commented):
We should add this to the documentation 😜

drupol:
I guess the latests changes will be helpful in this PR.

PR #630, opened 2023-10-22
Notify user when running non existing file

### LINKED ISSUE #628: Notify user when running a non existing phel file
## Current

If you use `phel run non-existing-file.phel`, there is no feedback from phel. Eg:

Screenshot 2023-10-22 at 12 57 40

## Expected

Show a message notifying the user that the file does not exists, instead of giving no feedback at all.

JesusValera:
### 📚 Description

> Closes #628 

Add a message when the argument (filepath or ns) on `phel run {file.phel}` does not exist.

Screenshot 2023-11-01 at 10 39 16


tests/php/Integration/Run/Command/Run/RunCommandTest.php
```diff
@@ -25,12 +25,12 @@ public static function setUpBeforeClass(): void
...
      */
-    public function test_run_by_namespace(): void
+    public function test_file_not_found(): void
     {
-        $this->expectOutputRegex('~hello world~');
+        $this->expectOutputString('File "non-existing-file.phel" not found' . PHP_EOL);
 
         $this->createRunCommand()->run(
-            $this->stubInput('test\\test-script'),
+            $this->stubInput('non-existing-file.phel'),
```
Chemaclass, line 33: Why not keeping both tests? Did you override it intentionally? 🤔

JesusValera: @Chemaclass the old one doesn't make sense anymore, so I removed it and created the other one.

I mentioned the new logic and also this test in the PR description over the `investigation` part.

Chemaclass (approved):

PR #631, opened 2023-10-26
chore: add `composer.lock`

drupol:
Hi,

This PR was created after a discussion with @Chemaclass.

Basically, since `phel` should be considered as compiler, it should therefore be reproducible.
Without `composer.lock`, it is impossible to ensure reproducibility of the package.

This PR fixes this. 

Created using PHP 8.0.30


Chemaclass (approved):
Thanks!

PR #644, opened 2023-12-21
Require min php version 8.2

Chemaclass:
### 🤔 Background

PHP 8.0 is not supported anymore. Latest PHP version is 8.3, and for what respect to Phel, we should aim to use the latest or close to the latest PHP version to run. I don't see a reason why not.

https://www.php.net/supported-versions.php
Screenshot 2023-12-21 at 22 47 01

### 💡 Goal

Require a more modern PHP version.

### 🔖 Changes

- Require min PHP 8.2 to run Phel

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [ '8.0' ]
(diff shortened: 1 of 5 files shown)


JesusValera (commented):
Changelog is missing?

JesusValera (approved):

Issue #464, opened 2022-04-16
Ignore the MixedReturnStatement errors from psalm level 1


Chemaclass:
Currently, if we set psalm level 1 we get lot of these errors:
```php
ERROR: MixedReturnStatement - src/php/Run/RunFactory.php:97:16 - Could not infer a return type
return $this->getProvidedDependency(RunDependencyProvider::FACADE_COMMAND);
```

The reason is because `getProvidedDependency()` returns a `mixed`, and this is OK (and code from Gacela).
```php
/**
 * @return mixed
 */
protected function getProvidedDependency(string $key)
{
    return $this->getContainer()->get($key);
}
```

The solution for this could be defining a rule somehow in the `psalm.xml` config file to ignore these errors.

> Interesting link: https://psalm.dev/docs/running_psalm/dealing_with_code_issues/

jenshaase:
Alternative solution could be to set the type:

```php
private function getCompilerFacade(): CompilerFacadeInterface
{
  /** @var CompilerFacadeInterface $result */
  $result = $this->getProvidedDependency(RunDependencyProvider::FACADE_COMPILER);
  return $result;
}
```

PR #661, opened 2024-02-13
Increase Psalm level 1

JesusValera:
### 📚 Description

Closes https://github.com/phel-lang/phel-lang/issues/464

Increase Psalm level 1 by adding to ignore some checkers.
All of them are `Mixed...`, that is, we depend on third-party tools like `Symfony` or `Gacela`, as a result, we need to deal with the type of var that was defined originally as `mixed`.

I suggest ignoring all of them as they only will make the code polluter with more lines and unnecessary information most of them.

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 <psalm
-    errorLevel="2"
+    errorLevel="1"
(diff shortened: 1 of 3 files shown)


Chemaclass (approved):
I like this idea. Thanks!

Issue #670, opened 2024-03-16
phel build broken when running with cache


Chemaclass:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.12           |
| PHP version         | 8.2           |

#### Summary

The phel build with cache (by default) fails for some strange reason. 

> **Note**:This bug was found actually by @bentxt here https://github.com/phel-lang/phel-lang/issues/668

⚡️Temporal solution to avoid this bug: use `phel build --no-cache`⚡️

#### How to reproduce

Using https://github.com/phel-lang/web-skeleton
When running `phel build` the first time, it works; it generates the `out/` directory

![Screenshot 2024-03-16 at 15 12 00](https://github.com/phel-lang/phel-lang/assets/5256287/35e779a3-200a-4ede-9b85-87a65d1d7a54)

However, when running it by second time, it outputs this error:

![Screenshot 2024-03-16 at 15 12 38](https://github.com/phel-lang/phel-lang/assets/5256287/61e880ac-c558-46e4-b4bf-643573d16b90)

#### Expected behavior

Running phel build should never not fail

Chemaclass:
@JesusValera @jenshaase I guess this might be because by default the cache is enabled, and in case the file was already built it does a `require_once $targetFile;` to avoid compile the file again. However, for some reason, the PHP might be triggering something and making the build fail... 🤔 

![Screenshot 2024-03-16 at 15 17 25](https://github.com/phel-lang/phel-lang/assets/5256287/d09dba42-ac0b-4e12-bb00-c49eab4b06ef)

Chemaclass:
I think the issue is coming from the web_skeleton, after building the it generates the `app.php` and in the second build then it `require_once $targetFile;` and that's what's creating that output
![Screenshot 2024-04-16 at 18 15 57](https://github.com/phel-lang/phel-lang/assets/5256287/194839e7-fdd5-430f-9887-948106b347c4)

![Screenshot 2024-04-16 at 18 14 24](https://github.com/phel-lang/phel-lang/assets/5256287/474475f8-6cdd-43f6-b795-b5532527f83d)

I think we could add a configuration list to set "NoCacheWhenBuilding" files/paths form the `phel-config.php`. This way, the user is responsible to disable the cache for the entry point of the apps, and this way fix this issue. That's my assumption so far.

PR #671, opened 2024-03-16
Use ArrayPrinter when printerName == array

Chemaclass:
### 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/670

### 💡 Goal

Avoid crash when running `phel build` with cache

### 🖼️ Screenshots

Example https://github.com/phel-lang/web-skeleton using latest phel:0.12 running `phel build` twice

Screenshot 2024-03-16 at 15 38 36

I am still not able to avoid/hide the message `cannot determine HTTP Method` from the output console. 
Any ideas, @jenshaase ?

### DIFF
--- a/src/php/Printer/Printer.php
+++ b/src/php/Printer/Printer.php
@@ -145,7 +145,7 @@ final readonly class Printer implements PrinterInterface
             return new NullPrinter($this->withColor);
         }
 
-        if ($printerName === 'array' && !$this->readable) {
+        if ($printerName === 'array') {
             return new ArrayPrinter($this, $this->withColor);
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
   * in favor of `->setMainPhpFilename()`
 * Add `phel fmt` alias for format
 * Add support for numeric on `empty?`
+* Fix `Printer` for arrays
 
 ## [0.12.0](https://github.com/phel-lang/phel-lang/compare/v0.11.0...v0.12.0) - 2023-11-01
 


src/php/Printer/Printer.php
```diff
@@ -145,7 +145,7 @@ private function creatScalarTypePrinter(mixed $form): TypePrinterInterface
             return new NullPrinter($this->withColor);
         }
 
-        if ($printerName === 'array' && !$this->readable) {
```
JesusValera, line 148: Any idea about the reason for adding this `!$this->readable` in the first place?

JesusValera (approved):

src/php/Printer/Printer.php
Chemaclass: Honestly, I don't remember. Maybe Jens?

Chemaclass: Honestly, I don't remember. Maybe @jenshaase?

jenshaase: @JesusValera @Chemaclass We use the Printer for two purposes:

1. For the Phel function `(print-str ...)` to print a value
2. In the compiler to emit valid Phel code (`OutputEmitter`)

For the second case we use the `readable` flag to only output valid Phel code. In this case a PHP array cannot be converted to Phel code and therefore is ignored. But for the `(print-str ...)` we want to print the array to show it to the user.

Chemaclass: @jenshaase as follow up question, is this PR-changes helping (removing the `&& !$this->readable`) or should we fix the original issue somewhere else and keep that flag check here? 
> In such a case, where should we look at? 👀

Chemaclass: @jenshaase as follow up question, is this PR-changes helping (removing the `&& !$this->readable`) or should we fix the original issue somewhere else and keep that flag check here?
In such a case, where should we look at? 👀

jenshaase: The error is probably somewhere else. Looks like the build wants to emit some PHP array, which it shouldn't. You probably have then code that is not executable.

Chemaclass: You are right, @jenshaase. I think the issue is coming from the web_skeleton, after building the it generates the `app.php` and in the second build then it `require_once $targetFile;` and that's what's creating that output
Screenshot 2024-04-16 at 18 15 57

Screenshot 2024-04-16 at 18 14 24

I think we could add a configuration list to set "NoCacheWhenBuilding" files/paths form the `phel-config.php`. This way, the user is responsible to disable the cache for the entry point of the apps, and this way fix this issue. That's my assumption so far.

Chemaclass:
Closing in favor of other approach/solution https://github.com/phel-lang/phel-lang/pull/671#discussion_r1567639735

PR #688, opened 2024-04-19
feat: add string functions.

smeghead:
Add the namespace phel\str and add the string function below.

split join reverse upper-case replace replace-first trim-newline capitalize lower-case upper-case trim triml trimr blank? starts-with?  ends-with?  includes?  re-quote-replacement escape index-of last-index-of split-lines

### 🤔 Background

I created a string manipulation function library that is not available in phel, using clojure as a reference.
https://github.com/smeghead/phel-str
This is a pull request to introduce the functions of this library to phel.

### 💡 Goal

You can now use the added functions.

```
phel:1> (require phel\str)
phel\str
phel:2> (str/join ["a" "b" "c"])
abc
phel:3> (str/upper-case "hello!")
HELLO!
```

### 🔖 Changes

added `src/phel/str.phel` file includes functions. and added namespace `phel\str`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,32 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+* Add `phel/str` functions
+  * `split`: Splits string on a regular expression
+  * `join`: Returns a string of all elements in coll
+  * `reverse`: Returns s with its characters reversed
+  * `upper-case`: Converts string to all upper-case
+  * `replace`: Replaces all instance of match with replacement in string
+  * `replace-first`: Replaces the first instance of match with replacement in string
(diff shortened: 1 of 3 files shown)


Chemaclass (commented):
Wow, this is awesome. Thank you!! 🙏 

Do you mind adding something like this to the `CHANGELOG.md`?
```md

## Unreleased

* Add `phel/str` functions
  * `split`: Splits string on a regular expression
  * `join`: Returns a string of all elements in coll
  * ... and the others... 
; one simple line with the tl;dr of the func would be enough
```

JesusValera (approved):
👏🏼👏🏼👏🏼

Chemaclass (approved):

PR #689, opened 2024-04-20
Rename compiler to transpiler

Chemaclass:
#### Compiler

A compiler is a program that translates code written in one programming language (the source language) into another language (the target language), typically a lower-level language like machine code or bytecode.

- Compilers are commonly used in languages like C, C++, and Java. They take the entire source code as input, analyze it, and then generate the equivalent executable code or bytecode.
- The output of a compiler is usually a standalone executable file or bytecode that can be run directly by the computer's processor or a virtual machine.
- Compilers often perform optimizations during the translation process to improve the efficiency and performance of the generated code.

#### Transpiler (Source-to-Source Compiler)

A transpiler, short for "source-to-source compiler," translates code from one programming language to another programming language at the same level of abstraction.

- Unlike compilers, which typically translate high-level code to low-level code, transpilers translate code from one high-level language to another high-level language.
- Transpilers are commonly used in web development, especially with languages like TypeScript, CoffeeScript, and JSX, which are transpiled to JavaScript.
- The purpose of transpilers is often to enable developers to write code in a more expressive or convenient language while still targeting a specific platform or runtime environment.

> Source: ChatGPT

According to these definitions, Phel is similar as TypeScript to JavaScript; It is a transpiler from Lisp to PHP.

### 💡 Goal

Get rid of the compiler wording use in favor of transpiler 

### 🔖 Changes

- Use transpiler instead of compiler concepts


Chemaclass:
@jenshaase @smeghead @JesusValera what do you think about this idea?

JesusValera (approved):
Yeah, I think that change makes sense, let's see what other people think :)

jenshaase:
I don't like the word transpiler at all. If you compare the definitions of both terms you will see that the only difference is the "level" of output. The output is only a small part of the program. Other parts of the program involve tokenization, parsing, ast transformation/optimization, etc.. These parts are the bigger and more interesting parts of the program. The need to be done in both compilers and transpilers. So in general there is not much of a difference. Therefore, I would stick to the classic term: compiler.

For me it would be ok to say that phel is a compiler program that transpiles Phel code to PHP Code. So the actual program is a compiler and the more high level view is a transpilation.

Chemaclass:
@jenshaase I didn't think about that, and I like how you put it "_that phel is a compiler program that transpiles Phel code to PHP Code. So the actual program is a compiler and the more high level view is a transpilation._", in such a case, I will close this PR - no merge needed and all good 👍🏼

PR #697, opened 2024-05-16
Fix setMainPhpPath() when not dir is not defined

Chemaclass:
### 🤔 Background

When using `setMainPhpPath()` if there is no directory defined, there is an internal error: `Undefined array key 1`

Screenshot 2024-05-16 at 17 00 17

### 🔖 Changes

Use `out` as default directory in case it's not defined in `PhelBuildConfig->setMainPhpPath()`

### DIFF
--- a/src/php/Config/PhelBuildConfig.php
+++ b/src/php/Config/PhelBuildConfig.php
@@ -6,6 +6,8 @@ namespace Phel\Config;
 
 use JsonSerializable;
 
+use function count;
+
 /**
  * @psalm-suppress DeprecatedProperty
  */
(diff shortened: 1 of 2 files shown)


JesusValera (approved):

src/php/Config/PhelBuildConfig.php
```diff
@@ -84,7 +86,12 @@ public function setMainPhpPath(string $path): self
     public function getMainPhpPath(): string
     {
         if ($this->mainPhpPath !== '') {
-            return $this->mainPhpPath;
+            $explode = explode('/', $this->mainPhpPath);
+            if (count($explode) === 2) {
+                return $this->mainPhpPath;
+            }
+
+            return sprintf('out/%s', $this->getPhpFilename());
```
smeghead, line 94: `out` is specified as a literal. Isn't it `$this->getDestDir()`?
If that were the case, I think line 94 would be unnecessary since it would be the same as line 97.

smeghead (commented):
I have a question about the `out` literal in `src/php/Config/PhelBuildConfig.php`.

src/php/Config/PhelBuildConfig.php
Chemaclass: you are right, good catch! This line is not needed

smeghead (approved):

Issue #704, opened 2024-05-27
Improve error handling for missing mandatory parameters


kambo-1st:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.14.1          |
| PHP version         | 8.2.20          |

#### Summary

This is maybe more of a feature request as I actually don't know if and how Phel guarantees the check number of mandatory parameters or if it leaves it to PHP by design. My feeling is that at this moment it seems a little bit too low level. 

#### Current behavior

For example for this piece of code (I forgot to use `reduce2` instead of `reduce`):

```phel
(def array1 [1 2 3])
(def array2 [4 5 6])
(reduce + (map * array1 array2))
```

I received the following exception:

```
Phel\Compiler\Domain\Evaluator\Exceptions\CompiledCodeIsMalformedException: Too few arguments to function Phel\Lang\AbstractFn@anonymous::__invoke(), 2 passed in /tmp/__phelnmQ4Gi on line 4 and exactly 3 expected
in /home/kambo/workspace/cli-skeleton/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php:14 (gen: /home/kambo/workspace/cli-skeleton/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php:14)

#0 /home/kambo/workspace/cli-skeleton/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Evaluator/RequireEvaluator.php(43): Phel\Compiler\Domain\Evaluator\Exceptions\CompiledCodeIsMalformedException::fromThrowable(Object(ArgumentCountError))
#1 /home/kambo/workspace/cli-skeleton/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Compiler/EvalCompiler.php(110): Phel\Compiler\Domain\Evaluator\RequireEvaluator->eval('// string
// ;;...')
#2 /home/kambo/workspace/cli-skeleton/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Compiler/EvalCompiler.php(69): Phel\Compiler\Domain\Compiler\EvalCompiler->evalNode(Object(Phel\Compiler\Domain\Analyzer\Ast\CallNode), Object(Phel\Compiler\Infrastructure\CompileOptions))
[...]

PR #710, opened 2024-06-01
Improve error handling for missing mandatory parameters

Chemaclass:
### 🤔 Background

Closes: https://github.com/phel-lang/phel-lang/issues/704

This is maybe more of a feature request as I actually don't know if and how Phel guarantees the check number of mandatory parameters or if it leaves it to PHP by design. My feeling is that at this moment it seems a little bit too low level.

### 💡 Goal

Improve exception error when missing params

### 🔖 Changes

Normalize the error msg on `CompiledCodeIsMalformedException` on the repl and run commands 

### BEFORE

Screenshot 2024-06-03 at 17 08 59

### AFTER

Screenshot 2024-06-03 at 17 08 48

### DIFF
--- a/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
+++ b/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
@@ -17,9 +17,14 @@ final class CompiledCodeIsMalformedException extends RuntimeException
 
     private static function normalize(string $msg): string
     {
-        $pattern = '/Too few arguments to function [^,]+, (\d+) passed in [^,]+ and exactly (\d+) expected/';
+        $pattern = '/Too few arguments to function [^,]+, (\d+) passed in [^,]+ on line (\d+) and exactly (\d+) expected/';
         if (preg_match($pattern, $msg, $matches)) {
-            return sprintf('Too few arguments to function, %s passed in and exactly %s expected', $matches[1], $matches[2]);
+            return sprintf(
+                'Too few arguments to function on line %s, %s passed in and exactly %s expected',
+                $matches[2],
+                $matches[1],
+                $matches[3],
+            );
         }
 
         return 'Error message not found';
--- a/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
+++ b/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
@@ -7,6 +7,7 @@ namespace Phel\Compiler\Domain\Evaluator\Exceptions;
 use Phel\Compiler\Domain\Analyzer\Ast\AbstractNode;
 use Phel\Compiler\Domain\Analyzer\Ast\Fnable;
 use Phel\Compiler\Domain\Analyzer\Ast\LocalVarNode;
+use Phel\Lang\SourceLocation;
 use RuntimeException;
 use Throwable;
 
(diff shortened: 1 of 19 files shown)


src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
```diff
@@ -11,6 +11,23 @@ final class CompiledCodeIsMalformedException extends RuntimeException
...
+        if (preg_match($pattern, $msg, $matches)) {
+            return sprintf(
+                'Too few arguments to function on line %s, %s passed in and exactly %s expected',
+                $matches[2],
+                $matches[1],
+                $matches[3],
+            );
+        }
+
+        return $msg;
```
Chemaclass, line 31: I think we could normalize more use cases here. I see there is also other potential error msg like `CompiledCodeIsMalformedException: cannot do cons 1` so we could use another regex for this, and others alike. 
What do you think about this posible solution?

src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
```diff
@@ -4,13 +4,48 @@
...
+        }
+
+        return new self($msg, 0, $e);
+    }
+
+    private static function normalize(string $msg, Fnable $node): string
+    {
+        $srcLoc = $node->getStartSourceLocation();
+
+        $pattern = '/Too few arguments to function.*, (?<passed>\d+) passed in .* and exactly (?<expected>\d+) expected/';
```
smeghead, line 31: The `CompiledCodeIsMalformedException::normalize` method converts the message that starts with `Too few arguments to function` into a user-friendly message.

First, which class throws an exception with a message that starts with `Too few arguments to function`?

I searched but couldn't find it.

smeghead (commented):
For my understanding, I have a question.

The `CompiledCodeIsMalformedException::normalize` method converts the message that starts with `Too few arguments to function` into a user-friendly message.

First, which class throws an exception with a message that starts with `Too few arguments to function`?

I searched but couldn't find it.

src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php
Chemaclass: Look at https://github.com/phel-lang/phel-lang/blob/main/src%2Fphp%2FCompiler%2FDomain%2FEvaluator%2FRequireEvaluator.php lime 37, while doing a require pho code, if that code is malformed, then PHP throws an exception and it's being captured and transformed into our custom code malformed exception

smeghead: *PHP* throw. I understood.
thank you

smeghead (approved):

jenshaase:
I haven't tested this yet. But it will probably not find this error:


```
(defn test [] (reduce + (map * arr1 aar2)))
```

Then it seems quite useless for me. In real code most function calls are in defn and not on the top.

Chemaclass:
@jenshaase ach, you're right. Right now the current changes are not working with your example... 

Screenshot 2024-06-03 at 15 28 57

however, it might be something doable updating the current changes mmm

Chemaclass:
I will work on adding a bit more context to the error message to check if there is a file, render the temp-filename and also display the exact PHP line that triggered that error, in which you will be able to see the actual function 🏗️

Chemaclass:
I talked to @jenshaase and we agreed to try to solve this from another angle. Instead of rendering the error on runtime (what I did; catching the PHP runtime error), we should try to do it on compilation time. Therefore, I will close the current PR and we will work on another PR in the coming days.

PR #711, opened 2024-06-03
Remove unnecessary condition in ReplCommandTest

Chemaclass:
### 🔖 Changes

Remove unnecessary condition inside `ReplCommandTest::buildDataProviderFromDirectory()`. 
That same condition is duplicated right above.


JesusValera (approved):
It seems Rector added this one, if CI passes, we can get rid of it 👍🏼

PR #716, opened 2024-06-07
fix: deprecation notice about signed binary number in REPL

geoffreyvanwyk:
### 🤔 Background

When evaluating a negative binary number in the REPL:

> phel:1> -0b1

> [2024-06-07T10:58:02+00:00] Error Unknown(22527) found!
file: phel-lang/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php:129
context: {"errno":8192}

It is also raised for a positively-signed binary number.

Phel evaluates binary numbers using the [bindec()](https://php.net/bindec) function, which only accepts unsigned values; therefore, `bindec()` raises a deprecation notice in the REPL.

### 💡 Goal

### 🔖 Changes

- `bindec()` also does not accept the underscores in numbers. For this reason, Phel removes them before passing the binary number to `bindec()` in the `AtomParser` class. This commit modifies the regular expression for binary numbers to capture the unsigned value of the number, then passes that to `bindec()`.
- It includes in `AtomParserTest::test_parse_binary_number()` assertions for signed binary numbers.
- It configures PHPUnit to convert deprecation notices into exceptions so that the new assertions can be useful.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -7,6 +7,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
+         convertDeprecationsToExceptions="true"
          defaultTestSuite="unit"
          executionOrder="random"
          processIsolation="false"
--- a/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
+++ b/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php
@@ -20,7 +20,7 @@ final readonly class AtomParser
 {
     private const REGEX_KEYWORD = '/:(?<second_colon>:?)((?<namespace>[^\/]+)\/)?(?<keyword>[^\/]+)/';
 
-    private const REGEX_BINARY_NUMBER = '/^([+-])?0[bB][01]+(_[01]+)*$/';
+    private const REGEX_BINARY_NUMBER = '/^([+-])?0[bB]([01]+(?:_[01]+)*)$/';
 
     private const REGEX_HEXADECIMAL_NUMBER = '/^([+-])?0[xX][0-9a-fA-F]+(_[0-9a-fA-F]+)*$/';
 
@@ -121,12 +121,13 @@ final readonly class AtomParser
     private function parseBinaryNumber(array $matches, string $word, Token $token): NumberNode
     {
         $sign = (isset($matches[1]) && $matches[1] === '-') ? -1 : 1;
+        $unsignedInteger = $matches[2] ?? $word;
 
         return new NumberNode(
             $word,
             $token->getStartLocation(),
             $token->getEndLocation(),
-            $sign * bindec(str_replace('_', '', $word)),
+            $sign * bindec(str_replace('_', '', $unsignedInteger)),
         );
     }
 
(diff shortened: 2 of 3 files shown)


geoffreyvanwyk:
**Before**

image

**After**

image

Chemaclass (commented):
The fix seems to be right. Can you please add a unit test for this bug as well? For example:
```php
# tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php:185
public function test_parse_negative_binary_number(): void
{
    $parser = new AtomParser(new GlobalEnvironment());
    $start = new SourceLocation('string', 0, 0);
    $end = new SourceLocation('string', 0, 5);

    self::assertEquals(
        new NumberNode('-0b101', $start, $end, -5),
        $parser->parse(new Token(Token::T_ATOM, '-0b101', $start, $end)),
    );
}
```

smeghead (commented):
Just before that, the regular expression `self::REGEX_BINARY_NUMBER` is used to determine whether it is a binary number.

Rather than removing the `-` with `str_replace`, wouldn't it be clearer to capture the regular expression and get a string to pass as an argument to `bindec`?

For example, by changing the regular expression `self::REGEX_BINARY_NUMBER` as follows, you can obtain the required string by referencing it with `$matches[2]`.

```php
    private const REGEX_BINARY_NUMBER = '/^([+-])?0[bB]([01]+(?:_[01]+)*)$/';
```

geoffreyvanwyk:
Hi @Chemaclass, initially I added the test, but realized that, unless PHPUnit turned deprecation warnings into exceptions, the test would pass regardless of the fix. I was worried that other tests might fail if PHPUnit turned deprecation warnings into exceptions.

I will add the test now.

Chemaclass:
@geoffreyvanwyk let's add the test for now, yes. And that's a really good point, we might should consider enable in phpunit throwing exceptions for warnings. Maybe there is some configuration flag at phpunit level 🤔

geoffreyvanwyk:
> @geoffreyvanwyk let's add the test for now, yes. And that's a really good point, we might should consider enable in phpunit throwing exceptions for warnings. Maybe there is some configuration flag at phpunit level 🤔

@Chemaclass, it can be added in **phpunit.xml.dist** by adding attribute `convertDeprecationsToExceptions="true"` to `<phpunit>` directly ([link](https://docs.phpunit.de/en/9.6/configuration.html#the-convertdeprecationstoexceptions-attribute)).

But I saw now it does not help this specific test, because the custom error handler catches the deprecation warnings before PHPUnit can see them. I could not find a way for PHPUnit to check the output from the custom error handler.

geoffreyvanwyk:
Hi @smeghead , I have implemented your suggestion. Now the "+" sign is also handled.

geoffreyvanwyk:
@Chemaclass, instead of adding separate test methods, I added two more assertions for "+"-signed and "-"-signed binary numbers to `AtomParserTest::test_parse_binary_number()`.

I checked the `convertDeprecationsToExceptions` PHPUnit attribute again, and saw that it does make a difference. That is what I saw the first time. Not sure why it did not work the second time. I have enabled it in **phpunit.xml.dist**.

geoffreyvanwyk:
> Have you check if the same problem occurs for the different bases? Octal, hexa, etc? It might worth checking and creating another or similar for them in case they are also broken

Yes, it does. I will do so.

Chemaclass:
@geoffreyvanwyk can you please update the `CHANGELOG.md` in this PR as well?

PR #717, opened 2024-06-07
feat: Check mandatory function parameters during compile time.

jenshaase:
### 🤔 Background

Issue: https://github.com/phel-lang/phel-lang/issues/704

### 💡 Goal

Fail on compilation time when passing wrong number of args to a function.

### 🔖 Changes

Added a `'min-arity'` meta data to the `FnNode` and check that when analyzing the `InvokeSymbol`

### 🖼️ Screenshots

Screenshot 2024-06-08 at 17 36 19

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 * Fix add check for readline extension in REPL to handle missing dependencies (#712)
+* Check mandatory function parameters during compile time (#717)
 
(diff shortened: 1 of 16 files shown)


jenshaase:
@Chemaclass Here is the draft MR. I only tested is roughly and some tests needs to be fixed. I added a meta field called 'min-arity'. This is checked during compilation.

jenshaase:
@Chemaclass Thanks for the improvements. Looks nice to me!

Chemaclass (approved):

PR #718, opened 2024-06-08
fix: deprecation notice for signed hexadecimals

geoffreyvanwyk:
### 🤔 Background

When evaluating a negative hexadecimal number in the REPL:

> phel:1> -0x1

> [2024-06-08T13:22:33+00:00] Error Unknown(22527) found!
file: phel-lang/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php:141 
context: {"errno":8192}

It is also raised for positively-signed hexadecimal numbers.

Phel evaluates hexadecimal numbers using the
[hexdec()](https://php.net/hexdec) function, which only accepts unsigned values; therefore, `hexdec()` raises a deprecation notice in the REPL.

### 💡 Goal

### 🔖 Changes

- `hexdec()` also does not accept the underscores in numbers. For this reason, Phel removes them before passing the hexadecimal number to `hexdec()`. This commit modifies the regular expression for hexadecimal numbers to capture the unsigned value of the number, then passes that to `hexdec()`.
- It adds two new tests to `AtomParserTest::class` for parsing positively-signed and negative hexadecimal numbers.
- It configures PHPUnit to convert deprecation notices into exceptions so that the new tests can be useful.

### DIFF
--- a/tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php
+++ b/tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php
@@ -200,6 +200,42 @@ final class AtomParserTest extends TestCase
         );
     }
 
+    public function test_parse_positively_signed_hexadecimal_number(): void
+    {
+        $parser = new AtomParser(new GlobalEnvironment());
+        $start = new SourceLocation('string', 0, 0);
+        $end = new SourceLocation('string', 0, 5);
+        $this->assertEquals(
+            new NumberNode(
+                '+0x001',
+                $start,
+                $end,
+                1,
+            ),
+            $parser->parse(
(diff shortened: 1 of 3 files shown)


geoffreyvanwyk:
**Before**

image

**After**

image

tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php
```diff
@@ -200,6 +200,42 @@ public function test_parse_hexadecimal_number(): void
         );
     }
 
+    public function test_parse_positively_signed_hexadecimal_number(): void
```
Chemaclass, line 203:
```suggestion
    public function test_parse_positive_signed_hexadecimal_number(): void
```

Chemaclass (approved):

PR #719, opened 2024-06-08
fix: deprecation notice for signed octals

geoffreyvanwyk:
### 🤔 Background

When evaluating a negative octal number in the REPL:

> phel:1> -01

> [2024-06-08T13:51:08+00:00] Error Unknown(22527) found!
file: phel-lang/src/php/Compiler/Domain/Parser/ExpressionParser/AtomParser.php:153 
context: {"errno":8192}

It is also raised for positively-signed octal numbers.

Phel evaluates octal numbers using the [octdec()](https://php.net/octdec) function, which only accepts unsigned values; therefore, `octdec()` raises a deprecation notice in the REPL.

### 💡 Goal

### 🔖 Changes

- `octdec()` also does not accept the underscores in numbers. For this reason, Phel removes them before passing the octal number to `octdec()`. This commit modifies the regular expression for octal numbers to capture the unsigned value of the number, then passes that to `octdec()`.
- It adds two new tests to AtomParserTest::class for parsing positively-signed and negative octal numbers.
- It configures PHPUnit to convert deprecation notices into exceptions so that the new tests can be useful.

### DIFF
--- a/tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php
+++ b/tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php
@@ -236,6 +236,42 @@ final class AtomParserTest extends TestCase
         );
     }
 
+    public function test_parse_positively_signed_octal_number(): void
+    {
+        $parser = new AtomParser(new GlobalEnvironment());
+        $start = new SourceLocation('string', 0, 0);
+        $end = new SourceLocation('string', 0, 5);
+        $this->assertEquals(
+            new NumberNode(
+                '+01',
+                $start,
+                $end,
+                1,
+            ),
+            $parser->parse(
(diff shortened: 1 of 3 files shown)


geoffreyvanwyk:
**Before**

image

**After**

image

tests/php/Unit/Compiler/Parser/ExpressionParser/AtomParserTest.php
```diff
@@ -236,6 +236,42 @@ public function test_parse_octal_number(): void
         );
     }
 
+    public function test_parse_positively_signed_octal_number(): void
```
Chemaclass, line 239:
```suggestion
    public function test_parse_positive_signed_octal_number(): void
```

Chemaclass (approved):

PR #720, opened 2024-06-08
Improve output for doc command

Chemaclass:
### 💡 Goal

Improve the output for the doc command

### 🔖 Changes

- Display "doc" command as table

#### BEFORE

Screenshot 2024-06-09 at 14 00 40

#### AFTER

Screenshot 2024-06-09 at 14 00 18

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file.
 * Fix deprecation notice for signed hexadecimals (#718)
 * Fix deprecation notice for signed octals (#719)
 * Check mandatory function parameters during compile time (#717)
+* Improve output for doc command (#720)
 
 ## [0.14.1](https://github.com/phel-lang/phel-lang/compare/v0.14.0...v0.14.1) - 2024-05-24
 
--- a/src/php/Api/Infrastructure/Command/DocCommand.php
+++ b/src/php/Api/Infrastructure/Command/DocCommand.php
@@ -15,7 +15,6 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 use function in_array;
-use function strlen;
 
 /**
  * @method ApiFacade getFacade()
@@ -31,7 +30,6 @@ final class DocCommand extends Command
         $this->setName('doc')
             ->setDescription('Display the docs for any/all phel functions')
             ->addArgument('search', InputArgument::OPTIONAL, 'Search input that look for a similar function name', '')
-            ->addOption('table', 't', InputOption::VALUE_NONE, 'Display a table')
             ->addOption(
                 self::OPTION_NAMESPACES,
                 null,
@@ -47,11 +45,7 @@ final class DocCommand extends Command
         $phelFunctions = $this->getFacade()->getPhelFunctions($namespaces);
 
         $search = $input->getArgument('search');
-        if ($input->getOption('table')) {
-            $this->printFunctionsAsTable($output, $phelFunctions, $search);
-        } else {
-            $this->printFunctionsAsText($output, $phelFunctions, $search);
-        }
+        $this->printFunctionsAsTable($output, $phelFunctions, $search);
 
         return self::SUCCESS;
     }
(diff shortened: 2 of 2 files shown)


JesusValera (commented):
I love it, my only concern is why we still using the old default printer as "default", and for table mode, you need to specify it explicitly.

I would even drop the old one and use the table as the default and only mode. What do you think @smeghead @Chemaclass? 😀

JesusValera (approved):

smeghead:
@Chemaclass @JesusValera 

It seems that the display process of the table format depends on the 'tput' command.

my docker environment has no `tput`, so doc command cause error.

```bash
6d3ff737a5cf:/usr/src# bin/phel doc id 
sh: tput: not found

Fatal error: Uncaught ValueError: str_repeat(): Argument #2 ($times) must be greater than or equal to 0 in /usr/src/vendor/symfony/console/Helper/Table.php:489
Stack trace:
#0 /usr/src/vendor/symfony/console/Helper/Table.php(489): str_repeat('-', -3)
#1 /usr/src/vendor/symfony/console/Helper/Table.php(421): Symfony\Component\Console\Helper\Table->renderRowSeparator(0, NULL, NULL)
#2 /usr/src/src/php/Api/Infrastructure/Command/DocCommand.php(89): Symfony\Component\Console\Helper\Table->render()
#3 /usr/src/src/php/Api/Infrastructure/Command/DocCommand.php(48): Phel\Api\Infrastructure\Command\DocCommand->printFunctionsAsTable(Object(Symfony\Component\Console\Output\ConsoleOutput), Array, 'id')
#4 /usr/src/vendor/symfony/console/Command/Command.php(279): Phel\Api\Infrastructure\Command\DocCommand->execute(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#5 /usr/src/vendor/symfony/console/Application.php(1029): Symfony\Component\Console\Command\Command->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#6 /usr/src/vendor/symfony/console/Application.php(316): Symfony\Component\Console\Application->doRunCommand(Object(Phel\Api\Infrastructure\Command\DocCommand), Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#7 /usr/src/vendor/symfony/console/Application.php(167): Symfony\Component\Console\Application->doRun(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#8 /usr/src/src/php/Console/Infrastructure/ConsoleBootstrap.php(24): Symfony\Component\Console\Application->run(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
#9 /usr/src/bin/phel(45): Phel\Console\Infrastructure\ConsoleBootstrap->run()
#10 /usr/src/bin/phel(46): {closure}()
#11 {main}
  thrown in /usr/src/vendor/symfony/console/Helper/Table.php on line 489

```

The table format display is very easy to see, so I agree with the default table display.

Is it possible to achieve the equivalent functionality without using 'tput', or to use the traditional display method in an environment where there is no 'tput' command?

Chemaclass:
oh, very interesting catch. The reason this was there is to get the total columns to be able to display the table. Otherwise, we need to have something (reasonable) fixed, or the table will be displayed really ugly... @smeghead, do you mind trying out finding an alternative for the tput command?

Chemaclass:
@smeghead I think this might do the trick. Can you try it out in your local/docker and try to reproduce the error?
```php
$colCount = (int)getenv('COLUMNS');
```
I found that there are these two global envs that we could use instead - in our case, `COLUMNS`:
Screenshot 2024-06-09 at 14 42 36

Chemaclass:
@smeghead yeah, they are using the same code there
```php
# Symfony\Component\Console\Terminal
public function getWidth(): int
{
    $width = getenv('COLUMNS');
    if (false !== $width) {
        return (int) trim($width);
    }

    if (null === self::$width) {
        self::initDimensions();
    }

    return self::$width ?: 80;
}
```

Actually, I prefer using that one, which is taking some more checks in consideration. Thanks for finding this out!

PR #721, opened 2024-06-08
Introduce application layer

Chemaclass:
### 🤔 Background

Currently, on these modules we have only Domain and Infrastructure, and the Domain folder contains most of the code (mixing the Application and Domain layers responsibilities).

- The Infrastructure is the outside of your business/domain logic. 
  - it connects to the I/O, to the Framework, and it talks to the Application layer
  - it is where the Adapters(concrete implements of the Ports) are
- The Application layer contains the use cases
  - it uses the Domain clases
  - it is the intermediary between the infrastructure and the domain
- The Domain is the core logic of the module/domain/business
  - it is where the Ports(interfaces) are
  - it should never talk to the Infrastructure

### 💡 Goal

Implement in intermediate level of abstraction between Infrastructure and Domain named Application.
The goal is to place in the Application layer the classes that are directly used from the Facade, helping the visibility of the responsibilities of the module.

### 🔖 Changes

Move the classes used by the Facade to the Application layer.


smeghead:
As a partial example, I tried to output a class diagram only under 'src/php/Build'.

### Before
Buildier-before

### After
Builder-after
I think it's a more appropriate structure.

smeghead (approved):

JesusValera (approved):

PR #722, opened 2024-06-09
Fix doc command

Chemaclass:
### 🤔 Background

https://github.com/phel-lang/phel-lang/pull/720#issuecomment-2156510943 

The display process of the table format depends on the `tput` command.
So, if you dont have `tput` the `doc` command wont cause an error.

### 🔖 Changes

- Use env `$COLUMNS` instead of `tput`


smeghead:
Even in an environment where there is no 'tput' command, the 'doc' command can be correctly displayed in table format.

image

smeghead (approved):

PR #724, opened 2024-06-11
fix: case expression cannot resolve phel\core/equals1

### LINKED ISSUE #723: `case` expression fails because it cannot resolve symbol `phel\core/equals1`
| Phel version        | v0.14           |
| PHP version         | 8.3.7           |

#### Summary

When attempting to evaluate a `case` expression with an arithmetic test expression and one or more pairs, the following error is raised:

> Cannot resolve symbol 'phel\core/equals1'
> in string:64

#### Current behavior

The expression:

```phel
(case (+ 7 5) 3 :small 12 :big)
```

Raises an error, instead of returning a value.

#### How to reproduce

Enter the above expression into the REPL.

#### Expected behavior

The expression should evaluate to `:big`.

geoffreyvanwyk:
### 🤔 Background

Resolves #723 

### 💡 Goal

Ensure that `case` works with arithmetic expressions.

### 🔖 Changes

- Make the `equals1` function public.
- Add tests for `case` involving arithmetic expressions.


Chemaclass:
Making the function `equals1` public doesn't seems right to me, because `=` is the real function to check for "equality"... however, this bug is a bit strange and making this change it fixes the issue 🤔 What do you think, @jenshaase ?

Screenshot 2024-06-11 at 10 02 54

jenshaase:
As @Chemaclass said equals1 shouldn't be public. I think the real problem here is the macro expansion. The expansion happens in the wrong namespace context. I can have a look at this.

src/phel/core.phel
```diff
@@ -258,7 +258,7 @@ This is PHP equivalent to `$args[0] . $args[1] . $args[2] ...`."
 # Basic sequence operation
 # ------------------------
 
-(defn- equals1 [a b]
+(defn equals1 [a b]
```
Chemaclass, line 261: I would suggest keeping `equals1` private, and try finding another workaround/fix somewhere else 🤔 
As @jenshaase mentioned: _the real problem here is the macro expansion. The expansion happens in the wrong namespace context._

Chemaclass (changes requested):

src/phel/core.phel
geoffreyvanwyk: I am stuck on this one. I saw that Clojure also does not allow private functions in macros because they end up in different namespaces.

Another solution might be to use the body of equals1 as an anonymous function in the case macro.

smeghead:
@jenshaase @Chemaclass

I was investigating the cause of the change in behavior when the first argument of `case`, expression, is an atom or an expression.

```clojure
phel:8> (def a 1)
phel:9> (case a 1 :one)
:one
phel:10> (case (+ a 0) 1 :one)
Cannot resolve symbol 'phel\core/equals1'
in string:10
```

I found that there is a process in `Analyzer#analyzeMacro` that allows private access and revokes the permission after analysis.

* src/php/Compiler/Domain/Analyzer/Analyzer.php

```php
    public function analyzeMacro(TypeInterface|array|string|float|int|bool|null $x, NodeEnvironmentInterface $env): AbstractNode
    {
        $this->globalEnvironment->setAllowPrivateAccess(true);
        $result = $this->analyze($x, $env);
        $this->globalEnvironment->setAllowPrivateAccess(false);

        return $result;
    }
```

The `analyzeMacro` seemed to be called recursively.
In the case of the `Cannot resolve symbol 'phel\core/equals1'` error, it seems to be trying to analyze `equals1` after private access has been revoked.

Do we need to control the `allowPrivateAccess` of the `GlobalEnvironment` so that it is also compatible with recursive calls, e.g., counts instead of flags?
I am not certain if this is the right way to do it, as I have only been able to partially investigate.

Issue #728, opened 2024-06-26
Investigate the form after the macro has been expanded (like `macroexpand` in clojure)


smeghead:
Macros are a very powerful feature, but they are very difficult to debug because they are processed in two stages: macro expansion and eval.

When using `defmacro` to define custom macros, we often want to see how the macros are expanded during debugging.

We believe that something like `macroexpand` in clojure would make it easier to define macros in phel.
https://clojuredocs.org/clojure.core/macroexpand

This will be useful not only when defining custom macros, but also when investigating bugs.

Chemaclass:
Just to make sure I understood this: the goal is to have `macroexpand` returning the result from the phel macro? Would it be sufficient to result the macro expansion without the evaluation?

smeghead:
> result the macro expansion without the evaluation

Yes, this is just what I want check.
This function is almost for debugging.

Chemaclass:
@smeghead could you start creating a failing phel test with the expected assertion that you would like to have? You can start with that and create a draft-PR, and we can work on the same branch together if needed 😄

smeghead:
@Chemaclass 
I pushed experimental draft branch.

https://github.com/phel-lang/phel-lang/tree/feat/add-macroexpand

https://github.com/phel-lang/phel-lang/blob/6fcdfeb79e5a5c957120c91d5e7ebd6de19d8b77/src/php/Compiler/Domain/Analyzer/Ast/MacroExpandNode.php#L26-L29

I tried to start implementing a `quoteNode` method to quote a Node, but it would probably be difficult to implement because it would have to do it for every Node.

Is there a better way?

Chemaclass:
@smeghead can you create a draft-PR? This way will be simpler to write notes/comment on your current changes 
<img width="390" alt="Screenshot 2024-07-07 at 13 52 15" src="https://github.com/phel-lang/phel-lang/assets/5256287/38bb3aa9-523e-4078-9e9a-118900eff604">

smeghead:
@Chemaclass 
It was very difficult to implement macroexpand on my own, so the work has not progressed.
However, my original goal was to have a function that would be useful for debugging when creating custom macros.
I found the function `compile` useful for this purpose.

> https://phel-lang.org/documentation/api/#compile
> (compile form)
> Returns the compiled PHP code string for the given form.

```clojure
phel:3> (def value "hello")
1
phel:4> (compile '(case value "x" 0 "hello" "HELO"))
[...]

PR #733, opened 2024-07-07
feat: add `macroexpand` special form. wip

smeghead:
### 🤔 Background

Investigate the form after the macro has been expanded (like macroexpand in clojure) #728

I want to add `macroexpand`.

### 💡 Goal

The goal is to be able to use `macroexpand` to examine the form after the macro has been expanded.

### 🔖 Changes

Added `macroexpand` as a special form.
However, I have not been able to implement the part where the macro expands the form and displays it as is, because of the complexity of how to implement it.

Is there a better way to achieve this than the current method?

### DIFF
--- a/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/MacroExpandSymbol.php
+++ b/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/MacroExpandSymbol.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Phel\Compiler\Domain\Analyzer\TypeAnalyzer\SpecialForm;
 
 use Phel\Compiler\Domain\Analyzer\Ast\MacroExpandNode;
-use Phel\Compiler\Domain\Analyzer\Ast\QuoteNode;
 use Phel\Compiler\Domain\Analyzer\Environment\NodeEnvironmentInterface;
 use Phel\Compiler\Domain\Analyzer\Exceptions\AnalyzerException;
 use Phel\Compiler\Domain\Analyzer\TypeAnalyzer\WithAnalyzerTrait;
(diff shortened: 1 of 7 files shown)


src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/MacroExpandSymbol.php
```diff
@@ -0,0 +1,46 @@
...
+        if ($form instanceof PersistentListInterface
+            && $form->get(0) instanceof Symbol
+            && $form->get(0)->getName() === Symbol::NAME_QUOTE
+        ) {
+            $form = $form->rest()->first();
+        }
+
+        return new MacroExpandNode(
+            $env,
+            $this->analyzer->analyze($form, $env),
```
Chemaclass, line 42: @jenshaase, I think the missing part here is, how do we pass the actual macro content here, so it is treated as such (like a literal)? Is this what is missing, @smeghead ?

src/php/Compiler/Domain/Emitter/OutputEmitter/NodeEmitter/MacroExpandEmitter.php
```diff
@@ -0,0 +1,24 @@
...
+final class MacroExpandEmitter implements NodeEmitterInterface
+{
+    use WithOutputEmitterTrait;
+
+    public function emit(AbstractNode $node): void
+    {
+        assert($node instanceof MacroExpandNode);
+
+        $form = $node->getValue();
+        $this->outputEmitter->emitNode($form);
```
Chemaclass, line 22: mmm and probably here, we might need to do something else to emit the macro code as PHP 🤔

src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/MacroExpandSymbol.php
Chemaclass: Right now the output is this 
Screenshot 2024-07-08 at 22 48 26

Chemaclass:
Closed in favor of https://github.com/phel-lang/phel-lang/pull/810

Issue #737, opened 2024-07-15
If the second argument of the `find` function specifies an empty vector, the function of the first argument is called.


smeghead:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.15           |
| PHP version         | 8.2.21           |

#### Summary

If the second argument of the `find` function specifies an empty vector, the function of the first argument is called.

#### Current behavior

In the `(find pred xs)` function, when an empty vector is specified for `xs`, the function specified for `pred` is executed.

#### How to reproduce

I am checking with `phel repl`.

The first argument of the `find` function is an anonymous function that prints the current date and time.
If I specify an empty vector as the second argument, I expect the function in the first argument not to be executed, but it is called and the current date/time is displayed.

```clojure
phel:1> (find (fn [x] (println (php/date "Y-m-d H:i:s")) (= x 1)) [1])
2024-07-15 08:24:04
1
phel:2> (find (fn [x] (println (php/date "Y-m-d H:i:s")) (= x 1)) [])
2024-07-15 08:24:10
nil
```

#### Expected behavior

In the `(find pred xs)` function, when an empty vector is specified for `xs`, the function specified for `pred` should not be executed.

PR #738, opened 2024-07-16
Fix `find` function for an empty vector

smeghead:
### 🤔 Background

Closes: #737 

### 💡 Goal

The `pred` function is not executed when an empty vector is searched for by the `find` function, etc.

### 🔖 Changes

Fix `loop` recursion termination conditions to evaluate using `empty?`.


Chemaclass (approved):
Nice one!

Issue #725, opened 2024-06-11
Global bindings are not immutable, contrary to documentation


geoffreyvanwyk:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | v0.14           |
| PHP version         | 8.3.7           |

#### Summary

The documentation states:

> `(def name meta? value)`
> This special form binds a value to a global symbol. **A definition cannot be redefined at a later point.**

#### Current behavior

A definition can be redefined without an exception being raised.

#### How to reproduce

Enter the following expressions into the REPL.

```phel
(def my-name "PHP")
(def my-name "Phel")
```

#### Expected behavior

An exception is raised with a message that `my-name` is already bound.

Chemaclass:
This is a known feature. On the REPL, `defn` allows you do redefine the same global binding - similar as in Clojure.
I am not sure if we want to change this behaviour; I would rather suggest update the docs to add this knowledge instead.
OK for you, @jenshaase?

geoffreyvanwyk:
> This is a known feature. On the REPL, `defn` allows you do redefine the same global binding - similar as in Clojure.

In this case, it is just `def`, not `defn`.

t-matsudate:
Hi, I'm @t-matsudate.

I consider this occurs because the `def` symbol analyzer allows any duplicates of namespace/name pairs.

https://github.com/phel-lang/phel-lang/blob/2a1342091d269ce2a83ecac26ade82e444c336f4/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/DefSymbol.php#L50

Therefore, I'd like to propose following code to resolve this issue.

```php
// DefSymbol.php
if (!$this->analyzer->hasDefinition($namespace, $nameSymbol->getName())) {
    $this->analyzer->addDefinition($namespace, $nameSymbol);
} else {
    throw AnalyzerException::definitionDuplication(...);
}
```

However I've been not sure specification of the Phel yet.
So I wish to hear everyone's considieration of this issue.

Chemaclass:
Oh, good spot. Now I get what you mean. Do you mind opening a PR with your proposal, I think that should be correct in order to preserving immutability of the definitions. @jenshaase, was it some reason why we have this behaviour that I cannot recall? Or we just simply missed it?

t-matsudate:
@Chemaclass Okay. please wait some moment. I'll create PR after implementing it with tests.

smeghead:
[...]

PR #740, opened 2024-07-27
Added: Tests to proof documentations.

t-matsudate:
### 🤔 Background

REPL can redefine vars though documentation told it can't.
(https://github.com/phel-lang/phel-lang/issues/725)

### 💡 Goal

Unifies the result to either of.

### 🔖 Changes

https://github.com/t-matsudate/phel-lang/commit/f22b86ae17134e3d6ccac7f0a6ef010658c516c6
https://github.com/t-matsudate/phel-lang.org/commit/301f080106f3dac0b3f9cab072eaa1040916bbb0

### DIFF
--- a/tests/phel/test/special-forms.phel
+++ b/tests/phel/test/special-forms.phel
@@ -3,3 +3,15 @@
 
 (deftest test-set-object-property
   (is (= "foo" (php/-> (php/oset (php/-> (php/new \stdClass) name) "foo") name)) "set value to stdClass"))
+
+(def duplicated "first-duplication")
+(def first? (= "first-duplication" duplicated))
+
+(deftest test-def-before-duplication
+  (is (= true first?)))
+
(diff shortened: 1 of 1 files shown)


smeghead:
@t-matsudate 
Thank you.
`def` is difficult to test because it cannot be run inside `deftest`.

It's a bit tricky, but how about defining the code using `def` as a string and eval it to execute it inside `deftest`?

```clojure
(deftest test-def-after-duplication
  (let [form (read-string "(do (def a \"first\") (def a \"second\") a)")]
    (is (= "second" (eval form)))))
```

smeghead (approved):

Chemaclass (approved):

Chemaclass:
I think it's good to have this test to proof that def now allow re-definitions. Once we change this behaviour to encourage immutability, then this test must be updated because it should break :)

Issue #741, opened 2024-07-31
If the second argument of the `some?` function specifies an empty vector, the function of the first argument is called.


smeghead:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.15           |
| PHP version         | 8.2.21         |

#### Summary

If the second argument of the find function specifies an empty vector, the function of the first argument is called.
This is a bug similar to #737.

#### Current behavior

In the `(some? pred xs)` function, when an empty vector is specified for `xs`, the function specified for `pred` is executed.

#### How to reproduce

I am checking with `phel repl`.

The second argument of `some?` is an empty `vector`, so the `pred` function should not be executed, but currently it is.

```clojure
phel:7> (some? |(throw (php/new \Exception)) [])
Phel\Compiler\Domain\Evaluator\Exceptions\CompiledCodeIsMalformedException: 
in /usr/src/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php:14 (gen: /usr/src/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Evaluator/Exceptions/CompiledCodeIsMalformedException.php:14)

#0 /usr/src/vendor/phel-lang/phel-lang/src/php/Compiler/Domain/Evaluator/RequireEvaluator.php(42): Phel\Compiler\Domain\Evaluator\Exceptions\CompiledCodeIsMalformedException::fromThrowable(Object(Exception))
(snip)
```

#### Expected behavior

In the `(some? pred xs)` function, when an empty vector is specified for `xs`, the function specified for `pred` should not be executed.

PR #747, opened 2024-08-01
Fix `some?` function for an empty vector

smeghead:
### 🤔 Background

Closes: #741

### 💡 Goal

The pred function is not executed when an empty vector is searched for by the `some?` function.

### 🔖 Changes

Fix loop recursion termination conditions to evaluate using `empty?`.


Chemaclass:
@smeghead would it make sense to add a test for this fixed functionality?

smeghead:
@Chemaclass 
Sorry, due to a manipulation error in git, the test was not included.
Tests have been added.

Chemaclass (approved):
Thanks!

Issue #748, opened 2024-08-01
A simple `binding` function call where `body` is atom causes an error.


smeghead:
| Q                   | A               |
|---------------------|-----------------|
| Phel version        | 0.15           |
| PHP version         | 8.2.21           |

#### Summary

A simple `binding` function call where `body` is atom causes an error.

#### Current behavior

When I evaluate `(binding [x 2] x)` I get an error.

#### How to reproduce

I am checking with repl.

```clojure
phel:1> (def x 1)
1
phel:2> (binding [x 2] (identity x))
2
phel:3> (binding [x 2] x)
Error: Call to undefined method Phel\Lang\Symbol::get()
in /usr/src/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/TrySymbol.php:35 (gen: /usr/src/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/TrySymbol.php:35)

#0 /usr/src/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/AnalyzePersistentList.php(56): Phel\Compiler\Domain\Analyzer\TypeAnalyzer\SpecialForm\TrySymbol->analyze(Object(Phel\Lang\Collections\LinkedList\PersistentList), Object(Phel\Compiler\Domain\Analyzer\Environment\NodeEnvironment))
#1 /usr/src/src/php/Compiler/Application/Analyzer.php(112): Phel\Compiler\Domain\Analyzer\TypeAnalyzer\AnalyzePersistentList->analyze(Object(Phel\Lang\Collections\LinkedList\PersistentList), Object(Phel\Compiler\Domain\Analyzer\Environment\NodeEnvironment))
#2 /usr/src/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/DoSymbol.php(45): Phel\Compiler\Application\Analyzer->analyze(Object(Phel\Lang\Collections\LinkedList\PersistentList), Object(Phel\Compiler\Domain\Analyzer\Environment\NodeEnvironment))
#3 /usr/src/src/php/Compiler/Domain/Analyzer/TypeAnalyzer/AnalyzePersistentList.php(56): Phel\Compiler\Domain\Analyzer\TypeAnalyzer\SpecialForm\DoSymbol->analyze(Object(Phel\Lang\Collections\LinkedList\PersistentList), Object(Phel\Compiler\Domain\Analyzer\Environment\NodeEnvironment))
[...]

PR #749, opened 2024-08-05
Fix `binding` function for atom body

smeghead:
### 🤔 Background

Closes: #748

### 💡 Goal

No error occurs when an atom value is specified in the `body` of the `binding` function.

### 🔖 Changes

In `binding` macro expansion, the way `body` is expanded has been corrected.


Chemaclass (approved):
Wow, very cool! Kudos for the simplicity in the solution :)

Issue #763, opened 2024-11-02
Add native phel symbols to API facade functions


Chemaclass:
## Current

Related to https://github.com/phel-lang/phel-lang.org/issues/87 

The `ApiFacade::getPhelFunctions()` is missing all `Phel\Lang\Symbol::NAME_*` 
for this reason, they are not being rendered on the API docs https://phel-lang.org/documentation/api/

## Expected

Add the phel native symbols as part of the `ApiFacade`:

```php
namespace Phel\Lang;

final class Symbol 
{
    public const NAME_APPLY = 'apply';

    public const NAME_CONCAT = 'concat';

    public const NAME_DEF = 'def';

    public const NAME_DEF_STRUCT = 'defstruct*';

    public const NAME_DO = 'do';

    public const NAME_FN = 'fn';

    public const NAME_FOREACH = 'foreach';

    public const NAME_IF = 'if';

    public const NAME_LET = 'let';

    public const NAME_LOOP = 'loop';

    public const NAME_NS = 'ns';

    public const NAME_PHP_ARRAY_GET = 'php/aget';

    public const NAME_PHP_ARRAY_PUSH = 'php/apush';

    public const NAME_PHP_ARRAY_SET = 'php/aset';

    public const NAME_PHP_ARRAY_UNSET = 'php/aunset';

    public const NAME_PHP_NEW = 'php/new';

    public const NAME_PHP_OBJECT_CALL = 'php/->';

    public const NAME_PHP_OBJECT_STATIC_CALL = 'php/::';

    public const NAME_QUOTE = 'quote';

    public const NAME_RECUR = 'recur';

    public const NAME_UNQUOTE = 'unquote';

    public const NAME_UNQUOTE_SPLICING = 'unquote-splicing';

    public const NAME_THROW = 'throw';

    public const NAME_TRY = 'try';

    public const NAME_PHP_OBJECT_SET = 'php/oset';

    public const NAME_LIST = 'list';

    public const NAME_VECTOR = 'vector';

    public const NAME_MAP = 'hash-map';

    public const NAME_SET_VAR = 'set-var';

    public const NAME_DEF_INTERFACE = 'definterface*';
```

PR #764, opened 2024-11-02
Add native phel symbols to api facade

Chemaclass:
### 🤔 Background

Closes: #763

### 💡 Goal

Add native phel symbols to API documentation

### 🔖 Changes

- Add manually the documentation for PRIVATE_SYMBOLS which are functions that does not exists on phel code

### DIFF
--- a/src/php/Api/Infrastructure/PhelFnLoader.php
+++ b/src/php/Api/Infrastructure/PhelFnLoader.php
@@ -119,7 +119,7 @@ Defines the namespace for the current file and adds imports to the environment. 
 Equivalent to PHP\'s `arr[index] ?? null`.
 Examples [here](/documentation/php-interop/#get-php-array-value).',
             'fnSignature' => '(php/aget arr index)',
-            'desc' => 'Equivalent to PHP\'s `arr[index] ?? null`.',
+            'desc' => "Equivalent to PHP's `arr[index] ?? null`.",
         ],
         Symbol::NAME_PHP_ARRAY_SET => [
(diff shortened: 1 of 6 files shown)


src/php/Api/Infrastructure/PhelFnLoader.php
```diff
@@ -17,13 +18,176 @@
...
+        ],
+        Symbol::NAME_IF => [
+            'doc' => '```phel
+(if test then else?)
+```
+A control flow structure. First evaluates test. If test evaluates to true, only the then form is evaluated and the result is returned. If test evaluates to false only the else form is evaluated and the result is returned. If no else form is given, nil will be returned.
+
+The test evaluates to false if its value is false or equal to nil. Every other value evaluates to true. In sense of PHP this means (test != null && test !== false).',
+            'fnSignature' => '(if test then else?)',
+            'desc' => 'A control flow structure. First evaluates test. If test evaluates to true, only the then form is evaluated and the result is returned. If test evaluates to false only the else form is evaluated and the result is returned. If no else form is given, nil will be returned.',
```
JesusValera, line 65: Is there any reason why only `NAME_IF` has the documentation and other fields filled while the others do not?

JesusValera (approved):

src/php/Api/Infrastructure/PhelFnLoader.php
Chemaclass: yes, it is WIP, I am working on it ;)

Issue #784, opened 2024-12-13
Macros can't be read from namespaces having hyphens


jasalt:
- PHP 8.2.20 (cli)
- Phel dev-main 059c8cd

I was looking to have a `time` macro similar to [Clojure](https://clojuredocs.org/clojure.core/time) and found an issue while translating it to Phel. Here's what I have so far following [Clojure sources](https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L3885) with added `gensym`'s as Phel does not support similar `#` gensym macro syntax:

```
(defmacro time
  "Evaluates expr and prints the time it took.  Returns the value of expr."
  [expr]
  (let [start (gensym)
        ret (gensym)]
    `(let [,start (php/microtime true)
           ,ret ,expr]
       (println "Elapsed time:" (* 1000 (- (php/microtime true) ,start)) "msecs")
       ,ret)))
```

It works initially as expected when defined in REPL:

```
(time (do (php/sleep 1) :well-slept))
# -> Elapsed time: 1000.1718997955 msecs
# -> :well-slept
```

However, I started having trouble with it after placing it in `utils.phel` file in my project when requiring it in REPL or by running as a script in general. Example here's with a fresh REPL session:

```
(require my-ns\utils :refer [time])
(time (do (php/sleep 1) :well-slept))
# -> (let [__phel_816 (microtime true) __phel_817 2] (println Elapsed time: (* 1000 (- (microtime true) __phel_816)) msecs) __phel_817)

(eval (time (do (php/sleep 1) :well-slept)))
# -> Elapsed time: 1000.1718997955 msecs
# -> :well-slept
```

Now the output is a list, as if it was "macroexpanded" or something, and it seems that `eval` has to be called on it to get it evaluated as expected. The same seems to happen when defining this macro in a file, [cli-skeleton example code](https://github.com/jasalt/cli-skeleton/tree/macro-eval-issue?tab=readme-ov-file).

I'm bit confused at which is at fault here, if it's a bug in my code that is just not triggering with REPL as it looked like it worked initially, or something else.

Also interested in feedback about the `time` macro itself if it's something that could be added to Phel. I'm not very experienced writing macros so there might be some mistakes because of that but I could make a feature issue or PR if it sound like a good idea.

smeghead:
I was researching this issue.
[...]

PR #791, opened 2025-02-02
Add doto macro

jasalt:
Add Clojure influenced macro saving some typing while mutating objects. Sending PR to core as I had trouble using it from a utility library (https://github.com/phel-lang/phel-lang/issues/784) but nevertheless this could be useful to include in `core.phel`. Similar macro has been in Clojure core library since 1.0 ([clojuredocs.org](https://clojuredocs.org/clojure.core/doto)).

I followed example of existing threading macros in `core.phel` which seem to originate from Clojure core also and added this near those as it seemed fitting.

One note is that I was left wondering why `core.phel` threading macros don't preserve the metadata as Clojure does, is it not possible in Phel or could it be that it just has been left unimplemented while the equivalent functions for reading/writing metadata seemingly exist (`set-meta!` and `meta`)? Following the example of existing macros, I left out preserving the metadata with the `doto` included in PR. The original version preserving metadata was following (didn't test metadata part though):

```clojure

(defmacro doto
  "Evaluates x then calls all of the methods and functions with the
  value of x supplied at the front of the given arguments.  The forms
  are evaluated in order.  Returns x."
  [x & forms]
  (let [gx (gensym)]
    `(let [,gx ,x]
       ,@(map (fn [f]
                (set-meta!
                 (if (list? f)
                   `(,(first f) ,gx ,@(next f))
                   `(,f ,gx))
                 (meta f))
                )
              forms)
          ,gx)))
;;## Example code (not in PR)
```

While preparing commit I noted a pre-commit hook test failing while I had added test to the macro, I incremented the hardcoded number in `ApiFacadeTest.php` expecting that it's supposed to be incremented by hand like that as I didn't notice info on it in contributor docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 - Use PHPStan level 2
+- Add `doto` macro (#791)
 
 ## [0.16.1](https://github.com/phel-lang/phel-lang/compare/v0.16.0...v0.16.1) - 2024-12-13
 
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1772,6 +1772,21 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
            ,@(interleave (repeat (count forms) name) forms)]
        ,name)))
 
+(defmacro doto
+  "Evaluates x then calls all of the methods and functions with the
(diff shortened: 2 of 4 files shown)


Chemaclass (commented):
Love it, thanks! Can you also add a line to the `CHANGELOG.md` file? This way will be easier for me when creating a new release :)

Chemaclass (approved):
Thank you!

PR #792, opened 2025-02-03
Add `time` macro

jasalt:
Add another macro from Clojure (since 1.0). Allows rough timing of executed code with execution time printed in stdout and execution result being returned to caller.

Test only checks that printing happens and expression value is returned as expected. Testing that measurement is actually effective might be useful but it gets bit tricky. I didn't notice any tests for the macro in Clojure core either so it's probably not that important. Macro is generally used ad-hoc during development and not as something a running system would rely on to produce correct results.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


Chemaclass (approved):
thanks!

PR #795, opened 2025-03-30
Add if-let and when-let macros

jasalt:
It does similar validation to Clojure in somewhat verbose way as private assert-args macro was left out:
https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L1849

Opinions / improvement ideas welcome.

### 🤔 Background

Noting these are quite popular in general and I have been missing them quite often..

### DIFF
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -1844,6 +1844,46 @@ returns 1. If `xs` has one value, returns the reciprocal of x."
        (definterface* ,name ,@fns)
        ,@defs)))
 
+(defmacro if-let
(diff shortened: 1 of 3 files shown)


jasalt:
Dang, seems that CI setup broke with `Error: Missing download info for actions/cache@v4.0.2`. Test suite ran ok locally..

jasalt:
Added macro comments that were forgotten..

Chemaclass:
The error seems unreleated to the pr itself, but to some weird missing stuff on the ci mmmm

Chemaclass (approved):

PR #798, opened 2025-04-29
fix: Fix `php/echo` does not work #729

### LINKED ISSUE #729: php/echo does not work
| Phel version        | 0.15         |
| PHP version         | 8.2           |

### Summary

- `print` works:
```phel
(php/print "hi")
``` 

- `echo` doesn't work:
```phel
(php/echo "hi")
``` 

### How to reproduce

Screenshot 2024-06-26 at 22 32 37

### Expected behavior

`php/echo` should works the same way as `php/print`

smeghead:
## 🤔 Background

fix #729 

``` 
(ns test)

(php/echo "Hi!") 
```

The above Phel code will compile to the following PHP code.

``` 
<?php 
// /usr/src/src/example.phel 
// ;;AAEA,OAAC,KAAD,MAAA,CAAA 
return echo("Hi!"); 
```

``echo`` is a language construct that does not return a value. It appears that a syntax error occurs when a language construct that does not return a value is placed after ``return`.

## 💡 Goal

`php/echo` should work as well as `php/print`.

## 🔖 Changes

Replace `php/echo` with `print` when outputting to PHP so that it is not affected by `echo`, a language construct that does not return a value.

The concern is that the PHP `echo` is not executed even though it is listed as `php/echo`.
If it should be important to run PHP `echo`, there is a way to create a wrapper function and call the wrapper function instead of `echo`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Add `doto` macro (#791)
 - Use PHPStan level 2
 - Add `if-let` and `when-let` macros (#795)
+- Fix `php/echo` does not work (#729)
 
 ## [0.16.1](https://github.com/phel-lang/phel-lang/compare/v0.16.0...v0.16.1) - 2024-12-13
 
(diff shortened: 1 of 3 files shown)


JesusValeraDev (approved):

Chemaclass (approved):
Alright! Thank you! I might create a new release in May 👍🏼

PR #811, opened 2025-05-30
Fix inmutable global binding

Chemaclass:
## 🤔 Background

Fixes: https://github.com/phel-lang/phel-lang/issues/725
Closes: https://github.com/phel-lang/phel-lang/pull/726

## 💡 Goal

Raise an exception when reusing a global binding name


Chemaclass:
@smeghead oh! good catch, you are 100% right. I will fix it in a follow up PR

PR #812, opened 2025-05-30
Add source filename:line on repl exceptions

Chemaclass:
## 💡 Goal

Improve DX when an facing exception on repl

### BEFORE (main)
Screenshot 2025-05-30 at 10 31 14

### AFTER (this pr)
Screenshot 2025-05-30 at 10 30 40


JesusValeraDev (approved):
Buenísima 💯

PR #831, opened 2025-05-31
Fix macro lookup for hyphen namespaces

Chemaclass:
Relates: https://github.com/phel-lang/phel-lang/issues/784

## Summary
- ensure macro lookup encodes namespaces with Munge
- add regression test for macros in hyphenated namespaces


Chemaclass:
Closed in favor of https://github.com/phel-lang/phel-lang/pull/828

PR #844, opened 2025-06-09
Add doctor command

Chemaclass:
## 🤔 Background

This is especially useful when using the PHAR executable, to make sure if you can run it or not.

## 🔖 Changes

- Create a new command `phel doctor`

Screenshot 2025-06-09 at 19 42 47


JesusValeraDev:
Did you add this `doctor` command somewhere in the doc? 🤔

PR #859, opened 2025-08-02
Add `symbol`

jasalt:
Needed way to define `php/CONSTANT`'s dynamically in macro. With this it's possible e.g. `(symbol (str "php/" "MY-CONSTANT"))`.

Looked at [Janet](https://github.com/janet-lang/janet/blob/ddc122958be34a18edd754743851536b5a16a3cc/src/core/corelib.c#L337) and [Clojure](https://github.com/clojure/clojure/blob/da1c748123c80fa3e82e24fc8e24a950a3ebccd9/src/clj/clojure/core.clj#L591) implementations, Clojure having explicit namespace support seemed more robust so took more influence from it as it seems to match `\Phel\Lang\Symbol` class static methods also nicely. 

Not sure of best way to define two different arities and couldn't grasp the usage of `{:inline-arity ...}` in some core library functions but better ideas are welcome.

Noted one warning in test suite but it's not related to this change, adding issue about it..

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+- Add `symbol` function
+
 ## [0.19.0](https://github.com/phel-lang/phel-lang/compare/v0.18.1...v0.19.0) - 2025-08-02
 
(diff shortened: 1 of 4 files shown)


Chemaclass (approved):
It is good as the first iteration for this function, thanks!

PR #878, opened 2025-08-17
feat: add `select-keys` function

jasalt:
Add convenience function familiar from Clojure for working with maps.

Draft as diagnosing why `contains-value?` is not behaving correctly with `nil` value..


Chemaclass:
The https://github.com/phel-lang/phel-lang/pull/880 was merged. Feel free to update this PR and also resolve the conflicts so that we can merge this PR as well 👍🏼

jasalt:
@Chemaclass yes, I think it's more readable while basically doing the same thing. Here you go.

Chemaclass (approved):
thanks!

Issue #879, opened 2025-08-17
`contains-value?` does not treat `nil` as value


jasalt:
| Q                   | A               |
|---------------------|-----------------|
| PHP version         |  8.3        |
| Phel version        | 0.19.1  ([925ea02](https://github.com/phel-lang/phel-lang/commit/925ea02f4d9b4960071b06e5e8a24f4ae9ff2932))          |

While writing tests for https://github.com/phel-lang/phel-lang/pull/878 I noted that `contains-value?` does not treat `nil` as a value:
```
(contains-value? [1 2 3 4 nil] nil)
## => false
```
Seems like a bug considering `(contains? {nil 1 :a 2} nil) # => true` which is also a behavior Clojure has.

jasalt:
Seems easy to fix, working on it..

PR #880, opened 2025-08-17
fix: `contains-value?` with nil value

jasalt:
## 🤔 Background

Fixes https://github.com/phel-lang/phel-lang/issues/879

## 🔖 Changes

Simplifies the function and makes it work with `nil` values.

Adds some extra tests and expected result of existing test case `(is (false? (contains-value? full-name nil)))` which opposes the new behavior.


Chemaclass (approved):
Thanks!

Chemaclass:
Please update the changelog.md with this new fix/feature under the top unreleased section

Issue #907, opened 2025-08-25
Inline comment syntax (like Clojure `#_`)


jasalt:
Clojure has a nice reader macro shortcut for ignoring the next form(s):

```clojure
(concat [1 2 3] [4 5 6] [7 8 9])     ;; => (1 2 3 4 5 6 7 8 9)
(concat #_[1 2 3] [4 5 6] [7 8 9])   ;; => (4 5 6 7 8 9)
(concat #_#_[1 2 3] [4 5 6] [7 8 9]) ;; => (7 8 9)
```
It can also be "stacked" to ignore multiple succeeding forms as in the last example.

This comes handy while debugging complex nested forms where it takes many editor actions to comment out just the specific inner form, without commenting out any extra closing parens.

Scheme also apparently has `#;` for the same effect. Phel using `#` for comments, maybe `;_` could work, but that's just initial idea.

Janet discussion on adding the feature while they didn't add it apparently yet: https://github.com/janet-lang/janet/discussions/1272

PR #914, opened 2025-08-25
Add inline comment with macro #_

Chemaclass:
## 🤔 Background

Resolves https://github.com/phel-lang/phel-lang/issues/907 by @jasalt 

## 🔖 Changes

- Introduced an inline comment reader macro `#_` in the lexer
  -  enabling single-form discards and adjusting downstream token indices accordingly
- Added a dedicated token (`T_COMMENT_MACRO`) and parser node (`CommentMacroNode`) to drop discarded forms while maintaining source locations
- Updated token stream handling to treat inline comments like regular comments

## 🖼️ Demo

Screenshot 2025-08-25 at 22 53 49


JesusValeraDev (approved):
<3

PR #915, opened 2025-08-26
Add ; as alternative comment symbol

jasalt:
## 🤔 Background

Most lisps as far as I know use `;` as comment character and Phel using `#` is a slight inconvenience when switching between languages. I thought initially that maybe `;` is avoided cause of it's strong meaning in PHP, but then I'm interested to know what practical implications it would have if it was allowed as a alternative comment character in Phel. 

This PR experiments with adding that as alternative to the current `#` comment character in lexer.
Tests show green which is a good sign.

Thoughts?

With this, following syntax is valid:

```clojure
(ns testing-semicolon)

;; Demo comment
## Alternative to #
(defn hello []
  (print "hello world"))

(hello)  ; Nice
```
Added benefit is also that it works better with Clojure syntax highlighting rules.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Fix vector and hash map literals in try
 - Fix REPL symbol resolution issue loading current dir
 - Add `#_` macro for inline comments
+- Allow `;` as alternative comment character
 
 ## [0.20.0](https://github.com/phel-lang/phel-lang/compare/v0.19.1...v0.20.0) - 2025-08-25
 
--- a/src/php/Compiler/Application/Lexer.php
+++ b/src/php/Compiler/Application/Lexer.php
@@ -20,7 +20,7 @@ final class Lexer implements LexerInterface
         "([ \t]+)", // Whitespace (index: 2)
         "(\r?\n)", // Newline (index: 3)
         '(#_)', // Inline comment (index: 4)
-        "([#;][^\n]*\n?)", // Comment (index: 5)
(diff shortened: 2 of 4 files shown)


Chemaclass:
mmm I don't dislike the idea. Can you try adding tests for this? Thanks!

Chemaclass:
## 🧪 Demo

Screenshot 2025-08-26 at 17 04 02

jasalt (commented):
I was looking to add the same `test_read_semicolon_comment` test also, but it's good to have the additional edge case tests too.

I'll use this with some code I'm writing so will see if there's any unwanted side effects.

jasalt:
> @jasalt what do you mean with `This breaks a bit when simply swapping # for the comment character in Emacs Clojure highlight rules and it only grays out a single line (Clojure uses # e.g. for lambda and auto gensym syntaxes).` I am not sure I understood what's the issue and how related to this changes are. Is that coming from this PR or something extra that we could tackle in other iteration?

This was more of a minor notice of what positive side effects I see with the `;` comment syntax made possible by the PR regarding code highlighting tools.

To reiterate, with `;` comment syntax, more standard lisp syntax highlighting rules can be re-used generally and I noted it fixing issue in my editor with `#_` syntax highlighting with rules inherited from Clojure. Previously when using `#` as comment character the behavior was following (contrast to earlier screenshot where it works):
Screenshot_20250827_072652
Example code in screenshot has minor change but it shows the inline s-exp comment highlighting rule in action similarly.

jasalt:
Haven't faced issues so far and I think it should work.

Chemaclass (approved):

PR #918, opened 2025-08-26
Show current phel repl version

Chemaclass:
### BEFORE

Screenshot 2025-08-27 at 23 09 25

### AFTER

Screenshot 2025-08-27 at 23 08 50

### DIFF
--- a/.github/RELEASE.md
+++ b/.github/RELEASE.md
@@ -2,7 +2,7 @@
 
 Guide to create a new release.
 
-1. Update the version in [ConsoleBootstrap](../bin/phel)
+1. Update the version in [VersionFinder](../src/php/Console/Application/VersionFinder.php)
 2. Update the version in [CHANGELOG.md](../CHANGELOG.md)
 3. Create a [new release](https://github.com/phel-lang/phel-lang/releases/new) from GitHub
 4. Attach the new PHAR to the new release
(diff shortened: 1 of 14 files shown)


Chemaclass:
Idea: if not in an confirmed/official version, display the commit hash next to it like: v0.20.0-beta#123456

This way it will be easier to identify when being in an official version or in development. How do other projects do something like this?
Maybe its not so important, but I wonder if others thought already about this

JesusValeraDev:
@Chemaclass but in your 2nd picture, you are in a branch, so, it's not v20 but some xyz commit (from v20).
Not sure if it brings value if you are not in a "official release version".

jasalt:
Yea, it would be convenient for troubleshooting to have a version notice in repl, plus commit hash (especially for non-release builds). The `phel version` command could have commit hash added similarly too.

Noted when testing custom Phar builds that the commit / branch info gets quite easily lost. Maybe you could dig it out from there but this would make it more convenient.

JesusValeraDev (approved):

jasalt:
@Chemaclass I don't think I've got this working, should this work when Phel is installed as Composer package in `./vendor/bin/phel repl`, or are there some gotchas with it? I see you have it working in your screenshots which is nice.

Chemaclass:
@jasalt thanks for saying. Actually, I noticed that it is not working as expected when used as vendor dependency. I will investigate this further

Issue #906, opened 2025-08-25
Multiline comments


JesusValeraDev:
### 📚 Description

So far, Phel does not allow multi-line [comments](https://phel-lang.org/documentation/basic-types/#comments).

> A comment begins with a # character and continues until the end of the line. There are no multi-line comments.

But it would be nice to be able to write them. One of the reasons (not the only one) is to collapse the comment when using the Phel plugin in JetBrains' IDE.

### 📓 To Do

- [x] Investigate how other Lisp languages handle multi-line comments
- [x] Common lisp seems to use `#|` as start comment, and `|#` as ending

jasalt:
I would avoid `##` because with many lisps, `;;` is used for regular single-line comment that gets indented to column 0, while `;` is used in comments that appear after code on same line.

Thus, modifying lisp editor modes for working with Phel would need more work. As example for Emacs, making existing Clojure mode highlight Phel (mostly) requires just:

```
(define-derived-mode phel-mode clojure-mode "Phel"
  "Major mode for editing Phel language source files."
  (setq-local comment-start "#")
  ...
)
```

Chemaclass:
@jasalt I am not sure if I understand what you are saying. What would you expect in your example? I assume `123` instead of nil, right?
```lisp
(defn foofun [] 123 (comment should return 123))
(foofun)
# => nil
```

JesusValeraDev:
> Should we try to have this for multiline comments? Or do you have other ideas, @jasalt @smeghead @JesusValeraDev ?

I like it 👍🏼

PR #931, opened 2025-08-31
Add multiline comments

Chemaclass:
## 🤔 Background

Resolves https://github.com/phel-lang/phel-lang/issues/906

## 🔖 Changes

Add multiline comments using "common lisp" syntax for multiline comments
- `#|` as start
- `|#` as ending


JesusValeraDev (changes requested):
Could you add some phel tests too?

PR #938, opened 2025-09-02
fix: `find-index` docstring correction

jasalt:
It had the same docstring as `find` which was incorrect, not mentioning it returns the index.


Chemaclass (approved):
ups, thanks!

PR #1002, opened 2025-10-05
Allow null on phel static collections arguments

Chemaclass:
## 🤔 Background

I noticed that when running phel on a web server, it's crashing because GLOBALS does not contain "argv" when I tried:

```phel
  (php/:: Phel (vector (php/aget php/$GLOBALS "argv"))))
```

I get a : **Fatal error: Uncaught TypeError: Phel::vector(): Argument #1 ($values) must be of type array, null given**

## 💡 Goal

Allow passing null as an argument, so this will be similar to an empty array, avoiding fatal errors at runtime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+- Allow null on Phel static collections (vector, list, map, set) arguments
+
 ## [0.23.0](https://github.com/phel-lang/phel-lang/compare/v0.22.2...v0.23.0) - 2025-10-05
 
 - Fix `doc` command on phar command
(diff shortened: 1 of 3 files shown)


JesusValeraDev (approved):

jasalt:
Interesting that this had broken. PR https://github.com/phel-lang/phel-lang/pull/1003 seemed like a precise fix for it.

This PR sounded bit scary first so I did a bit of testing if it added any novel `nil` conversion behavior. Quickly comparing to Clojure it seems to behave pretty much similarly so should be fine.

Weirdly in Clojure (Babashka) `(set nil)` != `#{nil}` though, not sure what is going on there.

Screenshot_20251007_230104

Chemaclass:
I would expect (set nil) is equal a set with nil. I am not sure why in clojure is an empty set(?). I guess this might be depend on the definition of set and nil in their context or a bug on their side haha

PR #1057, opened 2025-11-25
Fix example from deftest function

JesusValeraDev:
## 📚 Description

Add missing `)` in the example metadata section.


Chemaclass (approved):
thanks

PR #1072, opened 2025-12-24
Make signature an array and improve json response

JesusValeraDev:
## 📚 Description

- Transform `signature` from `string` to `array`
- Add `namespace`, `example`, and `signature` to `phel doc` command as json format response

### DIFF
--- a/src/php/Api/Application/PhelFnNormalizer.php
+++ b/src/php/Api/Application/PhelFnNormalizer.php
@@ -51,7 +51,7 @@ final readonly class PhelFnNormalizer implements PhelFnNormalizerInterface
             preg_match('#(```phel\n(?<signature>.*)\n```\n)?(?<desc>.*)#s', $doc, $matches);
 
             $signatureBlock = $matches['signature'] ?? '';
-            $signature = $this->parseSignatures($signatureBlock);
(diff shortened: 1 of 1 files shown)


src/php/Api/Transfer/PhelFunction.php
```diff
@@ -85,16 +86,20 @@ public function doc(): string
...
         return $this->signature;
     }
 
     /**
      * @deprecated in favor of the signature attribute
+     *
+     * @return list<string>
      */
-    public function signature(): string
+    public function signature(): array
```
Chemaclass: Keep `public function signature(): string` return first signature.

Create new method `public function signatures(): array`

src/php/Api/Transfer/PhelFunction.php
```diff
@@ -85,16 +86,20 @@ public function doc(): string
...
+     * @return list<string>
      */
-    public function fnSignature(): string
+    public function fnSignature(): array
     {
         return $this->signature;
     }
 
     /**
      * @deprecated in favor of the signature attribute
```
Chemaclass: delete this deprecated notice, as this will be the method to access the first signature from now on

Chemaclass (approved):
awesome!!!

PR #1077, opened 2025-12-28
Change "remove" to use Clojure semantics

### LINKED ISSUE #1076: remove does not work with vector and throws "Cannot remove" exception
| PHP version         | 8.4.3           |
| Phel version        | v0.27.0           |

#### Summary

The `remove` function throws an exception when used with a vector.

#### Current behavior

Calling `remove` with a vector results in a runtime error:

```clojure
phel:27> (remove [1 2 3] 1)

Notice: Only variables should be passed by reference in /tmp/phel/tmp/__phel_fd5a44dea238e137f7c9104ceb3b6112.php on line 4
InvalidArgumentException: Cannot remove
in /tmp/phel/cache/compiled/phel_core.php:4292 (gen: /tmp/phel/cache/compiled/phel_core.php:4292)                                                   

#0 /tmp/phel/tmp/__phel_fd5a44dea238e137f7c9104ceb3b6112.php:4 (gen: /tmp/phel/tmp/__phel_fd5a44dea238e137f7c9104ceb3b6112.php:4) : (phel\core\remove [1 2 3] 1)
#1 /usr/src/src/php/Compiler/Domain/Evaluator/RequireEvaluator.php(75): require('/tmp/phel/tmp/_...')
#2 /usr/src/src/php/Compiler/Application/EvalCompiler.php(111): Phel\Compiler\Domain\Evaluator\RequireEvaluator->eval('// string
// ;;...')
#3 /usr/src/src/php/Compiler/Application/EvalCompiler.php(70): Phel\Compiler\Application\EvalCompiler->evalNode(Object(Phel\Compiler\Domain\Analyzer\Ast\CallNode), Object(Phel\Compiler\Infrastructure\CompileOptions))
(snip)
```

The exception seems to be thrown because `vector` does not implement `RemoveInterface`.

#### How to reproduce

1. Start the Phel REPL
2. Evaluate the following expression:

```clojure
(remove [1 2 3] 1)
```

#### Expected behavior

`remove` works with vectors.

*Note:*  
In Clojure, `remove` behaves differently (`(remove pred coll)`).
The current `remove` in phel-lang appears to have specification derived from Janet's `array/remove`.

Considering that `remove` hasn't been working until now, aligning with Clojure's `remove` specification might be one option.
What do you think?

Chemaclass:
## 🤔 Background

Fixes #1076

The `remove` function was throwing an exception when used with vectors because:

1. The function signature followed Janet's `array/remove` style: `(remove coll offset &[n])` - removing elements by index
2. Vectors don't implement `RemoveInterface`, causing the "Cannot remove" exception
3. The function used `^:reference` for mutability, which doesn't work with immutable vectors

```clojure
;; Before: Janet-style (remove by index)
(remove [1 2 3] 1)
;; => InvalidArgumentException: Cannot remove
```

## 💡 Goal

Align `remove` with Clojure semantics where `remove` is the complement of `filter` - it returns elements where the predicate is false.

This approach:
- Makes `remove` work with **all sequence types** (vectors, lists, sets, maps)
- Follows Clojure's well-established API that users expect
- Provides a lazy sequence result, consistent with other sequence operations

### Breaking change: New signature

**Before (Janet-style):**
```clojure
(remove coll offset &[n])  ; Remove n elements starting at offset
```

**After (Clojure-style):**
```clojure
(remove pred coll)  ; Return elements where (pred elem) is false
```

### Examples

```clojure
;; Remove even numbers
(remove even? [1 2 3 4 5 6])
;; => (1 3 5)

;; Remove positive numbers
(remove pos? [1 -2 2 -1 3 7 0])
;; => (-2 -1 0)

;; Remove nil values
(remove nil? [1 nil 2 nil 3])
;; => (1 2 3)

;; Remove using a set as predicate
(remove #{:a} [:b :c :d :a :e :a :f])
;; => (:b :c :d :e :f)
```

### Implementation

`remove` is now simply defined as:
```clojure
(defn remove [pred coll]
  (filter |(not (pred $)) coll))
```

This leverages the existing `filter` function, making it work with any collection that `filter` supports.


smeghead (commented):
Thank you!
The Phel\Lang\RemoveInterface interface can be removed if it is no longer in use.

PR #1216, opened 2026-04-10
fix(compiler): pre-scan source namespaces for robust clojure remap

Chemaclass:
## 🤔 Background

PR #1215 fixed the eager clojure→phel remapping by checking the Registry before remapping. However, during the namespace extraction phase (before any file is compiled), the Registry is empty, so the check has a timing gap: `clojure.test` might not get remapped to `phel\test` during extraction, which could break dependency ordering in edge cases.

## 💡 Goal

Make the clojure→phel remap decision work correctly regardless of compilation phase. User-defined `clojure.*` namespaces (e.g. `clojure.core-test.portability`) should never be remapped, while references to phel standard library equivalents (e.g. `clojure.test`) should always remap.

## 🔖 Changes

- `Registry`: add declared namespace tracking (`addDeclaredNamespace`, `hasDeclaredNamespace`, `hasDeclaredNamespaces`) cleared on `clear()`
- `NamespaceExtractorInterface`: add `preRegisterDeclaredNamespaces()` method
- `NamespaceExtractor`: lightweight pre-scan (lex/parse/read only, no analyzer) before full extraction; registers all declared namespace names on the Registry
- `CachedNamespaceExtractor`: delegates pre-scan to inner extractor
- `NsSymbol::remapClojureNamespace()`: two-tier logic:
  1. If pre-scan ran: check if `clojure\*` is a declared source namespace (don't remap if so), otherwise remap unconditionally
  2. If no pre-scan (REPL mode): fall back to existing Registry definitions check
- Tests updated to use declared namespaces instead of Registry definitions; new tests for REPL fallback path and REPL unknown namespace

Closes #1210

### DIFF
--- a/src/php/Build/Application/CachedNamespaceExtractor.php
+++ b/src/php/Build/Application/CachedNamespaceExtractor.php
@@ -50,6 +50,10 @@ final readonly class CachedNamespaceExtractor implements NamespaceExtractorInter
      */
     public function getNamespacesFromDirectories(array $directories): array
     {
+        // Delegate pre-scan to the inner extractor so NsSymbol knows
+        // which clojure\* namespaces are user-defined source files.
+        $this->innerExtractor->preRegisterDeclaredNamespaces($directories);
+
         $allFiles = $this->findAllPhelFiles($directories);
 
         /** @var array<string, NamespaceInformation> $namespaces */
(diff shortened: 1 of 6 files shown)


Chemaclass:
Closing: the timing gap this addresses is theoretical — the test runner, build system, and REPL fallback from #1215 already cover all real execution paths.

JesusValeraDev (approved):

PR #1242, opened 2026-04-10
feat(test): add testing macro for grouping assertions

### LINKED ISSUE #1237: feat: add `testing` macro to `phel\test`
"Adds a new string to the list of testing contexts.  May be nested, but must occur inside a test function (deftest)." -- [Clojure doc](https://clojure.github.io/clojure/clojure.test-api.html#clojure.test/testing) / [sources](https://github.com/clojure/clojure/blob/4b9eadccce2aaf97e64bcc8e35c05c529df8fdd2/src/clj/clojure/test.clj#L597)

Used broadly in Clojure test code including in [clojure-test-suite ](https://github.com/jasalt/clojure-test-suite/blob/f9616bf8bb4ff633a405b06ba05596953344bc61/test/clojure/core_test/and.cljc#L7) which throws `[PHEL001] Cannot resolve symbol 'testing'` when ran with Phel.

Chemaclass:
## 🤔 Background

Clojure's `clojure.test/testing` is widely used to group assertions with context strings. Its absence causes `Cannot resolve symbol 'testing'` errors when running Clojure test suites in Phel (e.g. clojure-test-suite).

## 💡 Goal

Add `testing` macro to `phel\test` for Clojure compatibility.

Closes #1237

## 🔖 Changes

- Added `testing` macro to `phel\test` that executes its body (context string for readability)
- Tests cover basic usage and nested `testing` blocks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- `testing` macro in `phel\test` for grouping assertions with context strings, matching Clojure's `clojure.test/testing` (#1237)
 - `seq?` predicate function to check if a value is a seq (implements `LazySeqInterface`), matching Clojure semantics (#1231)
 - `condp` macro for predicate-based conditional dispatch, matching Clojure semantics including `:>>` result threading (#1217)
 - Allow `require` and `use` to accept quoted symbols in the REPL (e.g. `(require 'phel\str)`), matching Clojure semantics and enabling nREPL client compatibility (#1211)
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -235,6 +235,13 @@
      (binding [*current-test-name* ~(name test-name)]
               ~@body)))
 
+(defmacro testing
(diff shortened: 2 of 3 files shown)


jasalt:
@Chemaclass Looks good. Does this actually do following yet:

> "The context string is prepended to failure messages for better diagnostics."

Assuming that test runner side would also need some changes to read this content for reporting / failure messages.

jasalt:
I believe according to the description the string defined as first argument for `testing` should appear in the failing test which is not happening (though it'd be very nice). Only the string set as last argument to `is` is printed.

image

```
```

Chemaclass:
Good catch @jasalt — the initial implementation was just a pass-through `do`. Updated in 403ccb1a to actually integrate with the test runner:

- Added `*testing-contexts*` dynamic var that `testing` pushes context strings onto via `binding`
- `report` now prepends context to failure messages (e.g. `"math basics - addition"`)
- Nested `testing` blocks compose with `>` separator (e.g. `"outer > inner - msg"`)
- Added a test that verifies context appears in the captured failure data

PR #1272, opened 2026-04-10
feat(async): add pmap for fiber-based parallel map

Chemaclass:
## 🤔 Background

Follow-up to #793. Now that `phel\async` ships `async` / `await-all` on top of AMPHP (runtime dep as of #1271), a fiber-based `pmap` is a natural next step — it was requested by @jasalt [in the discussion](https://github.com/phel-lang/phel-lang/discussions/793#discussioncomment-16520304).

His earlier attempt with `amphp/parallel` hit `Serialization of 'Phel\Lang\AbstractFn@anonymous' is not allowed`. That's the multi-process/CPU-parallel story, which has real blockers (see below). This PR delivers the IO-parallel story via fibers, which covers the common case (fan out HTTP, DB, or file IO).

## 💡 Goal

Give Phel a Clojure-compatible `pmap` in `phel\async` so users can drop it in wherever they already use `map` when the work is IO-bound.

Developer experience mirrors [`clojure.core/pmap`](https://clojuredocs.org/clojure.core/pmap):

```clojure
(pmap fetch-url urls)
(pmap + [1 2 3] [10 20 30])           ; => [11 22 33]
(pmap + [1 2 3] [10 20 30] [100 200]) ; stops at shortest, like map
```

## 🔖 Changes

- `src/phel/async.phel`: add `pmap` with variadic collections, dispatching on arity via `case` (matches existing Phel idiom). Single-collection path uses `#(async (f %))`; multi-collection path threads `async` through an `apply map`.
- `tests/phel/test/async.phel`: cover single collection, order preservation, empty input, two-collection, N-collection, stops-at-shortest, zero-collection throws, and a concurrency smoke test using `delay` to prove results come back faster than sequential execution.
- `CHANGELOG.md`: entry under `## Unreleased`.

### Deliberate differences from Clojure

| Clojure | Phel | Why |
|---|---|---|
| Lazy seq | Eager vector | Fibers are cheap; laziness adds complexity for little gain and Phel idioms lean vector. |
| Semi-lazy window sized to CPU count | All futures spawned at once | No thread pool to protect. For huge N or rate-limited APIs, a bounded variant can land later. |
| Threads via `future` | Fibers via `async` | PHP reality — documented in the docstring so nobody expects CPU speedup. |

### Why this PR intentionally doesn't do CPU-parallel

Fiber `pmap` and a hypothetical process-based `ppmap` have **genuinely different contracts** (not just different implementations of the same one), and a user has to pick consciously — conflating them would cause runtime errors depending on what kind of function they passed:

| | Fiber `pmap` (this PR) | Process `ppmap` (future) |
|---|---|---|
| Parallelism | IO only (cooperative, 1 thread) | Real CPU (N worker processes) |
| `f` can be | Any closure, any captures | Named function only — closures don't serialize |
| Shared state | Atoms, dynamic bindings work | None — workers don't share memory |
| Errors | Normal exception propagation | Serialized across IPC |
| Cost per call | ~µs (fiber spawn) | ~ms (worker startup / channel round-trip) |

Three concrete blockers for `ppmap` today (tracked as a follow-up issue):

1. **Phel Lang types are not serializable.** No class in `src/php/Lang` implements `__serialize` / `__unserialize`. `PersistentVector`, `Keyword`, etc. can't round-trip through an amphp/parallel worker channel without forcing users to convert to PHP arrays/scalars at the boundary.
2. **Closures aren't serializable.** Even with the native PHP `Closure` emission from #1270, stock `Closure` can't be serialized. Options are `laravel/serializable-closure` (partial support) or requiring users to pass fully qualified named functions.
3. **Worker bootstrap.** amphp/parallel workers would need to require the Composer autoloader, bootstrap the Phel runtime, and load the user's namespace before running anything — a full subsystem including source discovery, stdio routing, and error marshalling.

Closes part of #793. `ppmap` tracked in a separate issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- `pmap` in `phel\async` for fiber-based parallel map, matching Clojure's `pmap` signature (IO-parallel via AMPHP) (#793)
 - `phel\async` module with `async`, `await`, `delay`, `await-all`, `await-any`, and `->closure` for AMPHP-based concurrency (#793)
 - `amphp/amp` promoted to a runtime dependency so `phel\async` works out of the box (including from the PHAR), without requiring consumers to install it separately
 - `reify` for anonymous protocol/interface implementation, matching Clojure's `reify` (#1226)
--- a/src/phel/async.phel
+++ b/src/phel/async.phel
@@ -48,3 +48,22 @@
   [futures]
   (let [php-futures (to-php-array futures)]
     (php/amp\future\awaitfirst php-futures)))
+
+(defn pmap
+  "Like `map`, but applies `f` to elements in parallel via fibers.
+
+  Returns a vector of results in the original order. With multiple
+  collections, applies `f` to corresponding elements from each collection,
+  stopping when the shortest collection is exhausted.
+
+  Best for IO-bound work (HTTP, DB, file IO). Because PHP fibers are
+  cooperative on a single thread, `pmap` does not speed up CPU-bound
+  computations — unlike Clojure's thread-based `pmap`."
+  {:doc "Parallel map via AMPHP fibers. IO-parallel, not CPU-parallel."
+   :example "(pmap inc [1 2 3]) ; => [2 3 4]"
+   :see-also ["map" "async" "await-all"]}
+  [f & colls]
+  (case (count colls)
+    0 (throw (php/new \InvalidArgumentException "pmap expects at least one collection"))
+    1 (await-all (map #(async (f %)) (first colls)))
+    (await-all (apply map (fn [& args] (async (apply f args))) colls))))
--- a/tests/phel/test/async.phel
+++ b/tests/phel/test/async.phel
@@ -1,6 +1,6 @@
 (ns phel-test\test\async
   (:require phel\test :refer [deftest is])
-  (:require phel\async :refer [async await await-all await-any delay ->closure]))
+  (:require phel\async :refer [async await await-all await-any delay pmap ->closure]))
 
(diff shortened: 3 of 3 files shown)


jasalt:
Nice diagnostics regarding threading and good idea to try this single threaded first. @Chemaclass

Regarding possible future support for actual threading/multiprocessing, maybe if that got accomplished it should use this `pmap` name, and this single-thread implementation would be renamed to `fpmap`, `pmap1` or something clever, just for Clojure compatibility sake.

Chemaclass:
Thanks @jasalt, I've been thinking about this a lot and I'm going to keep `pmap` as the fiber version for now rather than rename it.

Short version of the reasoning: the fiber `pmap` covers the common PHP case (IO fan-out over HTTP, DB, files) really well, and for CPU parallelism via `amphp/parallel` there's a significant amount of prerequisite work (Phel Lang types aren't serializable yet, closures can't cross worker boundaries without extra machinery, worker bootstrap is nontrivial with the Phel runtime on top of PHP) with a narrow addressable audience inside an already small language. I'd rather not do that engineering speculatively.

I've parked #1273 as "not planned unless a concrete CPU-parallel workload needs it". If you do hit a real CPU-bound use case where `amphp/parallel` directly isn't enough, ping me there and we can revisit with a specific workload in mind.

On the Clojure-compat point, you're right that this is a divergence. The docstring calls it out explicitly, and I think users migrating from Clojure are already adapting to a different host language. PHP is not the JVM and the concurrency primitives are different enough that one-to-one name mapping was never going to give one-to-one semantics.

jasalt:
@Chemaclass good call. It’s great enhancement already and there’s more practical gaps to fill before getting too much into parallelism concerns.

PR #1432, opened 2026-04-12
fix: unblock clojure-test-suite (fail-soft runner, nested class, ratio literals)

Chemaclass:
## 🤔 Background

Running [jasalt's clojure-test-suite](https://github.com/jasalt/clojure-test-suite) against current `main` was impossible in practice: the first file with a compile error aborted the whole run, so we had zero visibility into the other ~200 test files. Two further PHP-level bugs blocked individual files outright.

## 💡 Goal

Make the `.cljc` compatibility suite runnable end-to-end on stock Phel, and eliminate the three most common blockers surfacing from that run.

## 🔖 Changes

- **`phel test` is fail-soft per file.** Compile errors are reported and skipped; the rest of the suite still runs. Use `--fail-fast` for the old abort-on-first-error behavior.
- **`defstruct`/`defrecord`/`defexception`/`definterface` no longer fatal when nested in a function body.** The emitter now tracks when it is writing into a PHP class body (`FnAsClass`/`MultiFnAsClass`/`Reify`) and lifts class declarations into `eval()` in that case — e.g. a `defrecord` inside a `deftest`.
- **Ratio literals `N/M` accepted.** They read as `num / den` (float), mirroring how `1N`/`1.5M` already strip their suffix; `1/0` → `INF`, `0/0` → `NaN`. Migration doc updated.

End-to-end against the full suite: 3 files skipped (vs 10 before), **Passed 2868 / Total 3679** (jasalt's pre-PR baseline on a hand-stubbed subset: Passed 2579 / Total 3314).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ All notable changes to this project will be documented in this file.
 - `(empty? (range))` no longer hangs; `empty?` checks `first` for lazy sequences (#1366)
 - `is` no longer misinterprets `let`/`when`/`cond` forms as binary predicates (#1367)
 - `defrecord`/`defstruct`/`defexception`/`definterface` no longer emit invalid PHP namespace declarations in statement mode (#1358)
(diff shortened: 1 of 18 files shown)


Chemaclass:
Added `ratio-literal-nan.test` (`0/0` → `NaN`) and `ratio-literal-negative-inf.test` (`-1/0` → `-INF`) in 38431159 — the two zero-denominator branches are now pinned in the fixture suite. Thanks @raspi-jesus-bot.

PR #1444, opened 2026-04-14
fix(build): skip build output dir in namespace scans

Chemaclass:
## 🤔 Background

Running `composer test` printed dozens of `WARNING: Namespace '…' is defined in multiple locations` lines whenever a `./bin/phel build` had left an `out/` directory at the repo root. `FileCompiler::writeSourceReference` writes a `.phel` source copy next to every compiled `.php`, and the REPL's `NamespaceLoader` adds `getcwd()` as a scan root — so the recursive walker discovered every namespace twice.

## 💡 Goal

Prune the build output from recursive namespace scans so leftover compiled sources never shadow the real ones.

## 🔖 Changes

- `NamespaceExtractor` / `CachedNamespaceExtractor` accept absolute excluded directories plus a `destDirBasename` used to prune `<scan_root>/<basename>/` per walk.
- `BuildFactory` wires the configured output directory (absolute + basename) into both extractors.
- Unit tests cover absolute and basename-based exclusion.


Chemaclass:
Addressed in 3c8bb938 — extracted an `ExcludedScanPaths` value object (same shape as the suggested `ExclusionFilter`) so both extractors delegate through `$excludedPaths->contains($file, $scanRoot)` and the ~50 lines of duplicated normalize/isExcluded/prefixesForWalk helpers are gone.

Issue #1459, opened 2026-04-18
CompiledCodeCache is incompatible with Windows


jasalt:
- Phel v0.33.0
- PHP 8.4

Testing on Windows 11, Php 8.4, Composer, Git (native / non-WSL).

Running Phel fails on Windows due to hard coded POSIX/UNIX style path separators and possibly other issues. Stack trace from starting REPL:

```
PS  C:\demo\example-app> ./vendor/bin/phel
The system cannot find the path specified.
The system cannot find the path specified.
Welcome to the Phel Repl (v0.33.0)
Type "exit" or press Ctrl-D to exit.
PHP Warning:  mkdir(): No such file or directory in  C:\demo\example-app\vendor\gacela-project\gacela\src\Framework\Cache\FileCache.php on line 319
RuntimeException: Directory " C:\demo\example-app/C:\Users\username\AppData\Local\Temp/phel/cache/dependency-graph" was not created
in  C:\demo\example-app\vendor\gacela-project\gacela\src\Framework\Cache\FileCache.php:320 (gen:  C:\demo\example-app\vendor\gacela-project\gacela\src\Framework\Cache\FileCache.php:320)

#0  C:\demo\example-app\vendor\gacela-project\gacela\src\Framework\Cache\FileCache.php(63): Gacela\Framework\Cache\FileCache->ensureDirectory()
#1  C:\demo\example-app\vendor\phel-lang\phel-lang\src\php\Build\Infrastructure\Cache\DependencyTracker.php(32): Gacela\Framework\Cache\FileCache->__construct()
#2  C:\demo\example-app\vendor\phel-lang\phel-lang\src\php\Build\BuildFactory.php(184): Phel\Build\Infrastructure\Cache\DependencyTracker->__construct()
...
```

The odd part is how path for cache directory is produced, showing mixed up path separators and other weirdness:
```
RuntimeException: Directory " C:\demo\example-app/C:\Users\username\AppData\Local\Temp/phel/cache/dependency-graph" was not created
```

Workaround seems to be to disable the CompiledCodeCache by setting  `->isCompiledCodeCacheEnabled(false)` in `phel-config.php`.

PR #1466, opened 2026-04-19
ci: add windows-latest to compiler-tests matrix

Chemaclass:
## 🤔 Background

Issue #1459 (fixed in #1465) was a Windows-only path-handling bug that slipped through because CI only exercises `ubuntu-latest`. Native Windows users (non-WSL) hit it at runtime instead of at PR time.

## 💡 Goal

Run the PHPUnit compiler suite on `windows-latest` so regressions in path separators, temp-dir handling, `phar://` detection, line endings, and other POSIX-only assumptions are caught in CI.

## 🔖 Changes

- `.github/workflows/ci.yml` — `compiler-tests` job:
  - Adds `os` matrix dimension: `[ ubuntu-latest, windows-latest ]`.
  - `runs-on: ${{ matrix.os }}`.
  - `continue-on-error: ${{ matrix.os == 'windows-latest' }}` — Windows is advisory until the suite is fully green on that platform, so it does not block PRs.
  - `shell: bash` on the composer-cache step so the `$(...)` substitution and `>> $GITHUB_OUTPUT` redirection work under PowerShell default on Windows runners.

### Follow-ups (not in this PR)

- Flip `continue-on-error` off once Windows integration tests are stable.
- Consider adding `windows-latest` to `core-tests`, `examples`, `bash-tests` — those likely need `./bin/phel` shell-script work and POSIX-only fixture adjustments first.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,8 @@
+# Fixture files are compared byte-for-byte against compiler output;
+# force LF so Windows checkout (with core.autocrlf=true) does not turn
+# the expected `--PHP--` section into CRLF and break the diff.
+*.test                  text eol=lf
+
 /.claude                export-ignore
 /.github                export-ignore
 /adrs                   export-ignore
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -122,6 +122,7 @@ jobs:
         run: vendor/bin/phpunit --testsuite unit
 
       - name: Run integration tests to the compiler itself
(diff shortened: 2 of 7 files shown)


Chemaclass:
@JesusValeraDev please fix your @raspi-jesus-bot 🤣 
1. review only when the PR CI jobs are all green
2. approve only once per PR

Chemaclass:
this seems to be more complicated than expected. parking the idea to continue with other issues. Will come back another day

PR #1544, opened 2026-04-22
fix(core): make `binding` fiber-local and convey into futures

### LINKED ISSUE #1536: `binding` behavior with `future` differs from Clojure
*Phel v0.34.1-beta#2ee37d4 / PHP 8.4.16*

Would these be considered features of Phel/PHP or an unintended differences in behavior? REPL example:

```clojure
(def  ^:dynamic *x* :unset)
(defn test-fn [] *x*)

(require phel\async :refer [future]) ; required only for Phel

(let [f (future (test-fn))]
           (binding [*x* :now-here]
             @f))

;; Clojure => :unset
;; Phel => :now-here

(binding [*x* :outer]
           (let [f (future (test-fn))]
             (binding [*x* :inner]
               @f)))

;; Clojure => :outer
;; Phel => :inner

(binding [*x* :caller]
           (let [f (future
                     (binding [*x* :callee]
                       (future (test-fn))))]
             (binding [*x* :derefer]
               (let [derefed-f @f]
                 @derefed-f))))

;; Clojure => :callee
;; Phel => :derefer
```
Encountered in clojure-test-suite [binding.cljc](https://github.com/jasalt/clojure-test-suite/blob/74e3faa90e3ec7bf9c9bba1f28e4ded958fcb23e/test/clojure/core_test/binding.cljc#L57-L71)

JesusValeraDev:
Closes #1536

## 🤔 Background

`binding` in Phel mutated the global registry and then restored the old value in a `finally`. Because PHP fibers share that registry, a `future` body evaluated inside a later `binding` would see the *innermost* value at deref time instead of the one active where the future was created. The three snippets attached to #1536 produce `:now-here` / `:inner` / `:derefer` in Phel where Clojure produces `:unset` / `:outer` / `:callee`.

Two Clojure guarantees were missing:
- **Isolation** — concurrent threads (fibers, here) must not observe each other's `binding`s.
- **Conveyance** — `future`/`async` must capture the caller's dynamic bindings at creation time and reinstall them inside the worker.

## 💡 Goal

Make `binding` fiber-local for vars tagged `^:dynamic`, and snapshot/replay those bindings across `future`, `async`, `future-fiber`, and `future-call`, so the three issue snippets match Clojure's output. Preserve `with-redefs`-style semantics for non-dynamic vars so existing usage (`phel\mock`, ad-hoc redefs, etc.) is unaffected.

## 🔖 Changes

**Runtime**
- `src/php/Lang/DynamicScope.php` (new) — per-fiber stack of dynamic frames plus a per-fiber *recording* stack for the `binding` macro. `hasAnyBinding()` provides an O(1) probe for the hot `getDefinition` path.
- `src/Phel.php` — new `setVar`, `openBindingFrame`, `commitAndRunBindingFrame`, `abortBindingFrameIfOpen`, `snapshotDynamicBindings`, `withDynamicBindings`, `isDynamicVar`; overridden `getDefinition` short-circuits to the registry when no frame is active.

**Compiler**
- `SetVarEmitter` now emits `\Phel::setVar(...)`. Outside a `binding`, this still mutates the registry; inside one, it stages into the pending fiber-local frame.
- `DefSymbol` merges metadata attached to the *name symbol* (`(def ^:dynamic *x* 1)` used to be a no-op for `:dynamic` because the analyzer only read list meta).

**Phel stdlib**
- `binding` macro rewritten in `src/phel/core/io.phel` around `openBindingFrame` / `commitAndRunBindingFrame` with a `try/finally` abort guard.
- `future`, `async`, `future-fiber`, `future-call` in `src/phel/async.phel` snapshot caller bindings and wrap the fiber body in `withDynamicBindings`.
- `phel\ai/*http-post*`, `phel\test/*current-test-name*`, `phel\test/*testing-contexts*` are now tagged `^:dynamic` so concurrent fibers that rebind them stay isolated. `phel\mock` intentionally stays on the legacy mutation path because it rebinds arbitrary user-supplied function symbols (not typically multi-fiber).

**Tests**
- `tests/php/Unit/Lang/DynamicScopeTest.php` (new) — 13 unit tests covering push/pop/snapshot, fiber isolation, recording stack, exception safety, and the `hasAnyBinding` fast-path probe.
- `tests/phel/test/core/binding-conveyance.phel` (new) — 6 tests covering the three issue snippets, concurrent-fiber isolation, `future-fiber` conveyance, and non-dynamic `with-redefs` fallback.
- Updated `tests/phel/test/async.phel` macroexpand assertion and the `set-var` emission fixture.

### Performance
Measured with `phpbench` (10 iterations cold-start, PHP 8.5.4, opcache off, fresh cache each run):

| Bench | main | this PR | Δ |
|---|---|---|---|
| `bench_phel_run` (boot `phel\core` + deps) | 19.221ms ±1.43% | 20.263ms ±2.43% | +5.4% |
| `bench_core_file_compilation` | 23.748ms ±1.23% | 23.941ms ±0.74% | +0.8% (in noise) |

Every global-var read now pays one `Fiber::getCurrent()` + one `isset` probe via `DynamicScope::hasAnyBinding()` before falling through to the registry on the empty-stack fast path. The `bench_phel_run` delta reflects the fiber-local correctness cost; compilation is untouched.

### Test plan
- [x] `./vendor/bin/phpunit --filter DynamicScopeTest` (13/13)
- [x] `./bin/phel test tests/phel/test/core/binding-conveyance.phel` (9/9)
- [x] `composer test-compiler` — all green except the pre-existing `DataReadersAutoloadTest` flake (reproduces on `main`)
- [x] `composer test-core` — 4250/4250 green (the pre-existing `math-operation.phel` compile error also reproduces on `main`)
- [x] `composer phpstan`, `composer psalm`, `composer rector`, `composer csrun` — all green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+#### Core
+- `binding` on vars tagged `^:dynamic` now uses fiber-local storage, so concurrent fibers observe independent values and child futures/async blocks inherit the bindings active at their creation (Clojure-aligned binding conveyance). `^:dynamic` metadata on the name symbol of `def` is now honored. [#1536](https://github.com/phel-lang/phel-lang/issues/1536)
+
 #### Build
 - Directory scan skips unparseable `.phel` files instead of aborting; REPL boots even when the cwd tree contains malformed Phel sources
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 
 #### Core
-- `binding` on vars tagged `^:dynamic` now uses fiber-local storage, so concurrent fibers observe independent values and child futures/async blocks inherit the bindings active at their creation (Clojure-aligned binding conveyance). `^:dynamic` metadata on the name symbol of `def` is now honored. [#1536](https://github.com/phel-lang/phel-lang/issues/1536)
+- `binding` on `^:dynamic` vars is fiber-local; `future`/`async` convey caller bindings; `^:dynamic` on `def` name symbols is honored (#1536)
 
 #### Build
 - Directory scan skips unparseable `.phel` files instead of aborting; REPL boots even when the cwd tree contains malformed Phel sources
--- a/src/Phel.php
+++ b/src/Phel.php
@@ -64,8 +64,11 @@ final class Phel extends InternalPhel
      */
     public static function getDefinition(string $ns, string $name): mixed
     {
+        // Hot path: every resolved symbol hits this. Probe the scope for
+        // *any* live frame first (O(1)) so the common "no binding active"
+        // case skips the string concat and stack walk entirely.
         $scope = DynamicScope::getInstance();
-        if ($scope->hasBinding($ns, $name)) {
+        if ($scope->hasAnyBinding() && $scope->hasBinding($ns, $name)) {
             return $scope->getBinding($ns, $name);
         }
 
--- a/src/Phel.php
+++ b/src/Phel.php
@@ -6,6 +6,7 @@ use Phel\Lang\Collections\HashSet\PersistentHashSetInterface;
 use Phel\Lang\Collections\LinkedList\PersistentListInterface;
 use Phel\Lang\Collections\Map\PersistentMapInterface;
 use Phel\Lang\Collections\Vector\PersistentVectorInterface;
+use Phel\Lang\DynamicScope;
 use Phel\Lang\Keyword;
 use Phel\Lang\Registry;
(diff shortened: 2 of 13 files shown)


JesusValeraDev:
Re: `array_any` concern — `array_any`/`array_all`/`array_find`/`array_find_key` were [introduced in PHP 8.4.0](https://www.php.net/manual/en/function.array-any.php), and this project's minimum is `>=8.4` (composer.json). The rule also ships via rector (`AddArrayAnyRector`) and is already used elsewhere (`ProjectCompiler`, `NodeEnvironment`, `CombineGenerator`, etc.). Safe as-is — compiler tests passed on both PHP 8.4 and 8.5 in CI.

Chemaclass (approved):
Awesome work! Thanks

PR #1545, opened 2026-04-22
feat(core): move `future` macro from `phel\async` to `phel\core`

### LINKED ISSUE #1537: Move `future` from `phel\async` to core
In [clojure-test-suite binding.cljc](https://github.com/jasalt/clojure-test-suite/blob/74e3faa90e3ec7bf9c9bba1f28e4ded958fcb23e/test/clojure/core_test/binding.cljc#L4), `future` needs to be required from `phel\async` with reader conditional added to ns form.

Clojure has `future` available without explicitly requiring it, so it would be neat if Phel did the same.

JesusValeraDev:
## 🤔 Background

Closes #1537.

In Clojure, `future` is available in `clojure.core` without any explicit `require`. In Phel it lived in `phel\async`, so `.cljc` sources targeting both languages needed a reader conditional just to pull it in (see the [clojure-test-suite `binding.cljc` case](https://github.com/jasalt/clojure-test-suite/blob/74e3faa90e3ec7bf9c9bba1f28e4ded958fcb23e/test/clojure/core_test/binding.cljc#L4)).

## 💡 Goal

Make `future` available globally — no `(:require phel\async :refer [future])` needed — to match Clojure's out-of-the-box availability.

## 🔖 Changes

- Add `src/phel/core/futures.phel` with the (AMPHP-backed) `future` macro and wire it into `core.phel` so it loads with every namespace.
- Remove the `future` macro from `src/phel/async.phel`; the module still owns `async`, `await`, `delay`, `await-all`, `await-any`, `pmap`, `future-cancel`, `future-cancelled?`, `future-done?`, plus the fiber-backed primitives (`promise`, `deliver`, `future-call`, `future-fiber`, `future?`).
- Drop `future` from the `:refer` lists in `tests/phel/test/async.phel` and `tests/phel/test/core/binding-conveyance.phel` (the global binding is used instead).
- Add `tests/phel/test/core/future-in-core.phel` asserting the macro is reachable without requiring `phel\async` and still expands to a `PhelFuture` wrapper around `amp\async`.
- Update `CHANGELOG.md`, `docs/clojure-migration.md`, and the `docs/async-guide.md` cancel-on-error example to reflect the new home.

### DIFF
--- a/docs/clojure-migration.md
+++ b/docs/clojure-migration.md
@@ -176,7 +176,7 @@ Phel runs on PHP. A handful of Clojure features don't translate directly:
 | **Vars (Clojure sense)** | PHP has no thread-local bindings | `def` creates namespace-level bindings directly |
 | **`alter-var-root`** | No first-class vars to re-root | Use an `atom` with `swap!` for mutable state, or redefine the top-level binding with `def`. Calling `alter-var-root` at runtime throws `BadMethodCallException` with this hint |
 
-Phel does provide `future`, `future-cancel`, and `pmap` via the `phel\async` module, which uses AMPHP fibers. Semantics match Clojure's `future` where they can (including timeout-bounded `deref`), but cancellation is cooperative rather than thread-interrupt. For top-level scripting without an event loop, use the fiber primitives (`promise`, `deliver`, `future-call`, `future-fiber`); see [docs/async-guide.md](async-guide.md) for the decision guide.
+Phel provides `future` in `phel\core` (available without requiring any namespace, as in Clojure); `future-cancel` and `pmap` still live in `phel\async`. The implementation uses AMPHP fibers. Semantics match Clojure's `future` where they can (including timeout-bounded `deref`), but cancellation is cooperative rather than thread-interrupt. For top-level scripting without an event loop, use the fiber primitives (`promise`, `deliver`, `future-call`, `future-fiber`); see [docs/async-guide.md](async-guide.md) for the decision guide.
 
 ## Structural differences
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 
 - **BREAKING**: Minimum PHP version bumped from 8.3 to 8.4. PHP 8.3 is no longer supported.
 
+#### Core
+- `future` is now available from `phel\core` without requiring `phel\async`, matching Clojure's out-of-the-box availability (#1537)
+
 ### Fixed
 
 #### Core
--- a/docs/async-guide.md
+++ b/docs/async-guide.md
@@ -254,7 +254,7 @@ Wall time tracks the slowest branch, not the sum.
 
 ```phel
 (ns demo\cancel-on-error
-  (:require phel\async :refer [async await delay future future-cancel]))
+  (:require phel\async :refer [async await delay future-cancel]))
 
 (defn launch []
   (async
--- a/src/phel/async.phel
+++ b/src/phel/async.phel
@@ -3,8 +3,10 @@
 ;; Two cooperating concurrency layers in one module.
 ;;
 ;; - AMPHP-backed: `async`, `await`, `delay`, `await-all`, `await-any`,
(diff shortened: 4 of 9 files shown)


docs/clojure-migration.md
```diff
@@ -176,7 +176,7 @@ Phel runs on PHP. A handful of Clojure features don't translate directly:
...
 
-Phel does provide `future`, `future-cancel`, and `pmap` via the `phel\async` module, which uses AMPHP fibers. Semantics match Clojure's `future` where they can (including timeout-bounded `deref`), but cancellation is cooperative rather than thread-interrupt. For top-level scripting without an event loop, use the fiber primitives (`promise`, `deliver`, `future-call`, `future-fiber`); see [docs/async-guide.md](async-guide.md) for the decision guide.
+Phel provides `future` in `phel\core` (available without requiring any namespace, as in Clojure); `future-cancel` and `pmap` still live in `phel\async`. The implementation uses AMPHP fibers. Semantics match Clojure's `future` where they can (including timeout-bounded `deref`), but cancellation is cooperative rather than thread-interrupt. For top-level scripting without an event loop, use the fiber primitives (`promise`, `deliver`, `future-call`, `future-fiber`); see [docs/async-guide.md](async-guide.md) for the decision guide.
```
jasalt, line 179: Going this way, it would make sense that most of these functions were be moved to `phel\core` that on Clojure side lives in the core namespace too so there would be less surprises.

There's not too many Phel specific ones (e.g. `await-all`, `future-fiber`, `async` ...) that are not found in Clojure, so I'm not sure if it's worth having a separate async namespace for them either. 🤔

For reference there is a [`clojure.core.async`](https://clojuredocs.org/clojure.core.async) namespace also that contains higher level utilities, but the more primitive ones live in the `clojure.core`.

JesusValeraDev: Great point — I agree in principle. `promise`, `deliver`, `pmap`, `future-call`, `future-cancel`, `future-cancelled?`, `future-done?`, and `future?` all live in `clojure.core`, so requiring `phel\async` for them defeats the same `.cljc` portability goal that motivated #1537.

I'd rather not expand the scope of this PR though — `future` is the one the `clojure-test-suite` `binding.cljc` case needed unqualified, so landing it narrowly keeps the change reviewable. One wrinkle worth flagging for the follow-up: Phel's `delay` is time-based sleep, but `clojure.core/delay` is a lazy-computation wrapper — so that one is a deliberate divergence and can't just be promoted to core without either renaming or shadowing the Clojure meaning.

Opened #1548 to track the broader move (including the "is `phel\async` still worth having?" question) with acceptance criteria. Happy to pick it up after this one lands.

Chemaclass: I agree with you two, thank you guys!

jasalt: @JesusValeraDev yes, good points. There's some trickyness with the differences. The `pmap` also has implementation difference in Phel being single-thread based currently v.s. Clojure where it's actually parallel, though it's not wholly different semantics as `delay` v.s. `clojure.core/delay` that I didn't on a quick glance register... 

Would there be then an opportunity to make Phel `delay` closer to `clojure.core/delay` semantics or where the difference with it stands? I haven't used Phel `delay` so far.

Chemaclass (approved):

Chemaclass (approved):

docs/clojure-migration.md
JesusValeraDev: Good catch on `pmap` — you're right that it's a semantics divergence too, just a quieter one than `delay`. I'll fold both into #1548 so they're classified together rather than `pmap` sitting in the "just move it" bucket.

On `delay`: I haven't dug into whether aligning with `clojure.core/delay` (lazy computation wrapper) is feasible without breaking current users of the sleep-based version. That's exactly the call #1548 acceptance criterion #2 is meant to force — rename, realign, or consciously diverge. Happy to take your preference into account there; my lean is rename the sleep version (e.g. `sleep`) and reserve `delay` for the Clojure-aligned lazy wrapper, but I'd like to hear yours before committing.

Issue #1567, opened 2026-04-24
feat: Deprecate backslash (\) as namespace separator in favor of Clojure-compatible dot (.)


JesusValeraDev:
## Context

Phel historically uses PHP-style backslashes as the namespace separator in every context: namespace declarations (`(ns phel\core)`), requires (`(:require phel\walk ...)`), fully-qualified references (`phel\core/map`), class FQNs (`\Phel\Lang\ExInfoException`), etc.

Clojure uses dots: `clojure.core`, `clojure.core/map`, `java.lang.String`.

Recent work has made the dot form **equivalent** in several positions:

- #1177 — dot accepted as alternate namespace separator
- #1251 — dot separator in qualified FQN references at resolution time
- #1207 — auto-alias `clojure.*` namespaces to `phel.*`
- #1553 (this thread) — bare `Foo.Bar.Baz` resolves as PHP class FQN alias

The two syntaxes now coexist. For full Clojure compatibility and to produce one unambiguous style, the backslash form should be **deprecated** and eventually **removed**.

## Goal

Emit a one-time `E_USER_DEPRECATED` warning per file per offending pattern when user code uses `\` as a namespace separator, pointing at the dot replacement. Keep behavior backward compatible for at least one minor release. Remove the backslash form in a later major/minor bump.

## Scope of the deprecation

Positions that need to flag `\`-separated syntax written in **user** code:

1. `ns` forms — `(ns phel\foo)` → suggest `(ns phel.foo)`
2. `:require` / `:use` / `:refer` clauses — `(:require phel\walk)` → `(:require phel.walk)`
3. Fully-qualified refs at call sites — `phel\core/map` → `phel.core/map`
4. Class FQNs — `\Phel\Lang\ExInfoException` → `Phel.Lang.ExInfoException`
5. `load` forms (if they accept a namespace)
6. Reader-macro and quoting forms that carry namespace strings

Positions that must **not** warn:

- Phel's own stdlib — `phel\core`, `phel\walk`, `phel\string`, `phel\html`, `phel\http`, `phel\json`, `phel\test` — the stdlib is written in `\` form and boots the compiler itself; warnings there would fire on every CLI invocation
- Test fixtures that exercise the deprecated form deliberately
- Mungesed PHP class names (e.g. generated code the emitter writes back out)

## Proposed design

[...]

PR #1570, opened 2026-04-24
feat(stdlib): rewrite ns/require forms to Clojure-compatible dot syntax

JesusValeraDev:
## 🤔 Background

The backslash-deprecation work in Phase 1a / Phase 1b ([#1568](https://github.com/phel-lang/phel-lang/pull/1568), [#1569](https://github.com/phel-lang/phel-lang/pull/1569)) points users at the dot form, but the Phel stdlib itself was still written in backslash form. That left new users confused: "you tell me to write `phel.core`, but your own source says `phel\core`?". This PR converts the stdlib so the same syntax is idiomatic everywhere.

Closes the stdlib-rewrite sub-task of [#1567](https://github.com/phel-lang/phel-lang/issues/1567).

## 💡 Goal

Every `ns`, `in-ns`, and `:require` form under `src/phel/**/*.phel` uses dot syntax; PHP-interop forms (`:use \Phel\Lang\Foo`) stay untouched because `\` is the native PHP FQN separator there.

## 🔖 Changes

- **57 `.phel` files** under `src/phel/` updated — `(ns phel\foo)` → `(ns phel.foo)`, `(:require phel\walk ...)` → `(:require phel.walk ...)`, `[phel\walk :as w]` → `[phel.walk :as w]`, and a handful of comments/docstrings referencing `phel\test/assert-expr` style call sites.
- `BackslashSeparatorDeprecator::isPhelStdlibSource()` removed and the corresponding call site stripped. With the stdlib clean, the path-based suppression is no longer needed and would have been a false negative for any user project whose own sources happened to live under a path ending in `/src/phel/`.
- Matching unit test `test_suppresses_warnings_from_phel_stdlib_sources` removed.

## What is NOT changed

- `(:use \Phel\Lang\SourceLocationInterface)` and similar PHP class imports — the backslash syntax is legitimate PHP FQN there, and changing it would couple the decision to `:use` form redesign (out of scope; tracked in #1567).
- String literals that contain `phel\cli:`, `phel\n`, or `.phel` — those are data/messages, not namespace references.

## Validation

- `composer test-compiler` — 2707 tests pass; only the pre-existing `DataReadersAutoloadTest` flake (also fails on main)
- `composer test-core` — 4313/4313 pass
- `PHEL_WARN_DEPRECATIONS=1 ./bin/phel run <probe>` emits zero deprecations during stdlib boot — proof the stdlib is truly clean without needing a suppression allowlist
- PHPStan, rector, cs-fixer (risky) — clean

## Merge order

Independent of #1569. If #1569 merges first there will be a trivial conflict on `BackslashSeparatorDeprecator.php` (singleton constructors vs. suppression removal) which should auto-resolve.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 #### Compiler
 - Dot-separated bare symbols that look like class FQNs (e.g. `Phel.Lang.ExInfoException`) now resolve as aliases for `\Phel\Lang\ExInfoException`, improving `.cljc` source portability with Clojure and sibling dialects (#1553)
 - Opt-in deprecation warning for backslash (`\`) as namespace separator. Covers call sites, class FQNs, `ns` declarations, and `:require` targets (flat + vector forms). Set `PHEL_WARN_DEPRECATIONS=1` to surface migration targets. Dot (`.`) is the Clojure-compatible form and will be the only supported form in a future release — see `docs/migration/backslash-to-dot.md` (#1567)
+- Phel stdlib sources (`src/phel/**/*.phel`) rewritten to use Clojure-compatible dot syntax in `ns`, `in-ns`, and `:require` forms; the stdlib-source suppression in the backslash deprecator was dropped as it is no longer needed (#1567)
 
 #### Core
 - Hierarchy functions `isa?`, `derive`, `underive`, `parents`, `ancestors`, `descendants` accept an optional hierarchy argument; the hierarchy arities of `derive`/`underive` are pure and return a new hierarchy (#1543)
--- a/src/phel/ai.phel
+++ b/src/phel/ai.phel
@@ -1,6 +1,6 @@
-(ns phel\ai
-  (:require phel\http-client :as hc)
-  (:require phel\json)
+(ns phel.ai
+  (:require phel.http-client :as hc)
+  (:require phel.json)
   (:use RuntimeException))
 
 ;; -----------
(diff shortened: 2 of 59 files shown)


JesusValeraDev:
Both findings were real and are now fixed in aeba76f8 — thanks for catching them.

**1. Sub-namespaces fully converted.** The initial sed pass only matched the leading `phel\`, so nested segments kept the backslash. Ran targeted replacements for `phel.schema\` and `phel.test\` on the affected files. Every `ns`, `in-ns`, and `:require` under `src/phel/` is now uniformly dotted. Spot-check:

```clojure
(ns phel.schema.coercer
  (:require phel.core)
  (:require phel.schema.registry :as reg)
  (:require phel.schema.validator :as v))
```

**2. `apropos` docstring corruption.** The `[phel\` → `[phel.` pattern reached inside a string literal and left one symbol as `phel.\core/flat-map` while the other stayed `phel\\core/map`. Restored both to the backslash form:

```clojure
{:example "(apropos \"map\") ; => [phel\\core/flat-map phel\\core/map ...]"}
```

I kept the docstring examples in backslash form intentionally because the Registry still stores namespace keys with `\` internally — `apropos`, `ns-list`, `find-fn`, etc. will return symbols like `phel\core/map` at runtime today. Rewriting the example to dots would misrepresent the actual function output. When the backslash form is removed from the compiler entirely (tracked under #1567), those runtime-facing docstrings will update in the same sweep.

Validation after the fix:
- `composer test-compiler` — 2706 pass (only the pre-existing `DataReadersAutoloadTest` flake that fails on main)
- `composer test-core` — 4313/4313 pass
- `PHEL_WARN_DEPRECATIONS=1 ./bin/phel run <probe>` — zero warnings from stdlib boot

Chemaclass (approved):

PR #1573, opened 2026-04-24
feat(console): add --warn-deprecations CLI flag

JesusValeraDev:
## 🤔 Background

The backslash-separator deprecation warnings (#1568 / #1569) previously required `PHEL_WARN_DEPRECATIONS=1` env var. Works fine shell-wide but is clunky in CI configs where you want to flip it on just for a single command. Follow-up in the [#1567 issue](https://github.com/phel-lang/phel-lang/issues/1567) tracker calls for a proper CLI flag.

Related to #1567.

## 💡 Goal

`phel <command> --warn-deprecations` turns on the warnings for that invocation, without per-command wiring.

## 🔖 Changes

- New `Phel\Console\Application\WarnDeprecationsFlag::applyAndStrip(array $argv): array` — scans argv for `--warn-deprecations` (both plain and `=1` forms), configures the `BackslashSeparatorDeprecator` singleton, and returns argv with the flag stripped so Symfony per-command parsers do not complain about an unknown option
- `ConsoleBootstrap::run()` calls the helper after argv sanitization — the flag is consumed globally, so it works with every subcommand (`run`, `test`, `lint`, `compile`, `repl`, etc.)
- `BackslashSeparatorDeprecator` emits via plain `trigger_error` (dropped the `@`) so the default PHP error display surfaces the warnings without needing a user-installed error handler — matches user expectation when they turn the flag on
- Unit test coverage: flag-absent passthrough, plain flag stripping + enabling, `=1` value form, explicit flag-absent silence
- Migration guide (`docs/migration/backslash-to-dot.md`) now documents both the flag and the env var

## End-to-end probe

```bash
$ cat /tmp/probe/src/probe.phel
(ns probe (:require phel\walk))
(println (phel\walk/keywordize-keys {"a" 1}))

$ ./bin/phel run --warn-deprecations src/probe.phel
Deprecated: Backslash ('\') namespace separator in symbol 'phel\walk' at .../probe.phel:2 ...
Deprecated: Backslash ('\') namespace separator in symbol 'phel\walk/keywordize-keys' at .../probe.phel:3 ...
{:a 1}

$ ./bin/phel run src/probe.phel         # flag absent → silent
{:a 1}
```

## Validation

- `./vendor/bin/phpunit tests/php/Unit/Compiler/Analyzer/Environment/ tests/php/Unit/Console/` — 92 tests pass
- `composer test-compiler` — 2714 pass; only the pre-existing `DataReadersAutoloadTest` flake
- PHPStan, rector, cs-fixer (risky) — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,44 +7,43 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 #### Compiler
-- Dot-separated bare symbols that look like class FQNs (e.g. `Phel.Lang.ExInfoException`) now resolve as aliases for `\Phel\Lang\ExInfoException`, improving `.cljc` source portability with Clojure and sibling dialects (#1553)
-- `--warn-deprecations` CLI flag on the `phel` console turns on the backslash-separator deprecation warnings without needing the `PHEL_WARN_DEPRECATIONS=1` env var; strip-and-enable happens at bootstrap so every downstream command receives the cleaned argv (#1567)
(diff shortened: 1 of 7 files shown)


Chemaclass (approved):
Cool

PR #1582, opened 2026-04-25
feat(core): move clojure-parity async surface into phel.core

### LINKED ISSUE #1548: Evaluate moving Clojure-parity async functions from `phel\async` to `phel\core`
## 🤔 Background

Follow-up from [#1545 review](https://github.com/phel-lang/phel-lang/pull/1545#discussion_r3128343501) (cc @jasalt).

With `future` moving to `phel\core` (closes #1537), most of the remaining functions in `phel\async` are direct Clojure-parity APIs that live in `clojure.core` on the Clojure side. Forcing users to `(:require phel\async :refer [...])` for them creates the same friction for `.cljc` portability that motivated #1537 in the first place.

## 💡 Goal

Decide whether the Clojure-parity subset of `phel\async` should also move to `phel\core` (as extra files in `src/phel/core/`, same pattern used for `futures.phel`), and whether `phel\async` should keep existing at all.

## 🔖 Functions to classify

**In `clojure.core` on the Clojure side (candidates to move):**

- `promise`
- `deliver`
- `pmap`
- `future-call`
- `future-cancel`
- `future-cancelled?`
- `future-done?`
- `future?`

**Phel-specific (would stay wherever the namespace lands):**

- `async`, `await`, `await-all`, `await-any` — Phel-specific fiber combinators
- `future-fiber` — Phel-specific, pairs with `future-call` for top-level-without-event-loop use
- `->closure` — Phel→PHP Closure bridge

**Deliberate divergence (must stay where it is to avoid conflicting with Clojure semantics):**

- `delay` — Phel's version is time-based sleep; `clojure.core/delay` is a lazy computation wrapper. Promoting this to core would shadow the expected Clojure meaning. Needs either a rename or to stay in `phel\async`.

## ✅ Expectations / Acceptance criteria

1. **Decision recorded** on whether to:
   - (a) Move the Clojure-parity subset to `phel\core` and keep `phel\async` for the Phel-specific remainder.
   - (b) Collapse `phel\async` entirely into core (accepting that every namespace transitively loads `\Phel\Fiber\FiberFacade` plumbing).
   - (c) Do nothing — `future` was special because Clojure-portable `.cljc` code genuinely needs it unqualified, the rest can keep the explicit require.
2. **`delay` resolved** — either renamed (e.g. `sleep`), left in `phel\async`, or consciously kept as a Phel-specific override of the Clojure name with a docstring note.
3. **No regressions in `docs/async-guide.md`** — the "AMPHP vs fiber layer" decision guide must still make sense after the move.
4. **CHANGELOG** entry under `## Unreleased` describing the migration; migration note added to `docs/clojure-migration.md` table.
5. **Test impact** — `:refer` lists in `tests/phel/test/async.phel`, `tests/phel/test/core/binding-conveyance.phel`, and any other test touching these symbols are updated; a `future-in-core.phel`-style test is added per moved function asserting it is reachable without `(:require phel\async)`.
6. **Load-order safety** — `phel\core`'s entry already pulls `futures.phel`; verify that chaining the additional files doesn't introduce circular dependencies or force the fiber facade to load when it isn't used.

## ⚠️ Non-goals

- Renaming or redesigning the functions themselves.
- Changing semantics of any listed function.
- Touching `core.async`-style higher-level utilities (none exist in Phel yet).

## 📎 References

- #1537 — original motivation for moving `future`.
- #1545 — PR that moved `future` and surfaced this discussion.
- [`clojure.core.async`](https://clojuredocs.org/clojure.core.async) — referenced by the reviewer as the analogue for higher-level async, while primitives live in `clojure.core`.

JesusValeraDev:
## 🤔 Background

Follow-up from #1548 (and the [PR #1545 review thread](https://github.com/phel-lang/phel-lang/pull/1545#discussion_r3128343501)). Now that `future` already lives in `phel\core` (#1537), the rest of the Clojure-parity async surface — primitives that are in `clojure.core` on the Clojure side — still required `(:require phel\async :refer [...])` to use, which created the same `.cljc` portability friction the earlier move was meant to remove.

The issue's discussion converged on **option (b)**: collapse the Clojure-parity (and pairing fiber-combinator) symbols into core, leaving `phel\async` as the home of the one symbol whose semantics intentionally diverge from Clojure (`delay`).

## 💡 Goal

Make portable Phel code work without remembering which primitive lives where, while keeping the one knowingly-divergent symbol explicit.

## 🔖 Changes

- New `src/phel/core/async.phel` (loaded after `core/futures.phel` from `src/phel/core.phel`) hosts the moved surface:
  - AMPHP layer: `async`, `await`, `await-all`, `await-any`, `pmap`, `future-cancel`, `future-cancelled?`, `future-done?`
  - Fiber layer: `promise`, `deliver`, `future-call`, `future-fiber`, `future?`
  - Bridge: `->closure`
- `phel\async` slimmed to a single export, `delay`. **No rename to `sleep`** — applied @jasalt's feedback that it would clash with `php/sleep`. Docstring expanded to clarify (a) top-level vs fiber behaviour, (b) that this is *not* `clojure.core/delay`, and (c) when to reach for `php/sleep` instead.
- `pmap` docstring now explicitly states it is single-threaded (overlaps IO-bound work, does not parallelize CPU), matching Basilisp/ClojureScript and avoiding the implicit promise of `clojure.core/pmap`'s thread-pool semantics.
- `tests/phel/test/core/async-in-core.phel` added — one assertion per moved capability, mirroring the `future-in-core.phel` style from #1537. Existing `tests/phel/test/async.phel`, `tests/phel/test/async-fiber.phel`, and `tests/phel/test/core/binding-conveyance.phel` updated to drop the now-unnecessary `:refer` lists.
- Docs: `docs/async-guide.md`, `docs/clojure-migration.md`, `docs/examples/11_async-concurrency.phel`, and the Fiber module's `CLAUDE.md` updated for the new home.
- CHANGELOG entry under `## Unreleased` → `Changed` → `Core` flagged **BREAKING** for users currently `:refer`ing the moved symbols from `phel\async` (the fix is just to drop them from the require list).

## ✅ Acceptance criteria mapping

1. **Decision recorded** — option (b) with `delay` consciously kept in `phel\async`.
2. **`delay` resolved** — kept in `phel\async`, no rename, docstring expanded per @jasalt's review.
3. **`docs/async-guide.md`** — AMPHP-vs-fiber decision guide preserved; namespace location notes updated.
4. **CHANGELOG + migration note** — both updated.
5. **Tests** — moved-symbol `:refer`s dropped, `async-in-core.phel` regression file added (per-feature, like `future-in-core.phel`).
6. **Load-order safety** — `core/async.phel` loaded after `core/futures.phel`; `\Phel\Fiber\FiberFacade` is autoloaded only on first call, so namespaces that don't use these primitives still pay nothing at runtime.

## 🧪 Verification

- `composer test-core`: 4352/4352 ✅ (+9 new vs main)
- `composer test-compiler`: only the pre-existing `DataReadersAutoloadTest` flake (also fails on `main`)
- `composer csrun`, `composer phpstan`, `composer rector`: ✅
- `./bin/phel run docs/examples/11_async-concurrency.phel`: ✅ runs end-to-end

## ⚠️ Breaking change

Code currently doing `(:require phel\async :refer [async await pmap …])` will need to drop those symbols from the `:refer` list (they're available unqualified). The only symbol that still needs `(:require phel\async :refer [delay])` is `delay`. Documented in CHANGELOG and `docs/clojure-migration.md`.

Closes #1548.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ All notable changes to this project will be documented in this file.
 
 #### Core
 - `future` now available from `phel\core` without requiring `phel\async` (#1537)
+- **BREAKING**: `phel\async`'s Clojure-parity surface moved to `phel\core` so `.cljc` portable code mirrors `clojure.core`'s defaults (#1548). The following are now reachable without a require: `async`, `await`, `await-all`, `await-any`, `pmap`, `promise`, `deliver`, `future-call`, `future-fiber`, `future?`, `future-cancel`, `future-cancelled?`, `future-done?`, `->closure`. Update existing `(:require phel\async :refer [...])` clauses to drop these symbols. `delay` deliberately stays in `phel\async` because Phel's `delay` is a sleep primitive while `clojure.core/delay` is a lazy-thunk wrapper. `pmap`'s docstring now calls out that it is single-threaded (overlaps IO, does not parallelize CPU), matching Basilisp/ClojureScript
 
 ### Fixed
 
--- a/docs/async-guide.md
+++ b/docs/async-guide.md
@@ -1,6 +1,6 @@
 # Async Module Guide
 
-`phel\async` exposes two cooperating concurrency layers over PHP fibers. Both live in the same namespace so you can mix them, but each has its own best use.
+Phel ships two cooperating concurrency layers over PHP fibers, both available **from `phel.core` without an explicit require** so `.cljc` portable code mirrors `clojure.core`'s defaults. The single exception is `delay`, which stays in `phel.async` because Phel's `delay` is a sleep primitive while `clojure.core/delay` is a lazy-thunk wrapper — keeping the names separated avoids silently hijacking Clojure semantics.
 
 ## Contents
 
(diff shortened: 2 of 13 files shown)


Chemaclass (approved):
nice!

PR #1691, opened 2026-04-28
feat: Improve default test reporter output

jasalt:
## 🤔 Background

Default test report is not as informative as it could be.

## 💡 Goal

Adds missing information and applies minor formatting to make test reports more convenient.

## 🔖 Changes

- Include the test name
- Upper case headline prefix
- Rename "Test:" to ~~"Assert:"~~ "Form:" for clarity
- Rename "threw exception:" to "threw:"
- Show test file basename instead of absolute path (absolute path is too verbose)
- Show exception's message in addition to it's class if it's set

Modified formatting loosely follows default Clojure test runner.

### Old format:
```
~~~~~~~~~~
Failure in 'special symbols' (/home/user/clojure-test-suite/test/clojure/core_test/special_symbol_qmark.cljc:5)
                 Test: (quote var)
         evaluated to: var
  but doesn't satisfy: special-symbol?
~~~~~~~~~~
Error in 'records' (/home/user/clojure-test-suite/test/clojure/core_test/dissoc.cljc:5)
              Test: (= {:b 2, :c nil} (apply dissoc r [:a]))
   threw exception: Error
~~~~~~~~~~
Failure in 'metadata' (/home/user/clojure-test-suite/test/clojure/core_test/atom.cljc:6)
          Test: (meta (atom nil :meta (sorted-map :a "a")))
  evaluated to: nil
    but is not: = to {:a "a"}
~~~~~~~~~~
```
### New format:
```
~~~~~~~~~~
FAIL in (test-special-symbol? 'special symbols') (special_symbol_qmark.cljc:5)
                 Form: (quote var)
         evaluated to: var
  but doesn't satisfy: special-symbol?
~~~~~~~~~~
ERROR in (test-dissoc 'records') (dissoc.cljc:5)
              Form: (= {:b 2, :c nil} (apply dissoc r [:a]))
             threw: Error
      with message: Cannot use object of type Phel\Lang\AbstractFn@anonymous as array
~~~~~~~~~~
FAIL in (test-atom 'metadata') (atom.cljc:6)
          Form: (meta (atom nil :meta (sorted-map :a "a")))
  evaluated to: nil
    but is not: = to {:a "a"}
~~~~~~~~~~
```

### DIFF
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -485,29 +485,29 @@
   (let [{:exception exception :form form} data
         command-facade                    (php/new CommandFacade)
         printer                           (php/-> command-facade (getExceptionPrinter))]
-    (println "            Assert:" (readable-str form))
+    (println "              Form:" (readable-str form))
     (println "             threw:" (php/get_class exception))
     (when-let [exception-message (ex-message exception)]
       (println "      with message:" exception-message))
     (php/-> printer (printStackTrace exception))))
 
 (defn- print-predicate-failure [{:pred pred :value value :result result}]
-  (println "               Assert:" (readable-str value))
+  (println "                 Form:" (readable-str value))
   (println "         evaluated to:" (readable-str result))
   (println "  but doesn't satisfy:" (readable-str pred)))
 
 (defn- print-predicate-not-failure [{:pred pred :value value :result result}]
-  (println "            Assert:" (readable-str value))
+  (println "              Form:" (readable-str value))
   (println "      evaluated to:" (readable-str result))
   (println "  but does satisfy:" (readable-str pred) " (it shouldn't)"))
 
 (defn- print-binary-failure [{:pred pred :expected expected :actual actual :result result}]
-  (println "        Assert:" (readable-str actual))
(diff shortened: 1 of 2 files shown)


jasalt:
Question, is the exception stacktrace reporter functioning correct? There's existing call to `(php/-> printer (printStackTrace exception))` but I haven't seen any stack traces in reports before.

https://github.com/phel-lang/phel-lang/blob/54396ab0d55558bd231db83cb73ea4d833e079e9/src/phel/test.phel#L492

Regardless, full stack trace would not be useful too often and takes space so it would be better behind `-v` or `--stack-trace=full` flag or something.

jasalt:
The ”Assert:” here is not accurate.
I think ”Form:” would be more accurate.
Assert should include the full ’is’ form.
Editing…

Chemaclass:
I agree, we can improve when to actually display the full stack trace, which should be better behind an option or similar and not by default

JesusValeraDev (approved):

Chemaclass (approved):

PR #1707, opened 2026-04-28
fix(sorted-map): adapt bool predicates to comparator contract

### LINKED ISSUE #1705: `sorted-map-by` with `>` fails
*phel v0.34.1-beta#03b13c2*

```
FAIL in (test-eq 'sorted collections') (eq.cljc:5)
          Form: (sorted-map-by > 13 :a 14 :b 15 :c)
  evaluated to: {13 :c}
    but is not: eq to {15 :c, 14 :b, 13 :a}
```

Diag:
```
;;; < as argument works fine:

(sorted-map-by < 13 :a 14 :b 15 :c)
;; clj => {13 :a, 14 :b, 15 :c}
;; phel => {13 :a, 14 :b, 15 :c}


;;; > as argument doesn't:

(sorted-map-by > 13 :a 14 :b 15 :c)
;; clj => {15 :c, 14 :b, 13 :a}
;; phel => {13 :c}
```

Surfaced in:
https://github.com/jasalt/clojure-test-suite/actions/runs/25041645411/job/73346126704#step:7:292

SAY-5:
Closes #1705.

## Reproduction (from the issue)
```clojure
(sorted-map-by > 13 :a 14 :b 15 :c)
;; clj  => {15 :c, 14 :b, 13 :a}
;; phel => {13 :c}                  ← wrong, every key collapsed into one slot
```
`<` worked, `>` did not.

## Root cause
The Phel/Clojure idiom is to pass a **predicate** (`>`, `<`, etc.) to `sorted-map-by`. A predicate returns `bool`, but the internal `SortedArrayHelper::binarySearch` and the persistent-sorted-map insert path needed a tri-state -1/0/1 comparator. PHP coerces `false` → `0` in the binary search loop:
```php
$cmp = $comparator($array[$mid * 2], $key);   // false for >, ≤
if ($cmp < 0) ...
elseif ($cmp > 0) ...
else            return $mid * 2;             // ← treated as \"key found\"
```
So with `>`:
- `(> 13 14)` → `false` → `0` → binary-search reports key 14 already at index 0 → put **replaces** the value of 13 with `:b`.
- `(> 13 15)` → `false` → `0` → put **replaces** the value of 13 with `:c`.

End state: `{13 :c}`. The third put obliterates the first two.

## Fix
Two complementary changes:

### 1. `SortedArrayHelper::resolveComparator` adapts bool predicates
```php
return static function (mixed $a, mixed $b) use ($closure): int {
    $result = $closure($a, $b);
    if (is_bool($result)) {
        if ($result) return -1;
        return $closure($b, $a) === true ? 1 : 0;
    }
    return ((int) $result) <=> 0;
};
```
This mirrors Clojure's `clojure.lang.Util.compare`-style predicate adaptation:
- `(> 14 13)` → `true`  → `-1` (14 sorts before 13)
- `(> 13 14)` → `false` → probe `(> 14 13)` → `true` → `+1`
- `(> 13 13)` → `false` → probe `(> 13 13)` → `false` → `0`

Int comparators are normalized to `{-1, 0, 1}` so the binary search no longer relies on truthiness.

### 2. `PersistentSortedMap::empty` no longer double-wraps
The constructor already calls `resolveComparator` to build `$effectiveComparator`. `empty()` was *also* calling it before passing into the constructor, which (a) double-wrapped after the change above and (b) made `getComparator()` hand back the wrapper instead of the user's original closure. Now `empty()` only does the bare `callable → Closure` conversion.

## Test plan
**Unit tests for the comparator adapter** (`SortedArrayHelperTest`):
- [x] `test_resolve_bool_predicate_is_adapted_to_int` — descending `>` predicate produces -1 / +1 / 0.
- [x] `test_resolve_bool_predicate_ascending` — ascending `<` predicate.
- [x] `test_resolve_int_comparator_is_wrapped_and_normalized` — replaces the old closure-identity assertion with a contract test: the wrapper preserves *ordering* of an int comparator and normalizes the magnitude to `{-1, 0, 1}`.

**Integration tests for the persistent map** (`PersistentSortedMapTest`):
- [x] `test_sorted_map_with_greater_than_predicate_keeps_all_keys` — exact repro from #1705: 13/14/15 with `>` keeps all three keys and iterates as [15, 14, 13].
- [x] `test_sorted_map_with_less_than_predicate_keeps_all_keys` — `<` predicate iterates as [13, 14, 15].

Results:
- [x] `vendor/bin/phpunit --filter SortedMap` → **54 passed; 0 failed**.
- [x] PHP repro: `PersistentSortedMap::empty(..., fn(\$a,\$b) => \$a > \$b)->put(13,'a')->put(14,'b')->put(15,'c')` now iterates as `15 => c / 14 => b / 13 => a`, count = 3.

## Notes
- No changes to the Phel-level public API; the fix is entirely in the PHP runtime layer.
- `getComparator()` again returns the user's exact closure (existing `test_get_comparator` keeps passing).


Chemaclass (approved):

Chemaclass:
@SAY-5 please fix the pr ci jobs

Chemaclass:
Closing as superseded by #1741 (merged in f66e7abd), which fixed #1705 via a slightly different approach: a dedicated `SortedArrayHelper::adaptForBinarySearch` invoked at the search call sites, leaving `resolveComparator` as a pure `callable → Closure` conversion. Same root cause, same outcome, equivalent test coverage.                
                                                                                                                     
Thanks for the diagnosis and the clear write-up. Really helpful in confirming the fix on the other branch. Closing this one to avoid duplicate work.

PR #1721, opened 2026-04-28
feat(core): namespace introspection in phel\core

### LINKED ISSUE #1694: repl: Consider moving `phel\repl` definitions common with Clojure to `phel\core`
Many definitions currently in non-core namespace `phel\repl` reside in `clojure.core` instead making them directly available during REPL session in Clojure which is very convenient.
It would be handy if this was true for Phel also.

Definitions that are Phel specific could be left in `phel\repl` while common ones with Clojure moved to `phel\core`.

## Direct equivalents
Should be straightforward to move.

- `require` ([clojure](https://github.com/clojure/clojure/blob/b4ea0f824b2eea039dfc06b796ed601e35cbeab6/src/clj/clojure/core.clj#L6066) / [phel](https://github.com/phel-lang/phel-lang/blob/0dde915b515a119c398476d21ad4811723ce0ed4/src/phel/repl.phel#L133-L139))
- `use`
- `create-ns`
- `remove-ns`
- `intern`
- `ns-interns`
- `ns-publics`
- `ns-aliases`
- `ns-refers`
- `macroexpand-1`
- `macroexpand`
- `load-file` (available globally in Clojure via clojure.lang.RT)

## Near equivalent

There's also some cases where definition names don't match while functionality is (nearly) the same.
These could be considered case-by-case, e.g. if they would be renamed to be aligned better with Clojure in scenarios where they work the same and wouldn't bring surprises when switching language.

- Phel [`compile-str`](https://github.com/phel-lang/phel-lang/blob/0dde915b515a119c398476d21ad4811723ce0ed4/src/phel/repl.phel#L187) (v.s. Clojure [`compile`](https://github.com/clojure/clojure/blob/clojure-1.12.4/src/clj/clojure/core.clj#L6066) ?)
- Phel [`eval-str`](https://github.com/phel-lang/phel-lang/blob/0dde915b515a119c398476d21ad4811723ce0ed4/src/phel/repl.phel#L196) v.s. Clojure [`load-str`](https://github.com/clojure/clojure/blob/b4ea0f824b2eea039dfc06b796ed601e35cbeab6/src/clj/clojure/core.clj#L4115)

## Others

These listings are possibly missing some definitions that could be worth looking at.

Chemaclass:
## 🤔 Background

`phel\repl` exposes namespace inspection helpers (`loaded-namespaces`, `find-ns`, `ns-publics`, …) that are only convenient inside an interactive session. Issue #1694 asks for them to be reachable from any namespace, so tooling and library code can introspect the registry without depending on `phel\repl`.

## 💡 Goal

Expose a stable subset of namespace functions from `phel\core` so they are auto-available everywhere, while keeping existing `phel\repl` definitions untouched for backward compatibility.

## 🔖 Changes

- New module `src/phel/core/ns.phel` wired via `(load \"core/ns\")` in `core.phel`
- Adds: `loaded-namespaces`, `find-ns`, `create-ns`, `remove-ns`, `intern`, `ns-interns`, `ns-publics`, `ns-aliases`, `ns-refers`, `load-file`
- Test coverage in `tests/phel/test/core/ns.phel`
- Changelog entry under `## Unreleased`

Closes #1694

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ This release focuses on Clojure-compatible core behavior, PHP interop consistenc
 - Add `into-array` for `.cljc` interop (#1550)
 - Compare numeric values consistently in `==` (#1561)
 - Support `[size init-val-or-seq]` in primitive array helpers (#1562)
+- Add namespace introspection: `loaded-namespaces`, `find-ns`, `create-ns`, `remove-ns`, `intern`, `ns-interns`, `ns-publics`, `ns-aliases`, `ns-refers`, `load-file` (#1694)
 
 ### Changed
 
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -240,3 +240,6 @@
 (load "core/loops")
 (load "core/futures")
 (load "core/async")
+
+;; Namespace introspection and manipulation
+(load "core/ns")
--- a/src/phel/core/ns.phel
+++ b/src/phel/core/ns.phel
@@ -1,0 +1,132 @@
+(in-ns phel.core)
+
+(use Phel)
+(use Phel\Compiler\Application\Munge)
+(use Phel\Compiler\CompilerFacade)
+(use Phel\Compiler\Infrastructure\CompileOptions)
+(use Phel\Compiler\Infrastructure\GlobalEnvironmentSingleton)
+(use Phel\Lang\Symbol)
+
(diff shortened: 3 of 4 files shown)


jasalt:
@Chemaclass nice.

How about ’require’?
It’s fine via REPL startup script, but I have had occasional namespace switching glitches leading to ’require’ becoming missing and me ending up restarting REPL. Investigating to get it more solid…

Issue #1773, opened 2026-04-29
`intern` and missing `var-get`


jasalt:
*phel v0.34.1-beta#82e0da5*

When `intern` was added to `phel\core` namespace via [`ns.phel`](https://github.com/phel-lang/phel-lang/blob/414c2684c19928522846b710f7d57aba8e94b1b2/src/phel/core/ns.phel#L51-L62), this brought up a compilation error in Clojure Test Suite `intern.cljc` which was skipped previously as the definition was missing.

The error comes from usage of `var-get` function in [`intern.cljc`](https://github.com/jasalt/clojure-test-suite/blob/6c65f1359ba317251499b95b18b30f58054661e2/test/clojure/core_test/intern.cljc#L9) that is not implemented in Phel.

Due to vars not being implemented, at least for now (per discussion https://github.com/phel-lang/phel-lang/issues/1717#issuecomment-4337925740), what would be the resolution for the missing `var-get`?

Should the test be skipped with Phel specific conditional (A), or (B) could there be some clever "hack" for having `var-get` that could simply resolve the symbol values.

Also, the `intern` (and possible `var-get`) docstring could shortly document the difference in semantics between Phel and Clojure.

jasalt:
## Inspecting behavior of `def` and `intern`:

Looking for possibility of solution (B).

```clojure
(def a 123)
;; clojure => #'user/a
;; phel => #'user/a

(type (def a 123))
;; Clojure => clojure.lang.Var
;; Phel: ParseError: syntax error, unexpected token ";", expecting ")"

#'user/a
;; clojure => #'user/a
;; phel => 123

(type #'user/a)
;; clojure => clojure.lang.Var
;; phel => :int

(var-get #'user/a)
;; clojure => 123
;; phel: Cannot resolve symbol 'var-get'.

(intern 'user 'b 456)
;; clojure => #'user/b
;; phel => b

(type (intern 'user 'b 456))
;; clojure => clojure.lang.Var
;; phel => :symbol

b
;; clojure => 456
;; phel => 456
```

This hints that:

- `def` inside `type` errors in Phel (possible bug)
- `(def a 123)` and `(intern 'user 'b 456)` output different type of value literals, it would be more consistent if `intern` outputted `#'user/b`.

jasalt:
[...]

PR #1774, opened 2026-04-29
feat(repl): add var-get, preserve existing values in intern

jasalt:
## 🤔 Background

https://github.com/phel-lang/phel-lang/issues/1773

## 💡 Goal

Add `var-get` that passes clojure-test-suite.

## 🔖 Changes

Modified intern to find existing definitions without overwriting when called without a value argument, matching Clojure semantics. Added var-get helper function to dereference symbols returned by intern, passing clojure-test-suite intern.cljc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Core behavior, PHP interop consistency, and stricter compiler diagnostics.
 - Consistent numeric comparison in `==` (#1561)
 - `[size init-val-or-seq]` in primitive array helpers (#1562)
 - Namespace introspection: `loaded-namespaces`, `find-ns`, `create-ns`, `remove-ns`, `intern`, `ns-interns`, `ns-publics`, `ns-aliases`, `ns-refers`, `load-file` (#1694)
+- `var-get` resolves a symbol to its registry value (#1774)
 - `(rand n)` returns a number in `[0, n)` (#1696)
 
 #### REPL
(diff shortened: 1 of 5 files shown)


jasalt:
Huh, 

Locally passing test:
```
(def test-var-get-def 123)
(is (= 123 (var-get (symbol *ns* "test-var-get-def"))) "def in current namespace")
```

failing in CI with:
```
FAIL in (test-var-get 'general var-get behavior > successful lookups - def in current namespace') (ns.phel:32)
          Form: (var-get (symbol *ns* "test-var-get-def"))
  evaluated to: nil
    but is not: = to 123
```

jasalt:
Looking for alternate way to test `var-get` with `def` and noted issue with:

```
(var-get (def x 123))
;; clojure => 123
;; phel: ParseError: syntax error, unexpected token ";", expecting ")"
```

Could be related to [the issue comment](https://github.com/phel-lang/phel-lang/issues/1773#issuecomment-4340870090) where `(type (def a 123))` had same result.

jasalt:
@raspi-jesus-bot correct, 1. was a test isolation issue.

```
(def test-var-get-def 123)
(is (= 123 (var-get (symbol *ns* "test-var-get-def"))) "def in current namespace")
```
This succeeded in my earlier scenario as I ran tests only that specific ns with `./bin/phel test tests/phel/test/core/ns.phel`.

It fails with full test suite run, which I didn't come to try locally previously: `./bin/phel test`:

```
FAIL in (test-var-get 'general var-get behavior > successful lookups - def in current namespace') (ns.phel:32)
          Form: (var-get (symbol *ns* "test-var-get-def"))
  evaluated to: nil
    but is not: = to 123
```

jasalt:
Considering to leave `def` tests out for now from this PR, so this is ready now to review for merge.

Parser error issue with `(type (def x 123))` and `(var-get (def x 123))` should be resolved separately.

PR #1896, opened 2026-05-07
fix(test): --filter silently drops non-matching tests instead of skipping (#1888)

### LINKED ISSUE #1888: test: `--filter` should exclude non-matching tests, not emit skip events
## 🤔 Background

The `vendor/bin/phel test --filter <name>` option currently runs through every discovered test and emits a `:skipped` event for any test whose name does not contain the filter substring. Reporters render each skip as `S`, so a single matching test produces a long `SSSS...` prefix in the output and inflates the `Skipped:` counter.

Example from a downstream project:

```
➜  phelgeon vendor/bin/phel test --filter fatal-incoming-damage-flips-to-gameover tests/core_test.phel
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS.

Passed: 1
Failed: 0
Error: 0
Skipped: 45
Total: 1
Time: 00:03.620, Memory: 54.02 MB
```

The intent of `--filter` is selection, not skipping. Non-matching tests should be silently excluded (the way PHPUnit's `--filter` behaves), so the output reflects only the tests that were actually selected to run.

## 💡 Goal

Tests excluded by `--filter` should not appear in reporter output and should not be counted in `Skipped:`. Only tests that match the filter are discovered, run, and reported.

`:skipped` events remain valid for selector-based exclusion (`--include` / `--exclude` / tag selectors), where surfacing the skip is informative.

## 🔖 Changes

- In `src/phel/test.phel`, treat the legacy `--filter` gate as a discovery filter rather than a skip reason: drop matching tests from the discovered set instead of routing them through `:skip` with reason `:filtered`.
- Confirm reporters (`default`, `testdox`, JUnit) no longer emit `S` or increment `Skipped:` for filter misses.
- Add a Phel test covering: filter that matches one test out of many produces no `S` characters and `Skipped: 0`.

## Expected output after fix

```
➜  vendor/bin/phel test --filter fatal-incoming-damage-flips-to-gameover tests/core_test.phel
.

Passed: 1
Failed: 0
Error: 0
Skipped: 0
Total: 1
```

r-frolov:
## 🤔 Background

`vendor/bin/phel test --filter <name>` currently emits a `:skipped` event for every test whose name does not contain the filter substring. Reporters render each skip as `S`, so running with a narrow filter produces a long `SSSS...` prefix and inflates the `Skipped:` counter — even when only one test matches.

Closes #1888.

## 💡 Goal

Make `--filter` (and the underlying `:filters` option) act as a **discovery filter**: non-matching tests are silently dropped before they reach the reporter layer, matching PHPUnit's `--filter` behaviour. Tag/namespace selectors (`--include`, `--exclude`, `--ns`) continue to emit `:skipped` events, where surfacing the skip is informative.

## 🔖 Changes

- **`src/phel/test.phel` — two-phase filter design**
  - `run-test` now pre-filters `all-tests` by name (`:filter` + `:filters`) before passing to `partition-tests`. Non-matching tests are silently dropped; no events emitted.
  - `selector-options` no longer forwards `:filters` — tag/namespace selectors only. This removes the hidden coupling and the need for the `tag-options` workaround.
  - `skip-reason` simplified: the dead `:filtered` branch removed; `options` parameter dropped (was unused after the above change).
  - `partition-tests` updated to match the new `skip-reason` arity.

- **`tests/phel/selectors-runtime.phel` — updated and new tests**
  - `test-filter-by-name-substring` replaced by `test-filter-by-name-excludes-silently` (verifies 0 skip events and `Skipped: 0` in summary).
  - `test-legacy-filter-key-excludes-silently` added for the legacy `:filter` (singular) path.
  - `test-filter-and-tag-selector-combined` added to verify the two-phase design: name filter drops tests silently while a tag selector on the remaining tests still emits `:skipped` events correctly.

- **`CHANGELOG.md`** — entry added under `## Unreleased → ### Fixed → #### Testing`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,9 @@ All notable changes to this project will be documented in this file.
 - `bigdec` accepts `Rational`; non-terminating expansions raise `ArithmeticError` (#1873)
 - `+`, `-`, `*`, `/`, `abs`, `quot`, `rem`, `mod`, `rationalize` accept `BigDecimal` operands; `/` raises `ArithmeticError` on non-terminating expansions (#1875)
 
+#### Testing
+- `--filter` (and `:filters`) now acts as a discovery filter: non-matching tests are silently dropped instead of emitting `S` (skipped) markers and inflating the `Skipped:` counter (#1888)
+
 #### Lang
 - `=` between `int` and `BigInteger` is symmetric (#1830)
 - `+`, `-`, `*`, `**` on PHP ints auto-promote to `BigInteger` on overflow (#1830)
--- a/src/phel/test.phel
+++ b/src/phel/test.phel
@@ -1017,27 +1017,22 @@
             :meta meta}))))
 
 (defn- selector-options
-  "Extracts the selector subset from the full options map."
+  "Extracts the tag/namespace selector subset from the full options map.
+  Name-based filters (`:filter` / `:filters`) are handled separately as
+  discovery filters and are intentionally excluded here."
   [options]
   {:include     (:include options)
    :exclude     (:exclude options)
-   :ns-patterns (:ns-patterns options)
-   :filters     (:filters options)})
+   :ns-patterns (:ns-patterns options)})
 
 (defn- skip-reason
-  "Returns the reason the test should be skipped — `:filtered` for the
-  legacy `--filter` gate, `:selector` for any other selector miss, or
-  `nil` when the test should run."
-  [options sel-opts ns-name m]
-  (cond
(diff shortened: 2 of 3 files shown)


Chemaclass (approved):
LGTM. Two-phase filter design clean, tests cover both legacy  and new  plus combined name+tag case.

Chemaclass (approved):
LGTM. Two-phase filter design clean; tests cover legacy :filter, new :filters, and combined name+tag case.

PR #1900, opened 2026-05-07
perf(test): mtime+size fast-path, single-scan namespace collection

Chemaclass:
## 🤔 Background

`bin/phel test` walked three overlapping namespace-directory scans (test ns, bundled `phel.*`, dep closure) and re-read+md5'd every cached source on each cache hit, even when the file had not changed in hours.

## 💡 Goal

Cut redundant filesystem and hashing work in the `phel test` boot phase without weakening the existing content-hash correctness contract.

## 🔖 Changes

- `CompiledCodeCache` records `source_mtime` + `source_size`; `getByFingerprint` returns the compiled path when both still match. A 2s grace window and the size guard prevent same-second size-preserving writes from serving stale entries.
- `FileEvaluator::evalFile` tries the fingerprint fast-path first and falls back to the content-hash path on any mismatch.
- `NamespaceCollector::getDependenciesFromPaths` collapses three overlapping directory scans into one. Bundled `phel.*` detection is scoped to src+vendor roots so a fixture namespace like `phel.test.clojure-test-alias-regression` is not seeded as bundled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ All notable changes to this project will be documented in this file.
 - Cold REPL boot prunes `vendor/`, `.git/`, `node_modules/` at namespace-scan descent and memoises directory scans per process (#1885)
 - `bin/phel test --ns 'pat.**'` preloads only matching namespaces and their dependency closure
 - `bin/phel test` skips re-registering dependency-graph entries on cache hits and re-restoring a namespace's environment after the first file in it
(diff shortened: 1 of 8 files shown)


Chemaclass:
Closing: bench showed boot-only median 399 ms (main) vs 421 ms (branch) and full warm 8524 ms vs 8721 ms — fingerprint path adds two stat calls per evalFile entry that wash out the file_get_contents+md5 savings, and the test-command wall-time is dominated by ~7 s of test execution that this change does not touch. Replacing with a smaller refactor that keeps only the single-scan NamespaceCollector consolidation.

PR #1955, opened 2026-05-11
feat(filesystem): consolidate runtime artifacts under .phel/

### LINKED ISSUE #1954: DX: auto-write .phel/.gitignore on first directory creation
## Background

Phel writes per-project runtime state under several paths in a downstream user's project root:

- `.phel/last-failed.txt` (test runner — `phel test --last-failed`)
- `.phel/lint-cache/index.json` (lint cache)
- `.phel-repl-history` (REPL)
- `out/` (build output)

Every new project has to remember to add these to its own `.gitignore`, and most forget at least one.

### PR-A — auto-write `.phel/.gitignore` (this issue)

The first time Phel writes anything under `<project>/.phel/`, also write `<project>/.phel/.gitignore` containing:

```
# Created automatically by Phel.
*
```

Same pattern as pytest, mypy, ruff. Brand-new projects never see `.phel/` in `git status`, no user config needed.

Centralize the directory creation behind a single helper (`Phel\Filesystem\Application\PhelProjectDirectory::ensure(string $projectRoot)`) and route every existing `.phel/`-writing call site through it:

- `TestCommand` (last-failed)
- `LintCommand` / `LintCache` (lint-cache)

Behavior:
- `mkdir(.phel/, 0755, true)` if missing.
- `file_put_contents(.phel/.gitignore, "...", LOCK_EX)` only if missing — never overwrite user edits.

### PR-B — consolidate cache + repl-history under `.phel/` (deprecation cycle, separate issue/PR)

Out of scope for this issue. Will follow once PR-A merges.

## Acceptance for PR-A

- `composer test` green.
- Manual: in a scratch project, after a failing `phel test`, `.phel/.gitignore` exists with the auto-comment + `*`; `git status` shows nothing.
- Editing `.phel/.gitignore` and re-running keeps the edit.

Chemaclass:
## 🤔 Background

Phel writes per-project runtime state in scattered locations: `.phel/last-failed.txt`, `.phel/lint-cache/`, `<systemTemp>/phel/cache/`, `<projectRoot>/.phel-repl-history`. Every downstream project has to remember matching `.gitignore` entries, and the cache living in system temp is wiped on reboot.

Closes #1954.

## 💡 Goal

Single project-local directory `.phel/` for everything Phel writes inside a user's repo. Self-seeded `.gitignore` (`*`) makes it invisible to Git, so brand-new projects need zero config.

### Auto-`.gitignore`

- New `Phel\Filesystem\Application\PhelProjectDirectory::ensure(string $projectRoot)` static helper. Creates `.phel/` if missing; atomically writes `.phel/.gitignore` with `*` only if missing (`LOCK_EX`). Throws `FileException` on mkdir failure.
- `TestCommand` calls the helper before forwarding to the runner; gated on `!--list-only`.
- `LintCommand::maybeCache` calls the helper before instantiating `LintCache`.

### Cache consolidation

- `PhelConfig::cacheDir` default changes from `sys_get_temp_dir() . '/phel/cache'` (absolute) to `.phel/cache` (relative). `BuildConfig` resolves it against the app root, so the effective default is `<projectRoot>/.phel/cache`.
- `BuildConfig::getCacheDir()` honors `PHEL_CACHE_DIR` env var as the final override (wins over both default and explicit `setCacheDir()`).
- `PhelConfig::setCacheDir()` continues to override the default when no env var is set.

### REPL history migration

- `Phel\Run\Application\ReplHistoryPathResolver` resolves history at `<projectRoot>/.phel/repl-history`. If the legacy `<projectRoot>/.phel-repl-history` exists and the new path doesn't, it renames the file and emits a one-line stderr notice. Set `PHEL_QUIET_MIGRATION=1` to silence.
- `RunConfig::getPhelReplHistory()` removed; `RunFactory` builds the resolver from `getAppRootDir()`.

### Tests

- `PhelProjectDirectoryTest` (5 cases): create, preserve existing content, idempotent, trailing-separator, mkdir failure.
- `ReplHistoryPathResolverTest` (4 cases): fresh project, legacy migration emits notice, `PHEL_QUIET_MIGRATION=1` silences, both-present is no-op.
- `BuildConfigTest` (2 new cases): env override wins; empty env ignored.
- `PhelConfigTest` adjusted: split former `single_base_path` test to reflect new defaults.

### Docs

- New `docs/project-layout.md` documents the `.phel/` directory contract, override env vars, migration.
- `src/php/Filesystem/CLAUDE.md` lists `PhelProjectDirectory` in the structure tree + key constraints.
- `src/php/Run/CLAUDE.md` lists `ReplHistoryPathResolver` + migration constraint.

## Out of scope

- `out/` build directory unchanged (different lifecycle: deliverable artifact, sometimes committed/Docker-shipped).
- `phel init` `.gitignore` template untouched (it never listed `data/` or `.phel-repl-history`).
- cli-skeleton update is a follow-up PR.

## Manual verification

```
$ phel test
$ ls .phel/
.gitignore  cache/  last-failed.txt
$ cat .phel/.gitignore
# Created automatically by Phel.
*
$ git status            # clean

$ phel repl             # legacy .phel-repl-history -> .phel/repl-history
phel: migrated REPL history /path/.phel-repl-history -> /path/.phel/repl-history (set PHEL_QUIET_MIGRATION=1 to silence)

$ PHEL_CACHE_DIR=/tmp/ci-cache phel build   # cache lands at /tmp/ci-cache
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,9 @@ All notable changes to this project will be documented in this file.
 #### Compiler
 - Recursive self-calls emit `$this(...)`; ~3.66x faster on `fib(22)` (#1914)
 
+#### DX
+- Auto-write `.phel/.gitignore` (`*`) on first directory creation; user edits preserved (#1954)
+
 ### Fixed
 
 #### Compiler
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,7 +37,7 @@ All notable changes to this project will be documented in this file.
 - Recursive self-calls emit `$this(...)`; ~3.66x faster on `fib(22)` (#1914)
 
 #### DX
-- Auto-write `.phel/.gitignore` (`*`) on first directory creation; user edits preserved (#1954)
+- Runtime state consolidated under `.phel/` (cache, REPL history, last-failed, error log) with self-seeded `.gitignore`; legacy `.phel-repl-history` auto-migrated; relocate the whole directory via `setPhelDir()` or `PHEL_DIR` env, narrow cache override via `PHEL_CACHE_DIR` (#1954)
 
 ### Fixed
 
--- a/docs/project-layout.md
+++ b/docs/project-layout.md
@@ -1,0 +1,42 @@
+# Project Layout
+
+Phel writes per-project runtime state under a single `.phel/` directory at the
+project root. The directory is created lazily on first use and is auto-ignored
+by Git via a self-seeded `.phel/.gitignore` (`*`).
+
+| Path                            | Owner            | Purpose                              |
+| ------------------------------- | ---------------- | ------------------------------------ |
+| `.phel/cache/`                  | Build / compiler | Namespace + compiled-code cache      |
+| `.phel/lint-cache/index.json`   | Lint             | Per-file diagnostic cache            |
+| `.phel/last-failed.txt`         | Test runner      | Backing file for `phel test --last-failed` |
+| `.phel/repl-history`            | REPL             | Readline history (was `.phel-repl-history`) |
+| `.phel/error.log`               | Runtime          | Error log (was `/tmp/phel-error.log`) |
+| `out/`                          | Build            | Compiled PHP entry points (build artifacts; lifecycle differs) |
+
+## Overrides
+
+- **Relocate everything**: `$config->setPhelDir('/var/cache/phel')` in
+  `phel-config.php` moves cache, REPL history, last-failed file and error log
+  out of the project root. Useful for WordPress plugins, shared hosting, or
+  any setup where `.phel/` would otherwise sit under a web-accessible path.
+- **`PHEL_DIR` env var** — same effect at runtime; wins over `setPhelDir()`.
+- **`PhelConfig::setCacheDir($path)`** — narrower override for just the build
+  cache.
+- **`PHEL_CACHE_DIR` env var** — final cache override; wins over `PHEL_DIR`
+  and explicit `setCacheDir()`. Useful for CI / Nix builds.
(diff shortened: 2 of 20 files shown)


jasalt:
@Chemaclass nice to centralize these to one location.

Would it make sense to add option to move the whole `.phel` directory to alternate location?

I was thinking of situation where Phel used in WordPress plugin might lead to the `.phel` directory becoming accidentally visible to public internet if web server if not configured properly to keep the directory `www/wp-content/plugins/phel-plugin/.phel/` hidden. If there's no access to the web server config it might help working that around if the `.phel` directory could configured to be placed elsewhere easily. Not sure if that's very good workaround (seen hacks like that done in some WP plugins), but could explore the area a bit more to know that the setup would be secure.

Chemaclass:
Addressed the WP plugin concern: `.phel/` is now fully relocatable.

**`phel-config.php`:**
```php
return PhelConfig::forProject('my-plugin\\main')
    ->setPhelDir('/var/cache/phel'); // anywhere outside the document root
```

**Or via env (e.g. `wp-config.php`, CI, Docker):**
```bash
PHEL_DIR=/var/cache/phel phel build
```

Precedence: `PHEL_CACHE_DIR` > `PHEL_DIR` env > `setPhelDir()` > default `.phel`. Cache, REPL history, last-failed file and error log all route through the new `PhelProjectDirectory::resolve()` helper, so a single switch moves everything.

Also: `PhelProjectDirectory::ensure()` is now best-effort — on a read-only filesystem (Lambda, `:ro` Docker mount, sandbox runners) Phel silently runs without cache / last-failed / REPL history persistence instead of crashing.

PR #1963, opened 2026-05-11
feat(nrepl): resolve session-defined symbols in lookup/info/eldoc

Chemaclass:
## 🤔 Background

The built-in nREPL server's `lookup` / `info` / `eldoc` ops only consulted the static `ApiFacade::getPhelFunctions()` catalog. That catalog covers core and documented library functions, so anything defined inside a live session (the common nREPL flow from Calva, CIDER, Conjure, Cursive) came back as `no-info`. Concretely:

```
(defn greet [n] (str "hello " n))  ; eval over nREPL
;; sent later in the same session:
{:op "lookup" :sym "greet"}        ; -> status [done no-info]
```

The runtime registry already has the metadata (used by `phel\repl/get-symbol-info`); the nREPL handler just wasn't asking for it.

## 💡 Goal

Make `lookup` / `info` / `eldoc` return real metadata for session-defined symbols so editor tooling can resolve docs, arglists, and jump-to-source for code the user just typed.

### Api
- New `SymbolMetadataFinder` (and `SymbolMetadataFinderInterface`) walks the runtime registry through `Munge`, resolving symbols against the caller's namespace, then `phel.core`, then any loaded namespace. Wraps params-only arglists (`[n]`) into `(name args)` form so signatures stay consistent with docstring-derived ones.
- `ApiFacadeInterface::findSymbolMetadata(symbol, currentNs)` exposes the new finder.

### Nrepl
- `LookupOp` accepts a `SessionRegistry` and:
  - honors the request `ns` param when present,
  - falls back to the session's current namespace,
  - and only then to `"user"`.
- When the static catalog misses, defers to `findSymbolMetadata` so registry-backed entries (user `defn`, library defs not in the documented set) resolve too.

### Tests
- `SymbolMetadataFinderTest`: bare/qualified resolution, dot vs backslash, multi-arity arglists, docstring fallback, all-namespace scan.
- `LookupOpTest`: session-defined fallback, session-ns resolution, explicit-`ns`-overrides-session.
- `NreplServerTest`: end-to-end clone → defn → lookup returns populated `info` dict over the live socket.

### Changelog
- Added an "Nrepl" entry under the Unreleased section.

## ✅ Verification

```
$ ./bin/phel nrepl --port=17889
# clone → eval (defn greet [n] (str "hi " n)) → lookup greet
# returns: info{name=greet, ns=user, arglists-str=(greet n)}, eldoc[(greet n)], status[done]
```

All existing unit + integration tests pass. The two failing `PharExecutionTest` cases on this branch reproduce on `main` (stale bundled PHAR), so they are unrelated.


Chemaclass:
Thanks for the catch on the arglists key access. Verified: `DefSymbol::analyze()` writes the arglist under the plain string key (`$metaMap->put('arglists', ...)` — see `src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/DefSymbol.php:91,97`), so the current accessor matches the runtime contract. The end-to-end integration test (`NreplServerTest::test_it_returns_lookup_info_for_session_defined_symbols`) covers it: it evals a real `(defn greet [n] ...)`, performs a `lookup`, and asserts the normalised `(greet n)` arglist that only comes out of this branch.

That said, the asymmetry was real and worth hardening. Pushed `c299c9f5` to probe both `meta['arglists']` and `meta[Keyword::create('arglists')]`, plus a unit test that registers metadata under the keyword key to lock the contract.

PR #1991, opened 2026-05-14
feat(test): add --parallel flag for namespace-level parallel test runs

Chemaclass:
## 🤔 Background

`./bin/phel test` runs Phel test namespaces sequentially. Profiling on the current `main` suite shows wall-time is dominated by test bodies (89% warm cache, ~30s for 89 namespaces). Namespace boundaries are natural fork points: no shared mutable runtime state crosses them, so we can fan out work cheaply.

This PR introduces an opt-in `--parallel` flag that runs namespaces across a subprocess worker pool, mirroring Clojure's **eftest** with `:multithread? :namespaces`: opt-in, default workers ≈ CPU count, deterministic output ordering, no impact on existing serial runs.

## 💡 Goal

Cut test wall-clock for suites with non-trivial per-namespace work, while keeping the on-screen view deterministic and the existing serial path untouched.

### CLI

- `--parallel=<N|auto>` on `phel test`. `auto` triggers CPU detection (env var `PHEL_TEST_WORKERS`, fallback chain `nproc` → `sysctl -n hw.ncpu` → `/proc/cpuinfo` → `4`, capped at 8). Omit the flag for current serial behaviour. `--parallel=1` collapses back to serial.
- Auto-disabled when `--reporter=tap` (TAP needs sequential numbering), `--list` (no execution), or a profiler hook is installed (`Registry::$profilerHook`).

### Runtime

- New hidden subcommand `phel _test-worker`: long-lived subprocess; reads one work frame per namespace from stdin, captures stdout via `ob_start`, eval's `(do (phel\test/run-tests opts 'ns) (phel\json/encode {...stats...}))`, writes a JSON result frame.
- `Phel\Run\Application\Test\ParallelTestOrchestrator`: spawns the pool via `proc_open`, polls workers via `stream_select`, buffers each ns's output, flushes in input order so terminal output stays deterministic across worker scheduling jitter.
- `WorkerFrame`: length-prefixed JSON framing (`<8 hex digits>\n<body>`).
- `TestWorkerHandle`: encapsulates the live subprocess lifecycle, mirroring the `proc_open` idiom already used by `Phel\Watch\Application\Watcher\AbstractShellWatcher`.
- `CpuCountDetector`: cross-platform autodetect (no pcntl dependency).

### Phel

- New public fn `phel\test/get-failed-tests` so the parent can aggregate the `last-failed.txt` list across worker processes (workers run with `:last-failed-file nil` to avoid the race; parent writes the aggregated file).

### Out of scope (documented in plan)

- TAP reporter parallelism (falls back to serial; eftest doesn't either).
- Profiler + parallel (hook lives in parent only).
- `--fail-fast` cross-worker cancellation.
- Windows (`proc_open` works, but `stream_select` + signal semantics differ; CI is Linux/macOS only).

## ✅ Verification

Screenshot 2026-05-15 at 00 23 13


JesusValeraDev (approved):

Chemaclass:
Round 3 (0df0bbd): progress-bar UX, replacing the per-namespace dump.

### Before
Every namespace printed its own `--- [i/N] ns ---` header followed by a `Passed: 0 / Failed: 0 / Total: 0` mini-table. On a real run that meant dozens of empty blocks for transitive `phel.*` deps, drowning the useful signal.

### After
```
Running 33 namespace(s) across 2 parallel worker(s)...
 33/33 [============================] 100%    2 s — phel-test.are

Passed:  55
Failed:  0
Error:   0
Total:   55

Ran 33 namespace(s) across 2 worker(s) in 2.10s.
```

- Live `Symfony\ProgressBar` shows `current/total [bar] %% elapsed — current-namespace` updating in place.
- Per-namespace block only renders when the namespace had a failure or error (the progress bar pauses, the block prints under a `--- ns ---` header, then the bar resumes).
- Single aggregate `Passed / Failed / Error / [Skipped] / Total` block at the end, computed by the parent from per-worker counts.

### Wire-format change
Each result frame now carries a `counts` object (pass/failed/error/skipped/total) — fields already produced by `phel.test/get-stats`. New `Counts` value object on the PHP side; `OrderedResultBuffer` aggregates and exposes `totals()`.

### Tests
- New `CountsTest`: defaults, `fromArray` coercion (negatives clamp, numeric strings parse, garbage ignored), `add` accumulation, `hasFailures` predicate.
- `WorkerResultTest` / `OrderedResultBufferTest` extended for the counts payload, failure-only flush, and `finishProgress()`.
- `ParallelTestRunnerTest` adds an end-to-end assertion that the bar reaches `100%` and the aggregate `Passed/Failed/Total` block appears.

33 unit tests / 7 integration tests, all green.

PR #1992, opened 2026-05-16
feat(core): add lazy-seq? predicate

Chemaclass:
## 🤔 Background

While discussing [jank-lang/clojure-test-suite#883](https://github.com/jank-lang/clojure-test-suite/pull/883) (a portable `lazy-seq?` for that shared test suite), it became clear Phel lacked a public predicate that tells "this is *specifically* a lazy sequence" apart from "this is a list".

Today users have to fall back to interop:

```clojure
(php/instanceof x Phel\Lang\Collections\LazySeq\LazySeqInterface)
```

…which leaks the implementation namespace and is not the kind of DX we want to point Clojure newcomers to.

`seq?` already exists but matches both lists and lazy sequences, so it cannot answer "is this lazy?" on its own.

## 💡 Goal

Ship `lazy-seq?` as a first-class `phel.core` predicate so user code (and the shared Clojure test suite) can ask the question idiomatically and without PHP-class FQNs.

## 🔖 Changes

- **`src/phel/core/predicates.phel`** : new `lazy-seq?` defined right after `seq?`. Implementation is a single `php/instanceof` against the existing `LazySeqInterface` (already `use`d at the top of the file). Same interface check as `seq?`, `empty?`, `sequential?`, `seqable?`, and `realized?` use today, so it stays correct for any future impl behind the interface.
- **`src/phel/core/predicates.phel`** : `seq?` `:see-also` extended with `lazy-seq?` for doc-graph consistency.
- **`tests/phel/core/sequence-functions.phel`** : new `test-lazy-seq?` covering positive cases (`map`, `lazy-seq` macro, `iterate`) and negative cases (list, vector, map, nil, scalar).
- **`CHANGELOG.md`** : `## Unreleased` → `### Added` bullet.

No PHP source touched; no `Phel\Lang\Collections\LazySeq\*` semantics change.

### Why an interface check, not a concrete-class check

`Phel\Lang\Collections\LazySeq\LazySeq` is final today, but Phel core consistently dispatches lazy-seq logic on `LazySeqInterface`. Following the same convention keeps the new predicate future-proof if/when additional lazy-seq impls land behind that interface.

### Verification

```bash
composer test-core
./bin/phel test tests/phel/core/sequence-functions.phel
composer test-quality
```

Acceptance examples:

```clojure
(lazy-seq? (map inc [1 2 3])) ; => true
(lazy-seq? (iterate inc 0))   ; => true
(lazy-seq? '(1 2 3))          ; => false
(lazy-seq? [1 2 3])           ; => false
(lazy-seq? nil)               ; => false
```

### Note on `clojure-test-suite#883`

The PR there defines its own `lazy-seq?` inside `portability.cljc`, so the `:phel` branch in that file still has to use the `php/instanceof` form (calling `lazy-seq?` would self-recurse). Phel users outside that one file get the native predicate.


JesusValeraDev (approved):

dgr:
I'd just note on this that `lazy-seq?` is not an official Clojure core API. I'm surprised that Clojure doesn't have it, as it would be useful, but it's not there today.

PR #2042, opened 2026-05-20
feat(compiler): hoist pure collection literals + PHP 8.5 dev fixes

Chemaclass:
## 🤔 Background

Pure collection literals (`[1 2 3]`, `{:a 1}`, `#{:x :y}`) built inside a fn body were being rebuilt on every call, even though `PersistentVector` / `PersistentHashMap` / `PersistentHashSet` are immutable.

## 🔖 Changes

- Hoist pure collection literals to a per-fn `static` cache. First call builds the persistent collection; later calls reuse the same instance.
- Implemented via `PureLiteralDetector` + `BodyConstantScanner` + `ConstantScope`; `MethodEmitter` emits `static $__phel_const_N;` and the literal emitters render `($__phel_const_N ??= \Phel::vector([...]))`.
- IIFE wrappers (`let`, `do`, `foreach`, … in expression context) get matching `use(&$__phel_const_N)` so nested scopes still hit the same slot.
- Tests: unit (detector + scope + scanner), 4 integration fixtures under `ConstantHoisting/`, and a runtime test (`ConstantHoistingRuntimeTest`) asserting `assertSame` across calls.

### Side fixes

- `build/scripts/patch-psalm.php` + composer post-install/update hook: patches vendored Psalm 6.16.1 for the PHP 8.5 *"NAN coerced to string"* crash (Psalm 7 fixes it upstream but reports 1650 new errors — out of scope here).
- `ConstantScope`: use `SplObjectStorage::offsetExists()` instead of PHP 8.5-deprecated `contains()` (the notice was leaking into emitted code).
- `phpstan` bumped 2.1.50 → 2.1.55 (patch).

## Test plan

- [x] `composer test-all` green on PHP 8.4 and 8.5
- [x] 5611 Phel core tests + 3436 PHPUnit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,15 +18,15 @@ All notable changes to this project will be documented in this file.
 
 - BC: release tooling moved from `build/` to `tools/`; `build/` is now phar-only
 - CI: split `ci.yml` into `quality.yml` / `tests.yml` / `smoke.yml` with a shared `setup-phel` composite action; add concurrency cancellation and least-privilege perms (#2039)
-- Compiler: pure vector/map/set literals inside a fn body are now hoisted to a per-fn `static` cache, so each invocation reuses the same persistent collection instead of rebuilding it
-- Compiler: `if` / `?:` skip the `Truthy::isTruthy` wrap when the test is a known-bool expression (literal `true`/`false`, infix comparison, `is_int`/`is_a`/etc.), emitting a direct PHP conditional and saving one assign + two comparisons per branch
-- Compiler: `recur` emits direct param assignments when no expression reads a recur param that an earlier expression has already overwritten, skipping the per-arg `$__phel_N` temp shuffle in the common case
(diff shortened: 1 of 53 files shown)


JesusValeraDev (approved):

Chemaclass:
Addressed both review points:

1. **`array_all()` PHP version** — kept `array_all` as-is. It was added in PHP 8.4.0 (rector classifies the rule under `Php84\Rector`), and composer.json declares `>=8.4` with platform pinned at 8.4.6, so the call is safe on every supported runtime. Verified locally on 8.4 and 8.5.

2. **`BodyConstantScanner::children()` footgun** — added a docblock note on the method explaining the manual node-type registry and reminding future contributors to audit `Phel\Compiler\Domain\Analyzer\Ast` when adding new node types. Kept the `match` form rather than introducing `AbstractNode::children()`, since every node would need a list of its own child slots and the wider AST refactor is out of scope here. Commit: `docs(compiler): note manual node-type registry in BodyConstantScanner::children`.

PR #2126, opened 2026-05-26
fix: preserve REPL runtime namespace after startup

### LINKED ISSUE #2125: `require` in REPL switches namespace to `phel.reflect`
*Phel Repl (v0.40.0-beta#c6dff7c)*

Expectation is that ns would not switch after require but following currently happens:

```
./bin/phel
Welcome to the Phel Repl (v0.40.0-beta#c6dff7c)
user:1> (require [phel.test :as t])
phel.test
phel.reflect:2> *ns*
"phel.reflect"
```

Same tested with require using `clojure.test` that gets automatically aliased to this same Phel namespace, older Phel require syntax without the vector `(require phel.test :as t)` and also the form without `:as` part.

SauronBot:
## 🤔 Background

`phel repl` restored the compiler namespace to `user` after loading the startup namespace and bundled dependencies, but the runtime `phel.core/*ns*` var still reflected the last dependency evaluated during startup. In the reported case, `(require [phel.test :as t])` made that visible as `phel.reflect`.

## 💡 Goal

Closes #2125. Keep the REPL runtime namespace aligned with the prompt/compiler namespace after startup and after requiring `phel.test`.

## 🔖 Changes

- Restore `phel.core/*ns*` alongside `GlobalEnvironment::setNs()` in `NamespaceLoader`.
- Add a REPL integration regression test for `*ns*`, `(require [phel.test :as t])`, then `*ns*` again.
- Document the fix in `CHANGELOG.md`.

Validation:

- `printf '%s\n' '*ns*' '(require [phel.test :as t])' '*ns*' 'exit' | ./bin/phel`
- `vendor/bin/phpunit tests/php/Integration/Run/Command/Repl/ReplPromptNamespaceTest.php`
- `vendor/bin/phpunit tests/php/Integration/Repl/RequireNamespaceTest.php`
- `composer csrun`
- `git diff --check`

Note: `composer test-compiler` still hits the suite's configured 512 MB PHPUnit memory cap on this machine before completion; the same happens when invoking PHPUnit with a higher CLI memory flag because `phpunit.xml.dist` sets `memory_limit=512M`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ Control-flow lowering to PHP `match` (#2091):
 - `phel.cli`: Symfony Console 8.0 compat. `Command::setCode` closure now carries explicit `InputInterface` / `OutputInterface` types; clears the Symfony 7.3 deprecation warning for the same reason (#2094)
 - `phel compile`: dry-run no longer executes side-effecting forms; new `CompileOptions` `emitOnly` flag skips the evaluator step (#2095)
 - `defonce`: same-file redefinition is a silent no-op (was `DuplicateDefinitionException`) (#2096)
+- REPL startup now restores runtime `*ns*` to `user`, so `(require [phel.test :as t])` no longer leaves `*ns*` in the last loaded dependency namespace (#2125)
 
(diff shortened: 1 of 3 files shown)


Chemaclass (approved):
Amazing!

PR #2184, opened 2026-05-27
fix: seed clojure aliases for file runs

### LINKED ISSUE #2129: `defmethod` resulting to name resolution failure
- Phel https://github.com/phel-lang/phel-lang/commit/b456c216e1cf22eacad025fcceb8ace8dab90228
- PHP 8.4.16 (cli)

`phel-lang/clojure-test-suite` is currently behind few commits from upstream which I'm testing by running `jank-lang/clojure-test-suite` against `phel-lang/phel-lang` main branch (b456c21).

The
https://github.com/jank-lang/clojure-test-suite/commit/a7b3857d39475a20dfedd6cc6f29e8beca0479be commit modified the portable exception multimethod in way that doesn't seem to be compatible with Phel currently, test run fails with 132 files failing to compile due to `Cannot resolve symbol 'p/thrown?'`.

Minimal example (adapted from jank-lang/clojure-test-suite [portability.cljc](https://github.com/jank-lang/clojure-test-suite/blob/fc224d7e351581896aa1edc81b81534356ccc857/test/clojure/core_test/portability.cljc#L111-L126))  `issue-demo3.phel`:

```
(ns issue-demo3
  (:require [clojure.test :as t]))

(defmethod t/assert-expr 'p/thrown?
     [msg form]
     (let [body (rest form)]
       `(try
          (let [result# (do ~@body)]
            (t/do-report {:type :fail
                          :message ~msg
                          :expected '~form
                          :actual result#}))
          (catch ~'Throwable e#
            (t/do-report {:type :pass
                          :message ~msg
                          :expected '~form
                          :actual e#})
            e#))))
```

Running this results to odd function resolution with `-methods` postfix added which fails:

```
./bin/phel run issue-demo3.phel
[PHEL001] Cannot resolve symbol 't/assert-expr-methods' in /home/user/dev/phel/phel-lang/issue-demo3.phel:4
```

When evaluated in REPL the code leads to odd output also while it doesn't have the same error:
```
./bin/phel
Welcome to the Phel Repl (v0.40.0-beta#b456c21)
user:1> (ns issue-demo3
....:2>   (:require [clojure.test :as t]))
nil
issue-demo3:3>
nil
issue-demo3:4> (defmethod t/assert-expr 'p/thrown?
....:5>      [msg form]
....:6>      (let [body (rest form)]
....:7>        `(try
....:8>           (let [result# (do ~@body)]
....:9>             (t/do-report {:type :fail
....:10>                           :message ~msg
....:11>                           :expected '~form
....:12>                           :actual result#}))
....:13>           (catch ~'Throwable e#
....:14>             (t/do-report {:type :pass
....:15>                           :message ~msg
....:16>                           :expected '~form
....:17>                           :actual e#})
....:18>             e#))))
{not Printer cannot print this type: Closure, thrown? Printer cannot print this type: Closure, thrown-with-msg? Printer cannot print this type: Closure, output? Printer cannot print this type: Closure, :default Printer cannot print this type: Closure, p/thrown? Printer cannot print this type: Closure}
```

SauronBot:
## Summary

- seed matching bundled phel.* namespaces when file-run scripts depend on clojure.* aliases
- keep the original clojure.* dependency seed so normal resolution semantics stay unchanged
- add a run-command regression for the reported clojure.test assert-expr repro

## Verification

- ./bin/phel run /tmp/issue-demo3.phel
- ./vendor/bin/phpunit tests/php/Unit/Run/Application/BundledNamespaceDetectorTest.php tests/php/Unit/Run/Application/FileRunnerTest.php tests/php/Integration/Run/Command/Run/RunCommandTest.php
- composer test-core
- composer test-compiler
- composer test-quality
- composer validate --strict --quiet --no-interaction --no-check-all --no-check-lock
- composer test-agents
- composer test-tools
- find docs/examples -name "*.phel" -type f | sort | xargs -I {} bash -c './bin/phel run "$1" >/dev/null' _ {}
- vendor/bin/phpbench run --iterations=3 --warmup=1 --report=aggregate --progress=plain

Closes #2129

### DIFF
--- a/src/php/Run/Application/BundledNamespaceDetector.php
+++ b/src/php/Run/Application/BundledNamespaceDetector.php
@@ -9,6 +9,7 @@ use function array_values;
 use function file_get_contents;
 use function preg_match_all;
 use function str_replace;
+use function str_starts_with;
 use function substr;
 
 final readonly class BundledNamespaceDetector
@@ -59,5 +60,32 @@ final readonly class BundledNamespaceDetector
         }
 
         return array_values(array_intersect($bundled, array_keys($referenced)));
+    }
+
+    /**
+     * @param list<string> $dependencies
+     *
+     * @return list<string>
+     */
+    public function remapClojureDependencies(array $dependencies): array
+    {
+        $candidates = [];
+        foreach ($dependencies as $dependency) {
+            $normalized = str_replace('\\', '.', $dependency);
+            if (str_starts_with($normalized, 'clojure.')) {
+                $candidates['phel.' . substr($normalized, 8)] = true;
+            }
+        }
+
+        if ($candidates === []) {
+            return [];
+        }
+
+        $bundled = $this->bundledNamespaces->all();
+        if ($bundled === []) {
+            return [];
+        }
+
+        return array_values(array_intersect($bundled, array_keys($candidates)));
     }
 }
--- a/src/php/Run/Application/BundledNamespaceDetector.php
+++ b/src/php/Run/Application/BundledNamespaceDetector.php
@@ -10,16 +10,25 @@ use function file_get_contents;
 use function preg_match_all;
 use function str_replace;
 use function str_starts_with;
+use function strlen;
 use function substr;
 
+/**
+ * Selects bundled `phel.*` namespaces that file runs must seed before
+ * evaluating an ad-hoc script.
+ */
 final readonly class BundledNamespaceDetector
 {
+    private const string CLOJURE_PREFIX = 'clojure.';
+
     /**
      * Matches fully qualified bundled references in the source: `phel.X/sym`
      * or the legacy `phel\X/sym`. The trailing slash anchors the match to
      * actual call/var references rather than comments containing the prefix.
      */
     private const string FQN_PATTERN = '#\bphel[.\\\\][A-Za-z0-9_\-.\\\\]+/#u';
+
+    private const string PHEL_PREFIX = 'phel.';
 
     public function __construct(
         private BundledNamespaces $bundledNamespaces,
@@ -48,21 +57,19 @@ final readonly class BundledNamespaceDetector
             return [];
         }
 
-        $bundled = $this->bundledNamespaces->all();
-        if ($bundled === []) {
-            return [];
-        }
-
         $referenced = [];
         foreach ($matches[0] as $token) {
             $ns = str_replace('\\', '.', substr($token, 0, -1));
             $referenced[$ns] = true;
         }
 
-        return array_values(array_intersect($bundled, array_keys($referenced)));
+        return $this->intersectBundled($referenced);
     }
 
     /**
+     * Returns bundled `phel.*` namespaces matching Clojure-compatible
+     * dependencies such as `clojure.test`.
+     *
      * @param list<string> $dependencies
      *
      * @return list<string>
@@ -72,8 +79,8 @@ final readonly class BundledNamespaceDetector
         $candidates = [];
         foreach ($dependencies as $dependency) {
             $normalized = str_replace('\\', '.', $dependency);
-            if (str_starts_with($normalized, 'clojure.')) {
-                $candidates['phel.' . substr($normalized, 8)] = true;
+            if (str_starts_with($normalized, self::CLOJURE_PREFIX)) {
+                $candidates[self::PHEL_PREFIX . substr($normalized, strlen(self::CLOJURE_PREFIX))] = true;
             }
         }
 
@@ -81,6 +88,16 @@ final readonly class BundledNamespaceDetector
             return [];
         }
 
+        return $this->intersectBundled($candidates);
+    }
+
+    /**
+     * @param array<string, true> $candidates
+     *
+     * @return list<string>
+     */
+    private function intersectBundled(array $candidates): array
+    {
         $bundled = $this->bundledNamespaces->all();
         if ($bundled === []) {
             return [];
--- a/src/php/Run/Application/FileRunner.php
+++ b/src/php/Run/Application/FileRunner.php
@@ -74,6 +74,8 @@ final readonly class FileRunner
             CompilerConstants::PHEL_CORE_NAMESPACE,
             ...$this->bundledNamespaceDetector->detect($filename),
             ...$this->bundledNamespaceDetector->remapClojureDependencies($dependencies),
+            // Missing seeds are ignored by dependency resolution, so keeping
+            // raw `clojure.*` deps preserves user-defined Clojure namespaces.
             ...$dependencies,
         ]));
     }
--- a/src/php/Run/Application/FileRunner.php
+++ b/src/php/Run/Application/FileRunner.php
@@ -67,11 +67,14 @@ final readonly class FileRunner
      */
     private function primarySeeds(NamespaceInformation $scriptInfo, string $filename): array
     {
+        $dependencies = $scriptInfo->getDependencies();
+
         return array_values(array_unique([
             $scriptInfo->getNamespace(),
             CompilerConstants::PHEL_CORE_NAMESPACE,
             ...$this->bundledNamespaceDetector->detect($filename),
-            ...$scriptInfo->getDependencies(),
+            ...$this->bundledNamespaceDetector->remapClojureDependencies($dependencies),
+            ...$dependencies,
         ]));
     }
 
--- a/tests/php/Integration/Run/Command/Run/Fixtures/clojure-test-alias-assert-expr-script.phel
+++ b/tests/php/Integration/Run/Command/Run/Fixtures/clojure-test-alias-assert-expr-script.phel
@@ -1,5 +1,6 @@
 (ns clojure-test-alias-assert-expr-script
-  (:require [clojure.test :as t]))
+  (:require [clojure.test :as t]
+            [clojure.test :as p]))
 
 (defmethod t/assert-expr 'p/thrown?
   [msg form]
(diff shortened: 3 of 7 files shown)


src/php/Run/Application/BundledNamespaceDetector.php
```diff
@@ -60,4 +61,31 @@ public function detect(string $filename): array
...
+     *
+     * @return list<string>
+     */
+    public function remapClojureDependencies(array $dependencies): array
+    {
+        $candidates = [];
+        foreach ($dependencies as $dependency) {
+            $normalized = str_replace('\\', '.', $dependency);
+            if (str_starts_with($normalized, 'clojure.')) {
+                $candidates['phel.' . substr($normalized, 8)] = true;
```
Chemaclass: Magic `8` = `strlen('clojure.')`. Extract a named constant to match the codebase pattern (`SymbolResolver::CLOJURE_LANG_PREFIX`, `NsSymbol::remapClojureNamespace`):

```php
private const string CLOJURE_PREFIX = 'clojure.';
// ...
$candidates['phel.' . substr($normalized, strlen(self::CLOJURE_PREFIX))] = true;
```

Keeps intent obvious and avoids drift if the prefix ever changes.

Chemaclass (commented):
Solid fix; correct seeding strategy. A few polish suggestions inline.

**Summary**
- Logic is correct: both `clojure.X` and matching bundled `phel.X` seeded, analyzer-time `NsSymbol::remapClojureNamespace` then rewrites refs. `array_unique` dedups.
- Order preserved via `array_intersect($bundled, ...)`. Handles both `clojure.test` and legacy `clojure\test` forms.
- Short-circuits before `bundledNamespaces->all()` when no clojure deps present (good for cold-start).

**Cross-cutting nits**
- `src/php/Run/CLAUDE.md` BundledNamespaceDetector bullet should be updated: it now also seeds bundled targets for `clojure.*` requires, not only FQN-referenced bundles.
- PR body `Refs #2129` → `Closes #2129` if this fully resolves the issue.

src/php/Run/Application/BundledNamespaceDetector.php
```diff
@@ -60,4 +61,31 @@ public function detect(string $filename): array
 
         return array_values(array_intersect($bundled, array_keys($referenced)));
     }
+
+    /**
+     * @param list<string> $dependencies
+     *
+     * @return list<string>
+     */
+    public function remapClojureDependencies(array $dependencies): array
```
Chemaclass, line 77: Class now has dual responsibility: FQN scan in source (`detect`) and clojure->phel remap of declared deps (`remapClojureDependencies`). Both legitimately consume `BundledNamespaces`, so co-location is OK, but consider updating the class-level docblock to reflect the new responsibility — currently there's none, and the existing single-purpose mental model in `Run/CLAUDE.md` no longer matches.

src/php/Run/Application/BundledNamespaceDetector.php
```diff
@@ -60,4 +61,31 @@ public function detect(string $filename): array
...
+        if ($candidates === []) {
+            return [];
+        }
+
+        $bundled = $this->bundledNamespaces->all();
+        if ($bundled === []) {
+            return [];
+        }
+
+        return array_values(array_intersect($bundled, array_keys($candidates)));
```
Chemaclass, line 106: Minor: `detect()` and `remapClojureDependencies()` now share the exact same tail (`$bundled = ...->all(); if ($bundled === []) return []; return array_values(array_intersect($bundled, array_keys($X)));`). Could extract a private `intersectBundled(array $candidates): array`. Fine to leave if you prefer no abstraction.

tests/php/Integration/Run/Command/Run/Fixtures/clojure-test-alias-assert-expr-script.phel
```diff
@@ -0,0 +1,18 @@
...
+         (t/do-report {:type :fail
+                       :message ~msg
+                       :expected '~form
+                       :actual result#}))
+       (catch ~'Throwable e#
+         (t/do-report {:type :pass
+                       :message ~msg
+                       :expected '~form
+                       :actual e#})
+         e#))))
```
Chemaclass, line 19: Fixture defines `assert-expr` but never invokes it, so the only e2e signal is "no crash". Add a `(t/deftest …)` / `(t/is …)` form (or a `(t/run-tests)` call) so the test actually exercises the registered method. Without it, removing the seed could regress silently and this fixture would still pass.

tests/php/Integration/Run/Command/Run/RunCommandTest.php
```diff
@@ -85,6 +85,15 @@ public function test_run_by_filename_resolves_bundled_namespace_fqn_without_requ
         self::assertStringContainsString('phel.async/delay resolved', $output);
     }
 
+    public function test_run_by_filename_resolves_clojure_test_alias_before_script_eval(): void
+    {
+        $output = $this->captureRunOutput(
+            __DIR__ . '/Fixtures/clojure-test-alias-assert-expr-script.phel',
+        );
+
+        self::assertSame('', $output);
```
Chemaclass: `assertSame('', $output)` is a weak assertion — any script that doesn't print survives this. Pair with a fixture that emits a visible token (e.g. `(println "ok")` after a `t/is`) and assert on it, so the test fails if the seeded `phel.test` ns isn't actually loaded.

tests/php/Unit/Run/Application/BundledNamespaceDetectorTest.php
```diff
@@ -106,6 +106,28 @@ public function test_returns_empty_when_file_unreadable(): void
...
+    public function test_remap_skips_bundled_discovery_when_no_clojure_dependencies_exist(): void
+    {
+        $buildFacade = $this->createMock(BuildFacadeInterface::class);
+        $buildFacade->expects(self::never())->method('getNamespaceFromDirectories');
+
+        $commandFacade = $this->createStub(CommandFacadeInterface::class);
+
+        $detector = new BundledNamespaceDetector(new BundledNamespaces($buildFacade, $commandFacade));
+
+        self::assertSame([], $detector->remapClojureDependencies(['phel.test']));
```
Chemaclass, line 128: Coverage gap: no test for the case where a script has both a `clojure.test` require AND uses FQN `phel.test/foo`. `array_unique` in `FileRunner::primarySeeds` handles dedup, but an explicit unit test on the detector confirming no duplicates in `remapClojureDependencies` output (when called with already-`phel.X` inputs) would lock that behavior.

src/php/Run/Application/FileRunner.php
```diff
@@ -67,11 +67,14 @@ public function run(string $filename): void
...
     {
+        $dependencies = $scriptInfo->getDependencies();
+
         return array_values(array_unique([
             $scriptInfo->getNamespace(),
             CompilerConstants::PHEL_CORE_NAMESPACE,
             ...$this->bundledNamespaceDetector->detect($filename),
-            ...$scriptInfo->getDependencies(),
+            ...$this->bundledNamespaceDetector->remapClojureDependencies($dependencies),
+            ...$dependencies,
```
Chemaclass, line 79: Question: when `clojure.test` is also seeded (via `...$dependencies`), what does `getDependenciesForNamespace` do if no file named `clojure/test.phel` exists under any srcDir? If it silently skips, fine; if it throws on missing-ns, the seed would need pre-filtering. Worth a one-line comment here explaining the contract you're relying on.

src/php/Run/Application/BundledNamespaceDetector.php
SauronBot: Done in 6addd196. Added `CLOJURE_PREFIX` and `PHEL_PREFIX`, and replaced the magic offset with `strlen(self::CLOJURE_PREFIX)`.

src/php/Run/Application/BundledNamespaceDetector.php
SauronBot: Done in 6addd196. Added a class docblock for the expanded responsibility and updated `src/php/Run/CLAUDE.md` to document both FQN bundle seeding and `clojure.*` require remapping.

src/php/Run/Application/BundledNamespaceDetector.php
SauronBot: Done in 6addd196. Extracted the shared `intersectBundled()` helper so both paths use the same bundled namespace filtering.

tests/php/Integration/Run/Command/Run/Fixtures/clojure-test-alias-assert-expr-script.phel
SauronBot: Done in 6addd196. The fixture now invokes the registered `p/thrown?` assert expression through `t/is`, so this is more than a no-crash check.

tests/php/Integration/Run/Command/Run/RunCommandTest.php
SauronBot: Done in 6addd196. The fixture now prints `clojure-test-alias-ok` after the assertion path runs, and the PHP integration test asserts that token.

tests/php/Unit/Run/Application/BundledNamespaceDetectorTest.php
SauronBot: Done in 6addd196. Added coverage for equivalent `clojure.test` / `clojure\\test` dependencies plus an already-`phel.test` input returning one bundled seed.

src/php/Run/Application/FileRunner.php
SauronBot: Done in 6addd196. Added the contract comment in `FileRunner`: missing seeds are ignored by dependency resolution, so keeping raw `clojure.*` deps preserves user-defined Clojure namespaces.

Chemaclass (approved):

PR #2214, opened 2026-05-29
perf(compiler): inline single-expression pure defn at call site

### LINKED ISSUE #2135: perf(compiler): inline single-expression pure defn at call site
## Motivation

Direct positional dispatch landed in #2117. Next step: when callee body is a single expression and all params are used ≤1 time or substituted by literals, inline the body at the call site. Skips one PHP frame + enables downstream peepholes (constant folding through inlined args).

## Example

```phel
(defn inc [x] (+ x 1))
;; call site
(inc 5)
;; current: dispatches to fn, returns 6
;; target: emits (5 + 1), then constant folder collapses to 6
```

## Conditions

- Callee is a `defn` (resolved via `GlobalCallTarget::isGlobalFnCall`)
- Callee has exactly one arity, no variadic
- Callee body is one expression after `DoSimplifier` pass
- No `recur` in body
- Callee is **pure** per `PureExpressionDetector` (no side-effecting calls)
- Callee not marked `^:async` / `^:memoize` / `^{:memoize-lru N}` (`.claude/rules/before-suggesting-code.md`)
- Caller is not inside a `loop`/`recur` frame that would change tail semantics

## Files

- `src/php/Compiler/Domain/Analyzer/TypeAnalyzer/Simplification/` (new `CallInliner.php`)
- Or `src/php/Compiler/Domain/Analyzer/TypeAnalyzer/SpecialForm/InvokeSymbol.php` for the inlining branch

## Plumbing

The defn `FnNode` side-table landed in #2202 (`perf(analyzer): add defn FnNode side-table to global environment`). `GlobalEnvironmentRegistry` now exposes the analysed body of every `defn` across files — that is the hook the inliner needs. No new meta key required; read the `FnNode` directly.

Within same file, inliner sees AST directly.

## Relation to #2205

#2205 (`fold 3-arg reduce on pure fn + literal coll`) is the same compile-time-evaluate-pure-fn idea, scoped to `reduce`. This issue generalises the pattern to any call site. The inliner could subsume #2205's special-cased path once it lands.

## Bench-first gate

Per the `bench-before-perf-refactor` rule, land a permanent bench **before** the implementation PR. Mirror the shape of `CollectionHoistingBench` (#2208):

- `CallInliningBench` with three subjects:
  - **status_quo**: dispatched call through the current emit path
  - **inlined**: hand-written equivalent showing the target emission
  - **direct**: bare expression (lower bound)
- Decision: ship only if the inlined steady-state delta over status_quo justifies the analyser complexity. Reject if delta is within noise.

## Acceptance

- `CallInliningBench` lands first, in its own PR.
- Impl PR: `composer phpbench-ref` shows positive delta vs that bench.
- Compiled stdlib output unchanged where inlining off (gate behind `setOptimizationLevel >= 2`).
- Integration test exercising recursion / shadowed names confirms inliner respects scope.

## Risk

Macro hygiene — see `.claude/rules/macro-hygiene.md`. Param substitution must rename gensyms to avoid capture. Use existing `gensym` infra.

## Related

- #2202 — `FnNode` side-table (plumbing dependency, already merged)
- #2205 — 3-arg `reduce` fold (narrower precedent, may be subsumed)
- #2208 — bench-first pattern reference

## Status

Previous "Blocked by #2132 / Blocks #2141" sequence is **obsolete**: both dependencies merged ahead of this issue. No remaining hard blockers; bench-first gate is the only sequencing requirement now.

Chemaclass:
## 🤔 Background

Direct positional dispatch (#2117) and the `defn` `FnNode` side-table (#2202) landed earlier. The next step (#2135): when a call site targets a single-expression pure `defn`, splice the body in instead of dispatching through the resolved `AbstractFn`. This skips one PHP frame and exposes the body to downstream constant folding.

The bench-first gate `CallInliningBench` (#2209) already established the win. Measured here:

| subject | per-call |
|---|---|
| `status_quo` (dispatch) | ~0.074 μs |
| `inlined` | ~0.020 μs |
| `direct` (folded) | ~0.020 μs |

~3.7× faster at the call site.

While wiring this up, `PureExpressionDetector` turned out unsuitable as the body gate: it proves purity by *full constant evaluation*, so a symbolic body like `(+ x 1)` (free param) is reported impure. A dedicated structural oracle was needed to reach the symbolic frame-skip case the issue targets.

## 💡 Goal

Inline calls to single-arity pure `defn`s at the call site, gated behind `CompileOptions::setOptimizationLevel(2)`, leaving the default (level 0) output byte-identical.

- `(my-inc 5)` → `6` (inlined, then constant-folded)
- `(my-inc x)` → `(+ x 1)` (inlined, `my-inc` frame skipped)
- recursive / variadic / memoised / `^:async` / side-effecting callees → unchanged dispatch

## 🔖 Changes

- **`SymbolicPurityDetector`** — structural purity oracle: a node is pure when it only references literals / locals / globals and calls a closed allowlist of side-effect-free `phel.core` fns (or `php/` infix ops) on pure arguments. Mirrors `ConstantFolder`'s vetted scalar-op set, but decides purity *symbolically* (over free vars) rather than by evaluation.
- **`CallInliner`** — reads the callee body from the `GlobalEnvironment` `defFnNode` side-table and, when the gates pass (opt ≥ 2, single arity, non-variadic, no `recur`, not memoised/`^:async`, one pure expression body, pure args, arity match, not in an enclosing recur tail), rebuilds the body onto the call-site environment with each parameter replaced by its argument node, then constant-folds the result. Anything outside the node whitelist aborts the inline (falls back to dispatch). Because spliced nodes are already analysed, it works across namespaces and never re-enters the analyzer — no re-resolution, no capture.
- **`InvokeSymbol`** — adds the `CallInliner::tryInline()` branch for `GlobalVarNode` callees, after arg analysis and param-tag checks.
- Unit tests for both new classes; runtime acceptance tests (literal fold, symbolic frame-skip, shadowed-name hygiene, recursion/side-effect exclusion, opt-0 dispatch fallback).
- `CHANGELOG.md` (Unreleased) and `src/php/Compiler/CLAUDE.md` updated.

Closes #2135

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file.
 
 Opt-in (`CompileOptions::setOptimizationLevel(2)`):
 - `TailCallRewriter`: detect self-recursive `defn` calls in tail position and rewrite them into an implicit `recur` loop. Eliminates per-iteration PHP stack frames. Variadic and multi-arity defs are skipped. Stack-trace shape inside the loop changes from N frames to one (#2141)
+- `CallInliner`: splice the body of a single-arity pure `defn` at the call site instead of dispatching through the resolved `AbstractFn`, skipping one PHP frame and exposing the body to constant folding (e.g. `(my-inc 5)` → `6`). Only fires when every argument is pure (`SymbolicPurityDetector`) and the body is one pure expression; variadic, recursive, memoised and `^:async` defs are skipped (#2135)
 
 Compile-time folding and simplification:
 - `ConstantFolder`: fold pure `phel.core` calls on literal args (comparisons, boolean predicates, bitwise / `bit-*`, `count` / `first` / `last` / `nth` on literal collections, `str`, `min` / `max` / `mod` / `quot` / `rem` / `abs`). Skipped when runtime would throw or promote (#2088)
--- a/src/php/Compiler/CLAUDE.md
+++ b/src/php/Compiler/CLAUDE.md
@@ -65,6 +65,8 @@ Core compilation pipeline: Phel source -> tokens -> AST -> analyzed nodes -> PHP
 
 The simplification pass runs **after** the analyser-level `ConstantFolder`. It currently drops pure non-tail expressions from `(do ...)` bodies; purity is decided by `PureExpressionDetector`, which delegates `CallNode` checks to `ConstantFolder` so a call is "pure" only when the folder can statically compute its value (calls that would throw at runtime stay un-dropped).
 
+`Simplification/` also hosts the call inliner (opt level >= 2, issue #2135): `InvokeSymbol` calls `CallInliner::tryInline()` for a `GlobalVarNode` callee. When the callee is a single-arity, non-variadic, non-recursive, non-memoised pure `defn` (body read from the `GlobalEnvironment` `defFnNode` side-table) and every argument is pure, the body is spliced at the call site with parameters replaced by the argument nodes and envs rebased onto the caller. Purity here is structural via `SymbolicPurityDetector` (a closed allowlist of side-effect-free `phel.core` / `php/` ops), unlike `PureExpressionDetector` which proves purity by full evaluation. At opt level < 2 `tryInline` returns `null` immediately, so default output is unchanged.
+
 ## Structure
 
 ```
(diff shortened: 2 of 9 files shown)


Chemaclass:
Both review points addressed in 44fbfe22:

1. **Misleading test comment** — corrected. `str` is not in `SymbolicPurityDetector::PURE_CORE_FNS`, so `shout` is **not** inlined; the test now compiles `(shout "hi")` and asserts the dispatch is still emitted, alongside the runtime sanity check.
2. **Per-call `CallInliner` allocation** — `CallInliner` is now a constructor-injected `private readonly` field on `InvokeSymbol` (defaulting to `new CallInliner()`), so it is built once. Note special-form handlers are already cached per symbol name in `AnalyzePersistentList::symbolAnalyzerCache`, so `InvokeSymbol` (and now its inliner) is reused across call sites rather than reallocated.

PR #2325, opened 2026-06-03
ref: repo-wide code & documentation quality pass

Chemaclass:
## 🤔 Background

A repo-wide pass to raise code and documentation quality so that a human Phel/PHP expert can read and maintain the codebase with zero friction — independent of any agentic tooling. Both source trees are covered: the PHP compiler/runtime (`src/php/`) and the Phel core library (`src/phel/`).

Driven by a per-module audit, then applied in verified, gated, bisectable waves. Every finding was checked against the real code before applying (the audit produced false positives — e.g. flagging the PHP 8.4 builtin `array_any` as undefined — all of which were rejected). Each wave was committed only after the full quality gate (php-cs-fixer, psalm lvl1, phpstan lvl5, rector, compiler suite, core suite) passed.

## 💡 Goal

Best-achievable maintainability with **no behavior change** outside three genuine bugs the work surfaced (each fixed with a regression test).

## 🔖 Changes

Phased into focused commits:

- **docs** — explanatory class/method docblocks (intent, caching, invariants, side effects) across PHP modules, and `:doc`/`:see-also`/`:example` metadata across Phel namespaces. Pure additions, zero behavior risk.
- **ref (cleanups)** — verified dead-code removal, native union types on private methods, clearer local/private names, interface/impl return-type alignment.
- **ref (structural)** — behavior-identical helper extractions (dedup), `readonly` on stateless final classes; all signature/API-changing suggestions deferred.
- **ref (repl)** — reuse the existing public `phel.core/canonical-ns` instead of six raw `(php/:: Munge (canonicalNs ...))` interop calls.
- **test** — ~40 focused unit/integration tests for previously untested utilities (Munge, VersionFinder, ResourceUsageFormatter, PhelProjectDirectory, Bencode, ArgvInputSanitizer, CpuCountDetector boundaries, formatter indenters, quasiquote empty collections, schema coercion, string edge cases). Compiler suite 4020 → 4173; core 5863 → 5894.

Three real bugs surfaced and fixed (with tests):

- `if-let`/`when-let`/`if-some`/`when-first` were **non-hygienic** — they bound an internal gensym but emitted the literal symbol `temp-sym`, silently shadowing a user binding of the same name. Switched to `temp-sym#` auto-gensym.
- Quasiquoting an **empty list** (`` `() ``) threw `Index out of bounds` (`get(0)` with no count guard); now yields an empty list like empty vectors/maps.
- `str/index-of` on an **empty string** leaked a raw PHP `ValueError` from `mb_strpos`; now returns `nil` (and `0` for an empty needle), matching Clojure.

CHANGELOG updated under `## Unreleased`.

### Known follow-ups (documented, intentionally not in this PR)

Surfaced by the new tests but out of scope here (need behavior decisions / hard-to-test environments): `FswatchWatcher::parseLine()` always reports `KIND_MODIFIED`; `ErrorLog` and `RealFilesystem::clearAll()` swallow I/O errors silently.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- `if-let`/`when-let`/`if-some`/`when-first`: now hygienic — a user binding named `temp-sym` is no longer shadowed by the macros' internal temporary
+- Quasiquote of an empty list (`` `() ``) no longer throws `Index out of bounds`; it yields an empty list, like empty vectors/maps already did
+- `str/index-of`: returns `nil` for an empty string instead of leaking a raw PHP `ValueError` (`(index-of "" "a")` => `nil`, `(index-of "" "")` => `0`)
 - Lexer: error/source-map columns are now counted in code points, not bytes, so locations are correct for multibyte (UTF-8) source
 - Printer: structs print with the `.` separator (`(my.ns.point 1 2)`) instead of `\` (#2255)
 - Docs: function `:example` outputs now match what the REPL prints, so `phel doc` and the phel-lang.org API reference are accurate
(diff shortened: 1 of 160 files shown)


Chemaclass:
Superseded by a risk-tiered split (same commits, no work lost):
- #2326 — behavior fixes (the 3 bugs + regression tests)
- #2327 — behavior-identical refactors
- #2328 — docs/metadata + ~40 coverage tests

Each branch builds on current `main` and passes the full quality gate independently.

PR #2366, opened 2026-06-07
perf(compiler): use strtr in Munge::encodeWithMap (~4.9x faster per identifier)

### LINKED ISSUE #2364: perf(compiler): use strtr in Munge::encodeWithMap (~4.9x faster per identifier)
## 🤔 Background

`Munge::encodeWithMap()` (`src/php/Shared/Munge.php:106-118`) runs on **every emitted identifier** — every local var, global, function name, keyword/symbol — so it is one of the hottest helpers in the emitter phase.

Current impl rebuilds the search/replace arrays on every single call:

```php
private function encodeWithMap(string $str, array $mapping): string
{
    return str_replace(
        array_keys($mapping),   // new array allocated per call
        array_values($mapping), // new array allocated per call
        $str,
    );
}
```

The `SYMBOL_MAPPING` keys are all distinct single characters and none of the replacement strings contain a key character, so `str_replace` (sequential) and `strtr` (single-pass longest-match) produce **identical** output — but `strtr` takes the assoc map directly, no per-call array allocation.

## 💡 Goal

Replace `str_replace(array_keys(...), array_values(...), $str)` with `strtr($str, $mapping)`.

## 🔖 Changes

- `encodeWithMap()` → `return strtr($str, $mapping);`
- Existing `decodeNs()` passes `array_flip($this->nsMapping)` — still a valid `strtr` map, keep as-is.

## 📊 Validation (phpbench, head-to-head, 100k revs ×5 its)

| impl | mode |
|---|---|
| current `str_replace` | 3.40μs |
| proposed `strtr` | 0.70μs |

**~4.9× faster** per encode batch. Behavior verified identical across 21 representative symbol names (single chars, `?`, `/`, `.`, `\`, `$`, mixed). No output change, pure speedup on a per-identifier hot path.

Baseline reference: cold `phel\core` compile ≈ 33ms (`PhelBench::bench_core_file_compilation`).

JesusValeraDev:
## Summary

`Munge::encodeWithMap()` runs on every emitted identifier (local vars, globals, function names, keywords/symbols), making it one of the hottest helpers in the emitter phase. The current impl rebuilds the search/replace arrays on every call via `array_keys()`/`array_values()`.

`SYMBOL_MAPPING` keys are all distinct single characters and no replacement string contains a key character, so `str_replace` (sequential) and `strtr` (single-pass longest-match) produce identical output — but `strtr` takes the assoc map directly, with no per-call array allocation.

## Changes

- `encodeWithMap()` → `return strtr($str, $mapping);`
- `decodeNs()` still passes `array_flip($this->nsMapping)`, a valid `strtr` map — kept as-is.

## Validation

phpbench head-to-head (100k revs ×5 its):

| impl | mode |
|---|---|
| current `str_replace` | 3.40μs |
| proposed `strtr` | 0.70μs |

**~4.9× faster** per encode batch. All 24 existing Munge unit tests pass.

Closes #2364

### DIFF
--- a/src/php/Shared/Munge.php
+++ b/src/php/Shared/Munge.php
@@ -110,10 +110,6 @@ final readonly class Munge implements MungeInterface
      */
     private function encodeWithMap(string $str, array $mapping): string
     {
-        return str_replace(
-            array_keys($mapping),
-            array_values($mapping),
-            $str,
-        );
+        return strtr($str, $mapping);
     }
 }


Chemaclass (approved):
nice!

PR #2388, opened 2026-06-09
fix: map built-output runtime errors back to Phel source via sibling .map files

Chemaclass:
## 🤔 Background

`phel build` writes a `<file>.php.map` (VLQ source map) and a `<file>.phel` (source copy) next to every compiled file, but nothing ever read them: the error printer's `SourceMapExtractor` only looks at the first two lines of the failing PHP file for inline `// ` metadata comments, which built files don't have. So runtime errors in built output always pointed at generated PHP lines, and the `.map` artifacts were dead weight.

On top of that, the inline path was broken too: eval temp files start with `<?php` (and optionally `declare(ticks=1);`), so the metadata comments sit on lines 2-3 (or 3-4), where the two-line reader never found them.

## 💡 Goal

Make the existing source-map artifacts earn their keep: runtime errors in built output report `out/foo.phel:42` instead of the generated PHP line, and inline source maps in eval temp files are detected again.

## 🔖 Changes

- `SourceMapExtractor` now scans the first header lines of the compiled file for the `// <source>` + `// ;;<mappings>` comment pair (skipping the `<?php` opener and declare statements), and falls back to the sibling `<file>.map` + `<file>.phel` artifacts written by `FileCompiler` for built output.
- `SourceMapInformation` is now a normalized read model (original filename, raw VLQ mappings, 1-based code start line) instead of two raw comment lines, so `FilePositionExtractor` does a single well-defined line translation: `traceLine - (codeStartLine - 1)` through `SourceMapConsumer`.
- Unit tests for both extraction layouts plus edge cases (missing file, plain PHP file, orphan `.map` without `.phel` sibling), and an end-to-end integration test that compiles a fixture via `BuildFacade` and asserts the mapped position lands on the right `.phel` line.
- CHANGELOG entry under `## Unreleased`; Command module doc updated.

Follow-up candidate (not in this PR): `phel run` with the compiled-code cache enabled compiles with source maps disabled (`FileEvaluator::compileForCache`), so cache-hit traces remain unmapped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Fixed
+
+- Runtime errors in `phel build` output now map back to the Phel source: the error printer reads the sibling `<file>.php.map` and `<file>.phel` artifacts (previously written but never consumed) and reports `out/foo.phel:42` instead of the generated PHP line
+- Inline source maps in eval temp files are detected again: the extractor now scans past the `<?php` opener (and optional `declare` statements) for the metadata comments instead of only reading the first two lines
+
 ## [0.43.0](https://github.com/phel-lang/phel-lang/compare/v0.42.0...v0.43.0) - 2026-06-09
 
 ### Added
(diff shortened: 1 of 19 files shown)


Chemaclass:
Good catch on the `str_replace` suffix issue. Fixed in e8239986: the sibling path convention now lives in `Phel\Shared\SourceMap\SourceMapSiblings`, which derives the `.phel` path with `str_ends_with` + `substr` (suffix-only) and is shared by both the Build writer (`FileCompiler`) and the Command reader (`SourceMapExtractor`). Covered by `SourceMapSiblingsTest::test_source_file_only_touches_the_suffix`.

PR #2417, opened 2026-06-11
docs: simplify, de-duplicate, and correct docs/ guides

Chemaclass:
## 🤔 Background

The `docs/` guides had accumulated verbosity, content duplicated across several files, and a handful of stale examples that no automated check ever exercised. A cross-file audit (verifying every claim against the current compiler, core library, and CLI) surfaced both redundancy and concrete factual drift.

## 💡 Goal

Make the docs tighter, internally consistent, and correct, without losing any technical substance.

## 🔖 Changes

- **Trim verbosity** across all guides; keep every code block, table, and example.
- **De-duplicate** content repeated across files down to one canonical home plus cross-links (interop syntax, namespace-separator policy, reader conditionals, transients, optimization levels, cache/`.phel` relocation).
- **Fix stale/incorrect content**, verified against code:
  - `delay`/`force`/`delay?` exist in core (clojure-migration)
  - `find` signature `[pred-or-coll coll-or-key]` (data-structures)
  - `withBuildConfig`/`withExportConfig` accept configurator closures (configuration)
  - `:map` closed-key is `:closed`, not `:closed?` (schema)
  - `[:re ...]` needs a `#"regex"` literal or a delimiter-wrapped PCRE string (schema)
  - lazy `map`/`filter` realize one element at a time, not in fixed 32-chunks (lazy-sequences)
  - shell completion uses `completion <shell>`, not the removed Symfony-4 `_complete --generate-hook` (cli)
  - FrankenPHP worker export shape `App\PhelGenerated\...::handleRequest` (deployment)
  - fixed broken in-page anchors `#pheledn`/`#pheltransit` (data-formats)
- Restore the user-facing `./vendor/bin/phel` convention and remove introduced em dashes.

Docs-only; no runtime or API change, so no CHANGELOG entry.

### DIFF
--- a/docs/agent-docs.md
+++ b/docs/agent-docs.md
@@ -13,11 +13,18 @@ composer require phel-lang/phel-lang
 ./vendor/bin/phel agent-install --all --with-examples  # also copy runnable example apps
 ```
 
-Platforms: `claude`, `cursor`, `codex`, `gemini`, `copilot`, `aider`. Omit the platform with `--auto` to install only for agents already present (`.claude/`, `.cursor/`, `AGENTS.md`, ...).
+Platforms: `claude`, `cursor`, `codex`, `gemini`, `copilot`, `aider`. `--auto` installs only for agents already present (`.claude/`, `.cursor/`, `AGENTS.md`, ...).
 
-`.agents/` docs (rules, task recipes, quick-syntax) are copied by default; `--no-docs` skips them. Example apps are excluded by default; `--with-examples` includes `.agents/examples/`.
+| Flag | Effect |
+|------|--------|
+| (default) | Copies `.agents/` docs (rules, task recipes, quick-syntax) |
+| `--no-docs` | Skip the `.agents/` docs tree |
+| `--with-examples` | Also copy `.agents/examples/` (excluded by default) |
+| `--force` | Skip the `<path>.pre-phel.bak` backup; overwrite existing skill files and `.agents/` tree |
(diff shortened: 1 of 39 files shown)


jasalt:
@Chemaclass I noted another project adding docs front-matter metadata for extra context on how up-to-date and human verified the docs are, might be useful here too. Haven’t looked in docs in this repo / Phel website for a while in depth so unsure but thought to share as it seemed like it could be useful in some cases.

https://github.com/nooga/let-go/issues/159#issuecomment-4597823854

Chemaclass:
Thanks for the pointer @jasalt! I read through the let-go thread. My take for this repo: `docs/` is intentionally contributor-facing and small (~30 short guides, all just overhauled in this PR), so I'd rather keep it lean than add frontmatter + hook tooling.

Concretely: an auto-stamped `last-verified` would just mirror `git log -1 --format=%cs -- <file>`, which we get for free, and deletion tombstones are covered by git history. The one field that carries information git doesn't have is a human-attestation date, but with the current maintenance bandwidth it would likely rot into noise.

The core documentation lives in the [phel-lang.org](https://github.com/phel-lang/phel-lang.org) repo, which is the user-facing surface where staleness signals would matter most — if this pattern lands anywhere, that's the place. And if a guide here ever feels stale, fixing it directly beats labeling it. Appreciate the idea though, good to have the pattern on the radar if `docs/` grows.

jasalt:
Sure. Thank’s for the insight!

PR #2447, opened 2026-06-14
feat(build): ship precompiled stdlib bundle in the PHAR for fast cold starts

### LINKED ISSUE #2443: perf: AOT-precompile phel.core into the PHAR to kill cold-start core compilation
## Goal

Precompile `phel.core` (and other bundled `phel.*` modules) to PHP **ahead-of-time** and ship the result inside the distributed PHAR, so a cold `phel run` no longer pays the runtime compilation cost of core.

## Problem

Today `phel.core` is **not** precompiled in the PHAR/binary. It is compiled at runtime on the first invocation, then cached in `.phel/cache/compiled/`. Measured cold-start cost of compiling core: **~300-800ms** — the single largest contributor to cold `phel run` latency.

Relevant code:
- `src/php/Run/Application/NamespaceLoader.php` — seeds + loads core/bundled namespaces
- `src/php/Build/Application/FileEvaluator.php` — compile-or-cache path (`evalFile()`)
- Cache lives under `.phel/cache/compiled/` (see `ENABLE_COMPILED_CODE_CACHE` in `BuildConfig.php`)

## Proposal

- During PHAR build (`build/phar.sh`), AOT-compile all bundled `phel.*` namespaces to PHP + their serialized namespace-environment data.
- Bundle these artifacts in the PHAR.
- On `phel run`, load the precompiled core directly (plain `require`) instead of going through lexer→parser→analyzer→emitter.

## Acceptance

- Cold `phel run <trivial-script>` from a freshly extracted PHAR shows no core-compilation cost (verify with `--with-time`).
- Bundled-module behavior unchanged (all core/integration tests green).
- PHAR build produces and embeds the precompiled artifacts.

## Impact

Highest-ROI cold-start win: removes ~300-800ms flat from first run.

Chemaclass:
## 🤔 Background

A cold `phel run`/`phel test` in a fresh project pays ~1s recompiling the bundled `phel.*` stdlib (mostly `phel.core`) before it runs your code (measured: cold ~1.23s vs warm ~0.17s). The PHAR already precompiles stdlib into a path-keyed `cache/compiled/`, but that cache is keyed by the build-time source path, so it is never consulted from an arbitrary project (the PHAR is mounted at a different path on every machine). Net effect: the precompiled stdlib does not help cold starts in real projects.

## 💡 Goal

Make the PHAR's precompiled stdlib actually accelerate cold starts in any project, approaching warm-cache speed on the very first run.

## 🔖 Changes

- **`BundledCompiledCache`** (read-only, content-addressed): looks up compiled PHP + per-namespace env keyed by **source content hash**, so it is install-location independent.
- **`CompiledCodeCache`**: consults the bundle as a fallback on a writable-cache miss for both `get()` and `getEnvironment()`. The writable project cache still wins when present.
- **`BundledPrecompiler::exportFromCache()`**: re-keys a fully-built cache's `phel.*` entries (primaries **and** `(load ...)` secondaries) into the content-addressed bundle.
- **`BundledStdlibPrecompiler`** + hidden **`_precompile-bundled <dir>`** command: evaluates all bundled namespaces, then exports the bundle.
- **`BuildConfig::getBundledPrecompiledDir()`**: resolves `<pkgRoot>/cache/precompiled` (present only in the PHAR; `null`/inert for source & composer checkouts; `PHEL_BUNDLED_PRECOMPILED_DIR` overrides).
- **`build/build-phar.php`**: generates and ships the bundle.
- Docs (`docs/internals/benchmarks.md`) + CHANGELOG.

## ✅ Verification

- New unit tests for the bundled cache, the `CompiledCodeCache` fallback, and the exporter.
- Locally: generated the bundle, then a cold `phel run` with it dropped from **1.23s → 0.18s** and wrote **0** `phel.core` files to the project cache; **all 5964 Phel core tests pass** against the bundled core (output parity verified).
- Inert for source/composer checkouts (no bundle shipped) — existing behaviour unchanged.

Note: the older path-keyed `preCompileStdlib()` bundling is left in place; de-duplicating it against this content-addressed bundle is a sensible follow-up.

Closes #2443

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 
+- Cold-start speedup: the PHAR now ships a read-only, content-addressed precompiled `phel.*` stdlib bundle (`cache/precompiled/`) that the compiled-code cache consults on a miss. A cold `phel run`/`phel test` in any project reuses the precompiled stdlib (incl. `phel.core`) instead of recompiling it, so first runs approach warm-cache speed. The bundle is keyed by source content hash, so it works regardless of where the PHAR is mounted; it stays inert (no behaviour change) for plain source/composer checkouts (#2443)
 - `phel doctor`: a "Checking performance" section now reports whether OPcache is configured to persist the compiled-code cache across CLI runs (`opcache.enable_cli`, `opcache.file_cache`) and prints actionable tips when it is not, helping warm `phel run`/`phel test` invocations approach native PHP speed (#2444)
 
 ### Fixed
--- a/build/build-phar.php
+++ b/build/build-phar.php
@@ -271,6 +271,41 @@ final class PharBuilder
     }
 
     /**
+     * Generate the read-only, content-addressed precompiled stdlib bundle
+     * under cache/precompiled/ so it ships inside the PHAR. At runtime the
+     * compiled-code cache consults it (keyed by source content hash), so a
+     * cold `phel run` in any project reuses the precompiled stdlib instead of
+     * recompiling it. Unlike preCompileStdlib()'s path-keyed cache, this one
+     * is install-location independent.
+     */
+    public function preCompileBundled(): void
+    {
+        $bundleDir = $this->root . '/cache/precompiled';
+
+        // Clean previous bundle to avoid stale, source-mismatched entries.
+        if (is_dir($bundleDir)) {
+            $files = glob($bundleDir . '/*') ?: [];
+            foreach ($files as $file) {
+                @unlink($file);
+            }
+        }
+
+        $exitCode = 0;
+        passthru(
+            \sprintf(
+                'cd %s && php bin/phel _precompile-bundled %s 2>&1',
+                escapeshellarg($this->root),
+                escapeshellarg($bundleDir),
+            ),
+            $exitCode,
+        );
+
+        if ($exitCode !== 0) {
+            throw new RuntimeException("Bundled stdlib precompile failed with exit code {$exitCode}");
+        }
+    }
+
+    /**
      * Build the PHAR archive
      */
     public function build(): void
@@ -278,6 +313,7 @@ final class PharBuilder
         $this->validate();
         $this->prepareReleaseConfig();
         $this->preCompileStdlib();
+        $this->preCompileBundled();
         $this->cleanup();
 
         try {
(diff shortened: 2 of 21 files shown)


Chemaclass:
Putting this back to **draft** — the content-addressed bundle has an architectural mismatch with secondary-file loading.

`phel.core` pulls in ~29 secondary files via the emitted `(load ...)` runtime code, which resolves by **sibling `.php` / classpath path lookup**, not via the content-addressed `CompiledCodeCache`. When core is served from `cache/precompiled/<hash>.php`, the sibling check (`__DIR__/core/meta.php`) fails and, depending on the command flow, the load classpath may not contain phel's internal src — so `phel eval '(expr)'` and `phel test` break in the PHAR with `Cannot locate core/meta for (load ...)`. `phel run` works; eval/test don't.

Also: the bundle pushed the PHAR over the 2 MiB smoke budget (even after stripping inline source maps), because the precompiled stdlib is inherently ~480 KB compressed.

**Recommended redesign:** ship a *path-structured* compiled bundle (src-tree layout, e.g. `<bundle>/phel/core.php`, `<bundle>/phel/core/meta.php`) and register that dir on `LoadClasspath` so the existing `(load ...)` sibling/classpath mechanism resolves secondaries to precompiled `.php` — the same model deployed builds already use. The content-addressed `CompiledCodeCache` fallback can still serve primary namespace files. This is a larger change than a single clean PR and deserves review before landing.

Verified locally: cold `phel run` with the bundle dropped 1.23s → 0.19s and all 5964 core tests passed, so the perf win is real once secondary loading is handled correctly.

Chemaclass:
Attempted the path-structured redesign and uncovered a **third, deeper blocker** beyond `(load)` resolution and PHAR size:

When `phel.test` is served from the bundle in the PHAR, its **macros don't resolve** during user-test compilation — `Cannot resolve symbol 'deftest'` / `phel.test/run-tests`. A bundle `require` populates the runtime `Registry` but does not fully reconstruct the **analyzer's** macro/namespace environment the way a fresh compile (or the baked caller-source-dir fallback in `LoadEmitter`) does. This works in the dev tree but breaks in the read-only PHAR with a different app root.

So a correct fix spans three subsystems — compiled-code cache, `(load ...)` resolution (`LoadEmitter` bakes the caller's build-time source dir), and analyzer namespace/macro env restoration — and can only be validated by repeated PHAR builds. That's a maintainer-level design effort, not a single autonomous PR.

**Recommendation:** keep this branch/PR as the documented spike, and treat #2443 as a design task. The measured win (cold `phel run` 1.23s → ~0.19s) is real and worth pursuing deliberately. Leaving #2443 open.

Chemaclass:
Closing this spike. The content-addressed approach is a dead end (see the three-subsystem blockers above); the path forward is the path-structured design written up in #2443. Branch `perf/2443-aot-precompile-core` kept for reference. #2443 stays open as a design task.

PR #2489, opened 2026-06-19
fix(lang): make NAN usable as a map/set key

### LINKED ISSUE #2475: NAN map keys are unretrievable and duplicate on re-assoc
## 🐛 Bug

`NAN` used as a map key is unretrievable and duplicates on re-`assoc`.

```phel
(def nan (/ 0.0 0.0))
(get {nan "value"} nan)            ; => nil   (expected "value")
(contains? {nan "value"} nan)      ; => false (expected true)
(count (assoc (assoc {} nan 1) nan 2)) ; => 2  (expected 1)
```

## 🔍 Root cause

`Lang/Equalizer::equals(nan, nan)` returns `false`: it fast-paths `===` (false for NaN per IEEE-754) and integral numbers, then falls through to `false` for floats. `Hasher` gives NaN a stable hash, so the key lands in a bucket but the equality check never matches — lookups miss and `assoc` can't find the existing entry to overwrite.

## ⚖️ Design decision needed (Clojure parity)

This is **not** a trivial fix, because of a deliberate asymmetry in Clojure:

- `(= ##NaN ##NaN)` → **false** (value `=` uses `==`)
- `(get {##NaN 1} ##NaN)` → **1** (HashMap key equality uses `Double.equals`, which special-cases `NaN.equals(NaN) == true`)

So the correct fix likely makes **collection key/equality** (the path behind `get`/`contains?`/`assoc` and `=` on collections) treat `NaN == NaN` as `true`, **without** changing scalar `(= ##NaN ##NaN)` to `true`. We should confirm the intended Phel stance against the clojure-test-suite buckets before changing `Equalizer`, since it underpins both paths.

## ✅ Acceptance

- `(get {##NaN v} ##NaN)` returns `v`; `(contains? {##NaN v} ##NaN)` is `true`; re-`assoc` with the same NaN key updates in place (count stable).
- Scalar `=` semantics for NaN are decided explicitly (match Clojure: `(= ##NaN ##NaN)` stays `false`) and covered by tests.
- Same behavior verified for vectors/sets as keys/elements where applicable.

Found via an adversarial runtime bug-hunt; all three symptoms reproduced.

Chemaclass:
## 🤔 Background

`NAN` used as a map key was unretrievable and duplicated on re-`assoc`: `Equalizer::equals(NaN, NaN)` is `false` (IEEE-754), so although `Hasher` gives NaN a stable hash and the key lands in a bucket, the key-equality check never matches. Lookups missed and `assoc` could not find the existing entry to overwrite.

## 💡 Goal

Make collection key equality treat NaN as equal to itself (Clojure/Java `Double.equals` semantics) so NaN works as a stable map/set key, **without** changing scalar `(= NAN NAN)`, which stays `false`.

## 🔖 Changes

- Add `EqualizerInterface::equalsKey` — `equals()` plus a NaN-equals-itself special case — and use it at every map-node **key** comparison site (`IndexedNode`, `HashCollisionNode`, `PersistentArrayMap`, `TransientArrayMap`). Sets are map-backed, so they are covered too.
- Value comparisons and scalar `=` keep using `equals()` unchanged.
- Tests: `EqualizerTest` (scalar `=` stays false, `equalsKey` matches NaN), and a `nan-map-keys.phel` core test for `get`/`contains?`/`assoc`/set behavior.

Scope: handles direct NaN scalar keys and NaN set elements (the reported symptoms). A vector/map that *contains* NaN used as a key is out of scope — matching Clojure there needs a separate deep key-equality traversal that must not change `(= [NAN] [NAN])` (which stays `false`, as in Clojure).

The mandatory refactor pass surfaced no further changes.

Closes #2475

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- `NAN` used as a map/set key is now retrievable: `get`/`contains?` find it and re-`assoc` updates in place, while scalar `(= NAN NAN)` stays `false` (#2475)
 - `phel\html`: void elements (`br`, `img`, `input`, ...) are now always self-closing and ignore supplied children, instead of emitting invalid markup like `<br>content</br>`
 - `phel\transit`: empty maps now round-trip as maps (were decoding as empty vectors), and numeric string keys (e.g. `{"1" "one"}`) stay strings instead of becoming ints
 - Reading a struct field with a non-keyword key (e.g. `(get s 'field)`) no longer crashes with a PHP `TypeError`; symbols and strings match by name, other key types return `nil`
--- a/src/php/Lang/Collections/HashSet/PersistentHashSet.php
+++ b/src/php/Lang/Collections/HashSet/PersistentHashSet.php
@@ -9,6 +9,9 @@ use Phel\Lang\Collections\Map\PersistentMapInterface;
 use Phel\Lang\HasherInterface;
 use Traversable;
 
+use function is_float;
+use function is_nan;
+
 /**
  * @template V
  *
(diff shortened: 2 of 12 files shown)


Chemaclass:
Pushed a follow-up: the first commit's `equalsKey` leaked into structural `=` (distinct maps/sets with a NaN key compared equal), failing `clojure-test-suite` `eq.cljc` 'collections unequal'. Guarded the `=` paths (`AbstractPersistentMap::equals`, `PersistentHashSet::equals`) so a NaN key/element keeps distinct collections unequal, while `get`/`contains?`/`assoc` still match NaN; identical collections short-circuit via `===`. Full clojure-test-suite green locally (5474). Refactor pass surfaced no further changes.

Issue #2626, opened 2026-06-27
Compiler: infer primitive :tag on let/loop bindings so hot code needn't fall back to php/ interop operators


Chemaclass:
## Summary

In hot numeric code, real Phel projects are forced to write raw `php/` interop operators (`php/+`, `php/-`, `php/*`, `php/===`, `php/<`, …) instead of the `phel.core` operators (`+`, `-`, `*`, `=`, `<`, …), purely for performance. This issue analyses **why**, shows that 0.46's existing native-lowering already solves most of it *in principle*, and pins the one structural gap that blocks it in practice — plus a small prerequisite correctness fix.

The driver is a real codebase ([phel-doom](https://github.com/Chemaclass/phel-doom), a terminal raycaster). It contains **2,724** `php/` arithmetic/comparison/equality op call sites — `php/+` (631), `php/-` (593), `php/*` (543), `php/===` (348), `php/<` (188), `php/>=` (147), … — almost all in per-ray / per-pixel / per-frame loops where the `phel.core` fn-call dispatch (variadic arg array → `(case (count xs) …)` → `NumericOperations::add/…`, with BigInt/Ratio/coercion branches) is too expensive.

## What 0.46 already does well

0.46's type-specialized emission already lowers `phel.core` ops to native PHP operators when it can statically prove the operands are primitive (great work — #2148, #2175, and the 0.46 perf series):

- `(+ a b)` / `(- a b)` / `(* a b)` / `(< a b)` / `(<= a b)` / `(> a b)` / `(>= a b)` → native `+ - * < <= > >=` when **both** operands are statically int/float (`NumericOperationSpecialization::typedBinaryOpName`, `NUMERIC_BINARY_OPS`).
- `(= a b)` → native `===` when both operands are int/float/bool/string, **and** the keyword-literal case `(= x :kw)` → `===` (`eitherArgIsKeywordLiteral`, `NumericOperationSpecialization.php:150-156,327-334`) — correctly restricted to `Keyword` (interned) and excluding `Symbol`.
- `inc`/`dec` on tagged ints, variadic chains, nested typed-arith, and `php/->`/`php/::`/`php/new`/`php/oset` compiled to native expressions.

[...]

PR #2627, opened 2026-06-27
feat(compiler): infer primitive :tag on let/loop bindings (less php/ interop in hot code)

Chemaclass:
Implements the primary proposal in #2626 (follow-ups stay tracked there).

## What

Native lowering of `(+ a b)` / `(< a b)` / `(= a b)` only fires when both operands are statically-typed primitives. Until now `let` / `loop` bindings never carried a type tag, so morally-primitive bindings fell back to slow runtime dispatch — pushing hot code to raw `php/` interop.

This infers a primitive `:tag` (int/float/bool/string) on `let` / `loop` bindings from their init, so ops over them lower to native PHP operators automatically.

```clojure
(fn [v] (let [n (php/count v)] (+ n 1)))
```
```php
// before: (\Phel::getDefinition("phel.core", "+"))->__invoke($n_1, 1)
// after:
/** @var int $n_1 */
$n_1 = count($v);
return ($n_1 + 1);
```

## How

`BindingTypeInferrer` derives the tag from the binding's init (literals, typed locals, pure `php/` fns, core `+ - *` / `inc` / `dec`) and grafts it onto the binding — identical to a hand-written `^int`, so it flows through the existing emitter guards unchanged.

- `let` bindings type from their init; `loop` bindings keep the tag only when the init and every `recur` arm agree on one primitive.
- The tag binds to the binding's unique **shadow** instance, so a name reused in a nested scope resolves its own type (not an outer binding's).
- User-written tags always win. Anything not provably primitive stays on dispatch (a missing tag is always safe; a wrong one is never emitted).

Also fixes `KnownPhpFunctionReturnTypes` to return `float` for `floor`/`ceil`/`round` (PHP returns float), and shares the int/float promotion rule via `ArithmeticResultType`.

## Tests

`BindingTypeInferrerTest` (graft contract, user-tag-wins, floor→float, bail cases) plus integration fixtures for the inferred, chained, shadowed, and `recur`-disagreement paths. Full suite + static analysis green.


Chemaclass:
Thanks for the review! Addressed both points:

**1. `array_all()` / PHP 8.4** — `array_all` (with `array_any`/`array_find`/`array_find_key`) was added in **PHP 8.4**, not 8.5, so it's within the project's `>=8.4` floor. It's already used across the shipped codebase (e.g. `NumericOperationSpecialization`, `AbstractPersistentVector`, `LiteralCollectionFolder`, `CallSpecialization`), and the PHP 8.4 CI job is green. No change needed here.

**2. `slice.test` == `simple-let.test`** — confirmed this is **pre-existing on `main`**: `git show main:tests/php/Integration/Fixtures/Let/slice.test` already holds `(let [x 1 y 2] (php/+ x y))`, identical to `simple-let.test`. This PR only refreshed its expected output for the new `/** @var int */` doctags; it didn't introduce the duplication. Happy to dedup/rename it in a separate cleanup PR so this one stays focused.

Separately, a self-review pass caught and fixed a **correctness bug** this PR would otherwise have shipped: a `let`/`loop` binding whose name shadows an outer binding resolved its `:tag` to the *outer* binding (first-occurrence-wins by-name lookup), so e.g. `(let [n (php/count v)] (let [n (php/strtolower "x")] (+ n 1)))` emitted `("x" + 1)` (runtime TypeError). Fixed by binding the tag to the unique shadow instance and resolving from it; regression fixtures added (`let-shadowed-inner-*`). Also extracted `ArithmeticResultType` as a shared promotion rule, injected `BindingTypeInferrer`, and added `BindingTypeInferrerTest`.

PR #2656, opened 2026-06-28
docs(init): add composer install to next-steps output

MHJanny:
## 🤔 Background

phel init prints steps referencing ./vendor/bin/phel, but users haven't run composer install yet.

## 💡 Goal

Add composer install as step 1 so new users don't hit "command not found."

## 🔖 Changes

- Added "Install dependencies: composer install" as step 1 in printNextSteps()


Chemaclass (approved):
Thanks!

PR #2661, opened 2026-06-29
fix(core): validate defn/defmacro name is a symbol with a clean error

### LINKED ISSUE #2639: fix: malformed defn/macro args crash with internal PHP error instead of a clean message
## Background
`(defn 123 [x] x)` (non-symbol name) surfaces as a PHP-flavored internal crash:
```
Call to a member function getMeta() on int
  at src/phel/core/defs.phel:33 (compiled: .../phel.core__<hash>.php:61)
#0 ... (phel\core\defn-builder ...)
   ... 25 internal frames
```
exposing core internals + a compiled-cache path. Compare `(let [a 1 b] ...)` and `(if true 1 2 3)` which validate and give clean `[PHEL...]` messages.

## Fix sketch
Guard the top of `defn`/`defn-builder` (`src/phel/core/defs.phel`) to assert the name is a symbol and throw a readable Phel-level error, mirroring `let`/`if`. Broader principle: form-destructuring macros should validate shape first; `defn` is the highest-value one.

**Effort:** M. **User-facing:** yes.

Chemaclass:
## 🤔 Background

`(defn 123 [x] x)` (a non-symbol name) reached `(getMeta)` inside `defn-builder` and crashed with an internal PHP `Error` (`Call to a member function getMeta() on int`). Because that is an `Error`, not an `Exception`, it escaped the macro expander's `catch (Exception)` and surfaced raw, leaking core internals, a compiled-cache path, and a wall of internal frames. Compare `(let [a 1 b] ...)` and `(if true 1 2 3)`, which validate shape and give clean `[PHEL...]` messages.

## 💡 Goal

Make malformed `defn`/`defmacro` names fail with a clean, located `[PHEL005]` macro-expansion error naming the offending type, mirroring `let`/`if`.

## 🔖 Changes

- Guard the name at the top of `defn-builder` (`src/phel/core/defs.phel`): when it is not a `Symbol`, throw an `InvalidArgumentException`, which the macro expander wraps into a clean located `[PHEL005]` error.
- `defn`, `defmacro`, `defn-`, `defmacro-` all route through `defn-builder`, so all are covered.
- Added `DefnMalformedNameRuntimeTest` integration test (clean error for `defn`/`defmacro`, valid `defn` still compiles).
- Updated `CHANGELOG.md` Unreleased.

Closes #2639


Chemaclass:
Superseded by #2659 (already merged) — identical defn-builder guard, with a more thorough test suite there. Closing as duplicate.

Issue #2714, opened 2026-07-08
perf: specialize untyped reduce/map/filter/loop over seqs


Chemaclass:
## Summary

Specialize `reduce` / `map` / `filter` / `loop` over seqs when the collection and/or step-fn are known, instead of falling through generic dispatch per element.

## Problem

No specialized emitter exists for these (unlike `get`/`count`/`first`/`nth`/`assoc`, which do). On untyped input, each iteration:

- walks a 6–10 branch `phel.core` cond chain to decide the seq operation, and
- re-dispatches the step function through the registry / `AbstractFn` per element.

For hot loops this compounds — the per-element dispatch is the dominant cost.

## Proposed approach

Add specialized emitters (`NodeEmitter/Specialized/*`) that, when the collection type is known (PHP array, `PersistentVector`, typed seq):

- emit a native `foreach` loop, and
- **hoist the resolved step-fn out of the loop** (resolve once, invoke directly per element — monomorphize).

Fall back to the generic core fn when the shape is unknown.

Pairs naturally with idea #1 (type inference) — the more types are inferred, the more often this fires.

## Expected payoff

Removes a per-element dispatch multiplier in the most common iteration constructs.

## Validation

Benchmark tight `reduce`/`map` over a 10k-element vector; `PhelBench` for regressions.

---
_Part of a compiler→PHP runtime-perf sweep._

JesusValeraDev:
Implemented the `reduce` half in #2733. The rest of this issue needs re-speccing — the premise doesn't hold for `map`/`filter`:

**`map` and `filter` cannot be lowered to an eager `foreach`.** Both build a `LazySeq` over a PHP `Generator` (`Seq::map` / `Seq::filter` in `src/php/Lang/Seq.php`) and `copy-meta` onto the result. An eager loop would change the return type, break infinite/expensive seqs, drop the meta copy, and shift side-effect timing — a semantics break, not an optimization.

They also have little to gain: the issue says each iteration "re-dispatches the step function through the registry", but `f` is handed to the generator **once** and invoked directly per element. There is no per-element registry lookup to remove.

**`loop`** is a special form (`RecurEmitter`), not a call site, so it isn't reachable from the call-site specialization machinery — and #2731 already dropped the IIFE wrapper for return-context foreach.

[...]

PR #2733, opened 2026-07-11
perf(emitter): specialize reduce over typed vectors

JesusValeraDev:
## 🤔 Background

#2714 proposes specializing `reduce` / `map` / `filter` / `loop` over seqs into native `foreach` loops with the step function hoisted out of the loop.

Digging into the runtime, only **`reduce`** can actually be lowered this way:

- **`map` and `filter` are lazy.** Both build a `LazySeq` over a PHP `Generator` (`Seq::map` / `Seq::filter`) and `copy-meta` onto the result. An eager `foreach` lowering would not be an optimization but a semantics break: it changes the return type, breaks infinite/expensive seqs, drops the meta copy, and shifts side-effect timing. They also have little to gain — `f` is handed to the generator **once**, so there is no per-element registry dispatch to remove (contrary to the issue's premise).
- **`loop`** is a special form (`RecurEmitter`), not a call site, and #2731 already removed the IIFE wrapper for return-context foreach.
- **`reduce` (3-arity) is eager** and genuinely wasteful per element — this PR takes that.

So this PR delivers the `reduce` half. `map`/`filter` are deliberately left alone; I'd suggest re-speccing or closing the rest of #2714 in light of the laziness constraint.

## 💡 Goal

Lower `(reduce f init v)` — where `v` is a local tagged as a `PersistentVectorInterface` — to a native `foreach`, removing the per-element overhead the runtime `phel.core/reduce` pays.

The runtime body boxes the accumulator and an early-exit flag in two `Volatile` objects and walks the collection with a generic `dofor`, paying a `deref`/`reset` pair plus a `reduced?` dispatch on **every element**. The lowering removes all of it: the step fn is resolved once and hoisted out of the loop, and the accumulator lives in a plain PHP variable.

Measured **~13% faster** reducing a 10k-element vector (26.1ms → 22.8ms, best-of-40 after warmup).

Semantics are unchanged. In particular the `(reduced x)` early-termination contract is preserved: a `Reduced` result unwraps into the accumulator and breaks the loop.

## 🔖 Changes

- **`ReduceSpecialization`** — call-site eligibility: `phel.core/reduce`, 3 args, collection tagged `PersistentVectorInterface`. Registered in `CallSpecialization::isSpecialized()` so the call-slot scanner does not reserve an orphan cache slot. The 2-arity `(reduce f coll)` is excluded — it seeds from the collection and calls `(f)` on empty input, which the foreach lowering does not model.
- **`ReduceCallEmitter`** — emits the `foreach` with the hoisted step fn, the plain accumulator, and the `Reduced` early-exit guard. `PersistentVectorInterface` is an `IteratorAggregate`, so the `Seq::toIterable` adapter is not needed.
- **By-ref capture** — the lowering hosts its loop in an IIFE, so locals reached by `(php/ref x)` in an operand are captured by reference via `ByRefLocalCollector`, matching every other IIFE-wrapping emitter. Without it a by-ref write lands in the closure's copy and is silently lost; there's a regression test for it.
- **`INDEX_FORMAT_VERSION` 1.3 → 1.4** — the specialization changes emitted PHP for unchanged `.phel` source, and the compiled-code cache keys only on the source hash, so a same-version cache would keep serving pre-change output (the rule documented on the constant, same as #2732).
- **Tests** — unit tests for eligibility (arity, tag, wrong fn, seq/map tags all fall back), an integration fixture locking the emitted PHP, and Phel-level tests covering the runtime contract: `reduced` early exit, left-associativity, empty input, once-only evaluation of the hoisted fn/seed, parity with the runtime `reduce`, and the by-ref case.
- Docs: `Compiler/CLAUDE.md` records the new family and the laziness gotcha, so a future pass doesn't re-attempt the unsafe `map`/`filter` lowering.

Related to #2714


Chemaclass (approved):
thanks!

JesusValeraDev:
Verified all four points in the review against the branch — the assessment holds:

1. **IIFE overhead O(1), saving O(n)** — confirmed. `CallEmitter` emits the context prefix before dispatch, so the lowering must produce an expression and the loop lives in an IIFE (`ReduceCallEmitter.php:44-48`). One closure per call against removing the per-element `Volatile` deref/reset pair plus `reduced?` dispatch. Sound for any non-trivial vector.

2. **ByRefLocalCollector applied** — confirmed and grounded in a real fix (commit `8d23b04`). `ReduceCallEmitter.php:56` passes `new ByRefLocalCollector()->collect($node)` to `emitFnWrapPrefix`, matching every other IIFE-wrapping emitter (`LetEmitter`, `DoEmitter`, `ThrowEmitter`, `ForeachEmitter`, `TryEmitter`, `PhpObjectCallEmitter`). Without it a `(php/ref x)` operand write lands in the closure's copy and is silently lost.

3. **INDEX_FORMAT_VERSION bump** — confirmed necessary. `1.3` → `1.4`; the lowering changes generated PHP, so a same-source cache would otherwise serve pre-change bytecode.

4. **Excluding 2-arity reduce and map/filter** — right call, verified against the runtime `phel.core/reduce` (`transducers.phel:50-71`). The specialization gates on exactly three args (`ReduceSpecialization.php:57`). The 2-arity `(reduce f coll)` seeds the accumulator from the collection and calls `(f)` with no args on empty input — neither modeled by the foreach lowering. `map`/`filter` return a `LazySeq`; an eager foreach would change the return type, break infinite/expensive seqs, and shift side-effect timing (documented in `Compiler/CLAUDE.md`).

The `reduced` early-termination contract is preserved (`ReduceCallEmitter.php:115-130`: `instanceof Reduced` → `deref()` + `break`). Conservative and correct — no changes needed.

PR #2740, opened 2026-07-15
feat(core): add qualified-symbol?, qualified-keyword?, qualified-ident? predicates

Chemaclass:
## 🤔 Background

Phel already ships the `simple-symbol?` / `simple-keyword?` / `simple-ident?` predicates (true for **un-namespaced** idents) plus the umbrella `ident?`. Their Clojure counterparts for the **namespaced** case — `qualified-symbol?`, `qualified-keyword?`, `qualified-ident?` — were missing, leaving the family asymmetric.

## 💡 Goal

Complete the ident-predicate family so both halves (simple/qualified) exist, matching Clojure.

## 🔖 Changes

- `src/phel/core/predicates.phel`, placed right after `simple-ident?`:
  - `qualified-symbol?` — `(and (symbol? x) (some? (php/-> x (getNamespace))))`
  - `qualified-keyword?` — same for keywords
  - `qualified-ident?` — `(or (qualified-symbol? x) (qualified-keyword? x))`
  - Mirrors the exact shape of the existing `simple-*` predicates (which use `(nil? (getNamespace))`); the qualified variants invert that with `some?`.
- Tests in `tests/phel/core/type-operation.phel` mirroring the `simple-*` test blocks: namespaced true; simple/cross-type/`nil` false.
- `## Unreleased` CHANGELOG entry.

Full core suite green locally: 6323 passed, 0 failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ All notable changes to this project will be documented in this file.
 
 - `every-pred`: complement of `some-fn`; returns a predicate that is true only when every composing predicate is logical true for every argument, short-circuiting on the first false and returning a strict boolean (#2738)
 - `mapv` and `filterv`: eager vector-returning variants of the lazy `map` and `filter` (#2739)
-- `qualified-symbol?`, `qualified-keyword?`, and `qualified-ident?`: predicates for namespaced symbols/keywords, completing the ident-predicate family alongside the existing `simple-*` variants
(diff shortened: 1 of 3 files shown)


Chemaclass:
Closing: the parity suite (clojure-test-suite) requires `(qualified-keyword? (keyword "a/b/c"))` => true, but Phel's `keyword` string constructor doesn't parse a namespace (unlike `symbol`), so these predicates can't pass without a foundational fix to keyword ns-parsing + the intern key. Tracked in #2741. The predicates are correct against Phel's current keyword model but would turn the suite red, so they wait on #2741.

PR #2762, opened 2026-07-17
fix(lang): parse namespace in keyword string constructor

### LINKED ISSUE #2741: keyword string constructor doesn't parse namespace (inconsistent with symbol); intern-key collision
## Summary

`(keyword s)` does not parse a `ns/name` namespace out of its string argument, whereas `(symbol s)` does. This diverges from Clojure and from Phel's own `symbol`, and it blocks a faithful `qualified-keyword?` / `qualified-ident?` (the `clojure-test-suite` expects `(qualified-keyword? (keyword "a/b/c"))` => `true`).

## Repro

```phel
(namespace (symbol "foo/bar"))   ; => "foo"   (Symbol::create splits on first '/')
(namespace (keyword "foo/bar"))  ; => nil     (Keyword::create stores the whole string as the name)

(namespace (symbol "a/b/c"))     ; => "a"      name "b/c"
(namespace (keyword "a/b/c"))    ; => nil      name "a/b/c"
```

Clojure: `(namespace (keyword "a/b/c"))` => `"a"`, `(name (keyword "a/b/c"))` => `"b/c"` (split on the **first** `/`).

Root: `Symbol::create()` (`src/php/Lang/Symbol.php`) does `strpos($name, '/')` and splits; `Keyword::create()` (`src/php/Lang/Keyword.php`) does not — it takes `(name, ?namespace)` and stores `name` verbatim.

## Secondary: intern-key collision

`Keyword::create` interns under `$namespace !== null ? "$namespace/$name" : $name`. So `(keyword nil "a/b")` (name `a/b`, ns nil) and `(keyword "a" "b")` (name `b`, ns `a`) both key to `"a/b"` and collide in the intern pool — the second call returns whichever was interned first, regardless of the requested ns. (Observable today: a `:foo/bar` literal read earlier makes a later `(keyword "foo/bar")` inherit ns `foo` via the shared pool.)

## Note on the reader

Two-slash keyword **literals** also don't parse: `(namespace :a/b/c)` => `nil` in Phel vs `"a"` in Clojure. Any fix should align the fn, the reader, and the intern key so all three agree.

## Scope

Foundational (touches `Keyword`, the reader, and the intern key, with knock-on effects on equality/interning/serialization per `src/php/Lang/CLAUDE.md`), so it deserves its own focused change rather than riding along with the predicate addition. Filed while attempting `qualified-symbol?`/`qualified-keyword?`/`qualified-ident?` (PR #2740), which was closed because it can't pass the parity suite without this.

JesusValeraDev:
## 🤔 Background

`(keyword "foo/bar")` stored the whole string as the keyword's name, while `(symbol "foo/bar")` splits on the first `/`. This diverges from Clojure and blocked a faithful `qualified-keyword?` (#2740). The intern pool also keyed entries by the ambiguous `"$namespace/$name"` string, so `(ns nil, name "a/b")` and `(ns "a", name "b")` collided and returned whichever instance was interned first.

## 💡 Goal

Align `keyword` with `symbol` and Clojure: split a string argument on the first `/` into namespace and name, and make interning collision-free. The reader already parses `:a/b/c` correctly, so this brings the fn and the intern key in line with it.

## 🔖 Changes

- `Keyword::create(string)` splits the name on the first `/` (mirrors `Symbol::create`, bare `/` stays verbatim); an explicit namespace keeps the name verbatim.
- New `Keyword::createForNamespace(?ns, name)` verbatim constructor (mirrors `Symbol::createForNamespace`); used by the two-arg `keyword` fn, symbol input, and the literal emitter so unqualified slashed names round-trip.
- Intern pool keyed by namespace and name separately (nil ns distinct from empty-string ns).
- Final review pass: deduplicated `emitKeyword` emission and refreshed the `Keyword` row in `src/php/Lang/CLAUDE.md`.

Closes #2741


Chemaclass (approved):
awesome

PR #2764, opened 2026-07-18
fix(run): require active xdebug coverage mode when detecting coverage driver

JesusValeraDev:
## 🤔 Background

`CoverageDriver::detect()` picked xdebug as the line-coverage driver whenever the extension was loaded. But `xdebug_start_code_coverage()` only collects data when `coverage` is among the active `xdebug.mode` values — with any other mode (e.g. the common `develop` default), `phel test --coverage` emitted PHP warnings, reported "no project source files were executed", and produced empty reports. The four `TestCommandCoverageTest` integration tests failed on such machines instead of skipping.

## 💡 Goal

Treat xdebug without the `coverage` mode as "no driver available" so `phel test --coverage` fails fast with an actionable hint, and the integration tests skip gracefully on misconfigured environments.

## 🔖 Changes

- `CoverageDriver::detect()` also requires `coverage` to be an active xdebug mode (queried via `xdebug_info('mode')`, honoring the `XDEBUG_MODE` env override; xdebug builds too old to expose `xdebug_info()` are assumed usable).
- New `CoverageDriver::unavailabilityReason()` tells the user what is missing; `phel test --coverage` now prints `xdebug is loaded but 'coverage' is not an active mode; re-run with XDEBUG_MODE=coverage or set xdebug.mode=coverage in php.ini.` instead of the generic "neither is loaded" text.
- Unit tests cover both detection branches; the existing `TestCommandCoverageTest` end-to-end suite now skips (instead of failing) when the subprocess has xdebug in a non-coverage mode, and still fully passes under `XDEBUG_MODE=coverage`.
- CHANGELOG entry under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,10 @@ All notable changes to this project will be documented in this file.
 
 - Multi-arity function calls skip the variadic `__invoke` dispatch. Multi-arity fn classes now emit fixed-arity `invokeArityN` methods, and build-mode call sites with a statically known arity route through them via a guarded shortcut, avoiding the per-call `...$args` packing, `$args[N]` re-indexing, and closure indirection. Roughly 1.5-2x faster per multi-arity call in microbenchmarks (larger under JIT); single-arity calls are unchanged (#2760)
 
+### Fixed
+
+- `phel test --coverage`: xdebug is only used as the coverage driver when `coverage` is among its active modes; otherwise the command fails fast with a hint to re-run with `XDEBUG_MODE=coverage` instead of emitting PHP warnings and an empty report (#2764)
+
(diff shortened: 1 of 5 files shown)


JesusValeraDev:
CI failure was a pre-existing parallel-test race, unrelated to the coverage change: `test_run_by_filename_outside_config` wrote (and deleted) `outside-script.phel` inside the scanned test tree, so a concurrent worker's namespace scan (`EvalCommandTest` stdin test via `loadPhelNamespaces`) could pick the file up and then fail reading it after deletion. Fixed by creating the script in the system temp dir, which also matches the test's outside-the-configured-dirs intent.

Chemaclass (approved):

Issue #2787, opened 2026-07-20
phel list:modules / cache:warm fatal in the phel-lang repo: Gacela module discovery walks tests/


Chemaclass:
## Problem

`./bin/phel list:modules` (and the Gacela-backed `debug:modules` / `cache:warm` paths) fatal when run inside the phel-lang repo. Gacela's module discovery `class_exists()`-walks the whole app root and dies loading `tests/.../ReplTestCommand.php`.

Pre-existing behavior, reproduced on `main`; NOT a Gacela 1.18 regression (verified during the 1.18 upgrade in #2786).

## Options

1. Scope discovery in `Phel::configFn()` via `GacelaConfig::setAppModulePaths()` (e.g. `src/php` for this repo). Open question: the right default for end-user Phel projects, which typically have no Gacela modules at all.
2. Upstream hardening in gacela-project/gacela: skip `tests/` / `Tests/` directories during discovery (1.17 already anchors test-module *name filtering* on `\Test\`/`\Tests\` namespace segments; discovery walking is the remaining gap).

## Acceptance criteria

- [ ] `./bin/phel list:modules` and `./bin/phel cache:warm` exit cleanly in this repo
- [ ] No behavior change for end-user Phel projects
- [ ] If fixed upstream instead: constraint bump + verification, this issue links the upstream PR

PR #2795, opened 2026-07-20
test(repl): resurrect the dormant REPL integration fixture suite

JesusValeraDev:
## 🤔 Background

`tests/php/Integration/Run/Command/Repl/Fixtures/` and `FixturesWithCoreLib/` hold 23 `--INPUT--`/`--EXPECT--` REPL fixtures. Their runner was named `ReplTestCommand.php`, and PHPUnit only collects files matching the `*Test.php` suffix, so the whole suite had been **silently uncollected** for a long time. It also could not be loaded at all under PHPUnit 13 (it overrode the now-`final` `TestCase::__construct()`), which is what made Gacela's module discovery fatal in #2787.

Resurrecting the suite turned out to be worth it: it surfaced **one real product bug and three real harness bugs**. The two product/harness fixes with the widest blast radius were split out and already merged:

- #2794 — REPL hung on a mistyped closing bracket (real product bug)
- #2793 — `ReplTestIo::writeln()` never terminated lines

This PR is the remainder: wire the suite back up, plus the fixture/harness corrections that don't belong in those two.

## 🔖 Changes

**Collect the suite**
- Rename `ReplTestCommand.php` → `ReplCommandTest.php`; drop the illegal `__construct` override; bootstrap Gacela's app root from the fixture directory (so `require`-from-cwd works) while pointing `src-dirs` at the real `src/phel` (so the core-lib fixtures resolve); process isolation per fixture.

**Harness corrections (test-double diverged from production)**
- **Printer**: the double used `Printer::nonReadable()`, but production switched to `readableWithColor()` in #315fae6e7 (May 2015) so strings are quoted. Aligned to `Printer::readable()` (color is stubbed off separately). `__FILE__` now yields `""`, matching `./bin/phel repl`.
- **stdout capture**: `println`/`doc` print via hard-coded `php/print` to real stdout, which the double dropped. It now owns a dedicated output-buffer level and drains it into its transcript on each write, so printed text interleaves with results exactly as the real REPL does (`ReplOutputTrait` echoes results to the same stdout). Concatenating captured stdout after the fact does **not** reproduce the interleaving, hence the drain.

**Fixtures refreshed for documented behaviour the dormant suite never saw**
- dot namespace separator (`phel.html`), `def`/`defn` returning a var handle, `[PHEL001]` error prefixes, `<function:name>`, and the blank line Ctrl-D prints on leaving multiline mode (production since 2b60de3b3).
- `use-namespace.test` → `use-php-class.test`: `use` is PHP-class-only by design (docs + `UseAliasRegistrar`), so the old "`use` imports a namespace" expectation was obsolete; repurposed to cover what `use` actually does. Namespace import via `require` is already covered by `require-namespace.test`.

**Sibling tests that captured their own stdout** now read from the IO (`ReplDefaultTapTest`, `ReplPromptNamespaceTest`). `ReplDefaultTapTest` matches the `tap> <value>` line by prefix, since the unified stream means a bare `tap> ` substring would also hit the echoed input `user:N> (tap> ...)`.

## ✅ Verification

Every fixture expectation was checked against `./bin/phel repl`, not copied from harness output. All 23 fixtures now run and pass; full REPL suite green (56 tests); `composer test` green end to end (6511, static analysis clean).

## 📝 Notes

- No production code and no `CHANGELOG.md` entry: the product-visible fixes shipped in #2793/#2794; this is test infrastructure.
- Two fixtures (`require-current-dir`, `doc-deprecated`) also carried stale result values (`(ns …)` → `nil`, `defn` → var handle) on top of the stdout issue; both corrected.


Chemaclass (approved):
Nicee

Issue #2859, opened 2026-07-26
Road to 1.0: what has to be true before the first stable release


Chemaclass:
Phel is at `v0.50.0` (`src/php/Shared/VersionFinder.php`). As of 2026-08-17 the tracker has one open issue, this roadmap, and no open PRs. The remaining 1.0 work is no longer about discovering missing language features. It is about finishing the public stability story and ecosystem handoff.

## What 1.0 promises

Two promises, and everything below exists to make one of them true:

1. **Language stability.** Phel source that compiles on `1.0.0` compiles on every `1.x`. Reader syntax, special forms and the public `phel.*` core API do not break inside the major.
2. **Embedding stability.** The PHP surface (`\Phel`, `*Facade`, `*FacadeInterface`, `Phel\Lang\*`, `Phel\Shared\*`, `Phel\Config\*`) follows semver, so a project wiring Phel into its own tooling can take `1.x` updates without reading a diff.

Anything else is written down as unsupported or out of scope instead of being left ambiguous.

## Where we stand

| | |
|---|---|
| Source | 1160 PHP files, 64 Phel files under `src/` |
| Tests | 623 PHPUnit test classes, 239 Phel test files, plus the nightly `clojure-test-suite` run |
| Static analysis | PHPStan level 9 over all of `src/`, Psalm level 1 |
| Tooling | REPL, nREPL, LSP, `format`, `lint`, `doc`, `build` (AOT), PHAR, `watch`, `profile`, `doctor`, `analyze`, `mutate` |
| Editors | VS Code extension, IntelliJ plugin |
| Language | macros, protocols, `defmulti`/`defmethod`, transducers, transients, persistent collections, atoms, futures and fibers, BigInt/BigDecimal, structs, schema |

The language is not what is missing. The commitments are mostly written and gated now.

---

## Current 1.0 readiness checklist

- [x] Define the public PHP embedding surface. Done in #2866: `docs/stability.md`, `PublicApiSurfaceTest`, and `tests/php/Unit/Architecture/public-api.snapshot.txt` define and gate it.
- [x] Mark internals. Done in #2866: internal PHP classes carry `@internal`, and `InternalAnnotationTest` pins the split both ways.
- [x] State the 1.x deprecation policy. Done in #2866: one compiler channel, one minor of notice before removal, removals only at a major.
[...]

PR #2866, opened 2026-07-26
feat: define and freeze the 1.0 surfaces, and gate them

Chemaclass:
## 🤔 Background

Sections 2 to 6 of #2859. It started as four stacked PRs (#2867, #2868, #2869); they are folded in here because the gates each one adds only hold given fixes from the others, so reviewing them separately left every branch but the last red.

1.0 makes two promises: Phel source that compiles on `1.0.0` keeps compiling on every `1.x`, and the PHP embedding API follows semver. Neither was written down, and `## Unreleased` alone carries four **BREAKING** PHP-API bullets. That cadence is fine at 0.x and impossible at 1.x.

The recurring theme: a surface that exists only in prose is not frozen, and a benchmark nobody compares cannot fail. Every claim here is paired with something that fails a build.

## 💡 Goal

Define both surfaces, prove the platform claims, and turn the three missing quality gates on.

### Section 2: the PHP embedding API

**[docs/stability.md](https://github.com/phel-lang/phel-lang/blob/feat/2859-freeze-embedding-api/docs/stability.md)** — six rules define the public set: `\Phel`, `Phel\<Module>\<Module>Facade(Interface)`, and all of `Phel\Shared\`, `Phel\Lang\`, `Phel\Config\`. Whole namespaces rather than hand-picked lists, because those are the ones a consumer cannot avoid: values crossing a facade boundary, the contracts facades speak in, and the object `phel-config.php` constructs. Plus which changes are breaking, the 1.x deprecation policy, the PHP and platform support policies, and what is explicitly *not* covered (emitted PHP text, diagnostic wording, cache format).

**The gate** — `PublicApiSurfaceTest` renders the surface to a committed snapshot (266 symbols) and fails on drift; `composer api-surface:update` regenerates it, and that diff *is* the BC review.

I did not use `roave/backward-compatibility-check` as the issue suggested: it compares against the last release tag, so a break surfaces after it ships. This gates the PR that introduces it and needs no new dependency. The one thing roave would catch that this does not (a vendor base class changing an inherited signature) is named in the doc.

**`@internal` on 792 classes**, so the split reaches IDEs and static analysis. `InternalAnnotationTest` pins both directions. `Phel\Phel` was already `@internal use \Phel instead`, so it stays internal and the members it declares (`bootstrap()`, `run()`, `configFn()`) fold into the `\Phel` entry, which is where a consumer reaches them.

### Section 3: the language surface

**[docs/spec/language-surface.md](https://github.com/phel-lang/phel-lang/blob/feat/2859-freeze-embedding-api/docs/spec/language-surface.md)** — frozen reader syntax, the closed 40-entry special-form list, the standard-library contract, and the stated non-goals (no STM/refs, no agents, no `core.async`, no character type) moved out of `internals/faq.md` where a user would never find them. `LanguageSurfaceSpecTest` parses the special-form table out of the markdown and compares it to the analyzer's dispatch registry, so adding a form fails the build until the spec is updated too.

**[docs/spec/clojure-divergences.md](https://github.com/phel-lang/phel-lang/blob/feat/2859-freeze-embedding-api/docs/spec/clojure-divergences.md)** — 66 files in `clojure-test-suite` carry a `:phel` branch explaining a divergence and none of it was visible to users. Catalogued by cause, including the one place Phel is *stricter*. `case` returning `nil` with no match is listed as a known gap rather than a decision, because the suite's own comment calls it "arguably a real semantic gap".

**`CoreApiSurfaceTest`** pins all 956 public `phel.*` definitions with their arities. A lost arity breaks source compatibility as surely as a lost definition, and that is the kind nobody notices.

### Section 4: the platform claims

macOS joins the supported tier (compiler, core and PHAR suites, blocking). The PHAR is the artifact people download and was built and smoke-tested on Linux only; its size check used GNU `stat -c%s`, replaced with POSIX `wc -c`.

Windows runs a reduced suite as a declared best-effort tier. `continue-on-error` sits on the *steps*, not the job: on the job it leaves a red check on every PR and trains people to ignore it. The job stays green and reports the outcome to the summary. It currently reports 56 failures, nearly all tests asserting `ends with "/src.phel"` against a Windows path — test-side assumptions, which is exactly what the tier is for.

README gained a requirements section; there was none.

### Section 5: the quality gates

- **Coverage** is measured under pcov, published to the run summary by `tools/check-coverage.php`, and gated on a floor.
- **Benchmarks** now measure the base revision first under `--tag=baseline`, then assert the PR against it using the tolerance already sitting unused in `phpbench.json`. Both runs on the same runner; benchmark definitions stay at the PR's version so what gets compared is this change, not dependency drift.
- **Mutation testing** over `Lang/` and the analyzer, weekly. Only the unit suite runs against each mutant, deliberately: `Cons::walkTail()` passes the whole unit suite with its body disabled and only `test-core` notices, so a mutant surviving the unit suite is the gap worth seeing.

### Section 6: the docs

**[Upgrade guide 0.49 to 1.0](https://github.com/phel-lang/phel-lang/blob/feat/2859-freeze-embedding-api/docs/migration/upgrade-0.49-to-1.0.md)** — six steps, and step 1 is "find out whether you are affected", because a clean `--warn-deprecations` run means the upgrade is a version bump. It leads on `,`, the only removal that does not fail loudly: `` `(f ,x) `` still parses and quietly *quotes*. Every claim is verified against the shipped binary, not the changelog.

**A documentation gate** — `:doc` required on every public definition, `:example` pinned as a ratchet at 128.

## 🐛 Bugs the new gates found

- **`defmulti` named its parameter list with `(gensym)`**, so `phel doc assert-expr` rendered `(assert-expr & __phel_3167)` with a different number every run. Now `(assert-expr & args)`.
- **`assert-expr`'s documentation was in a `;;` comment**, not a docstring, so `phel doc` showed nothing.
- **`defmulti` generated two undocumented public definitions** (`<name>-methods`, `<name>-prefers`). They must stay public for cross-namespace `defmethod`, and now carry docstrings.
- **PHP 8.5 resolves a `self` return type to the concrete class name** where 8.4 reports `self`. Caught by the 8.5 job on the first push; the renderer normalises both to the resolved name.
- **Three core tests assert concurrency by wall clock** and fail on the macOS runner, where a slow scheduler tick is indistinguishable from work that did not run concurrently. Tagged `^:timing-sensitive` and excluded on macOS only. Linux still runs them, so a genuine loss of concurrency is still caught. I deliberately did not loosen the thresholds, because that would hide the real thing if it is real.

## ⚠️ Two ratchets need one calibration pass

`COVERAGE_FLOOR` is seeded at 80 and the MSI floors at 0. I could not measure either locally: this machine has only Xdebug, a full `unit,integration` run under it was still under 2% after ten minutes, and `pecl install pcov` fails here on a missing `pcre2.h`. Both jobs publish their real figure to the run summary; I will set the floors from the first green run on this PR rather than guess.

Related to #2859


Chemaclass:
Superseded by four focused PRs, all merged green: #2870 (public API contract + snapshot gates), #2871 (language spec, Clojure divergences, upgrade guide), #2872 (`@internal` sweep), #2873 (CI platform matrix, coverage, benchmark comparison, mutation testing).

Same scope, nothing dropped. This one bundled five concerns into 828 files; splitting it put the 796-file mechanical sweep in its own PR and left ~41 reviewable files across the other three. Each branched off freshly-merged `main`, so every CI run reflected the real merged state and no PR ever conflicted.

Issue #2785, opened 2026-07-20
Break the remaining Shared -> Compiler dependency cycle (CompilerFacadeInterface signatures)


Chemaclass:
## Background

#2781 broke most of the Shared DIP cycle by moving pure contracts/DTOs into `Phel\Shared` (Build, Command, Interop now cycle-free). One back-edge cluster remains: `Shared/Facade/CompilerFacadeInterface` references Compiler-owned types (AST nodes, analyzer environment, lexer types) across 11 edges, keeping `{Compiler, Config, Filesystem, Lang, Shared}` in one SCC.

This was deliberately deferred: those signatures are the compiler's real public surface, and moving AST/lexer types into Shared wholesale would gut the Compiler module. It needs a design pass, not a mechanical move.

## Possible directions (to evaluate)

1. Slim `CompilerFacadeInterface` to primitives/DTOs at the boundary (serialize node/env handles) and keep rich types behind the facade.
2. Introduce narrow Shared-safe interfaces for the few types leaking through (e.g. a `SourceLocation`/`NodeInterface` contract) and let Compiler implement them.
3. Accept and document the cycle as intentional (Compiler is the language core; everything depends on it anyway).

Module-level graph + per-edge details are in the dependency analysis attached to #2781's review notes (Tarjan SCC over `use` statements, comments/strings stripped).

## Acceptance criteria

- [ ] Decision recorded (break vs document-as-intentional) with rationale in `src/php/Shared/CLAUDE.md`
- [ ] If breaking: `{Compiler, Config, Filesystem, Lang, Shared}` SCC decomposes without violating Gacela facade/factory rules
- [ ] `composer test` green

PR #2874, opened 2026-07-26
ref: type cross-module wiring and enforce module boundaries (gacela 2.0 ready)

Chemaclass:
## 🤔 Background

Gacela 2.0 ships this week. This branch started as the migration and turned into two separable things:

1. **The migration is a no-op for Phel's source**: verified by running the whole suite against `gacela-project/gacela: dev-main`.
2. **Two features Phel was already entitled to on 1.21 and had not adopted**: typed cross-module wiring, and the rule that enforces module boundaries.

Only (1) needs 2.0. So the constraint here stays at **`^1.21`** and the bump is a one-line follow-up once 2.0 is tagged, rather than putting a moving dev branch in `composer.lock` where a release could pick it up.

## 💡 Goal

Land everything that works on stable gacela, and prove 2.0 is a non-event before it ships.

### Cross-module dependencies are keyed by interface, not by string constant

Providers already used `#[Provides]`, but keyed on `FACADE_*` constants. Gacela types `getProvidedDependency(Foo::class)` as `Foo` while string keys stay `mixed`, so all 31 call sites carried a hand-written `@var`.

```diff
-    public const string FACADE_RUN = 'FACADE_RUN';
-
-    #[Provides(self::FACADE_RUN)]
-    public function runFacade(Container $container): RunFacade
+    #[Provides(RunFacadeInterface::class)]
+    public function runFacade(Container $container): RunFacadeInterface
```

```diff
-        /** @var RunFacadeInterface $facade */
-        $facade = $this->getProvidedDependency(WatchProvider::FACADE_RUN);
-
-        return $facade;
+        return $this->getProvidedDependency(RunFacadeInterface::class);
```

**30 constants and 31 `@var` docblocks removed.** The line count is not the point: a `@var` is an *assertion* that can lie about what the provider actually bound, and nothing checks it. The binding id now **is** the type, so the two cannot disagree. I confirmed the typing is real rather than a silent `mixed` fallback by planting a deliberate mismatch and watching PHPStan catch it.

Each provider method now also returns the interface it publishes, so the signature reads as the contract rather than the implementation behind it.

### `CrossModuleViaFacadeRule` is enabled

`.claude/rules/php.md` has required modules to reach each other only through a Facade for a long time. Only hand-rolled tests enforced parts of it, and they check *Factories*, not Application classes. `Phel\Shared`, `Phel\Lang` and `Phel\Config` are declared shared kernels (they are what every module is *supposed* to reach into; see `docs/stability.md`).

It found 35 violations, in three groups, all pinned with counts rather than waived so a **new** cross-module reach fails the build:

- **24**: the console composition root registering every module's commands. Its actual job.
- **6**: `FilesystemFacadeInterface` lives in `Phel\Filesystem\` while every other cross-module contract lives in `Phel\Shared\Facade\`. Referencing it *is* going through a facade; the rule cannot tell. Moving it is the real fix and is **breaking** against the surface frozen in #2870, so it waits for a major.
- **5**: genuine reaches into another module's `Domain`: `NodeEnvironment` from Api and Build, `DeprecationWarnings` from Console, `ExtractorException` from Run. Pinned individually so fixing one removes its own entry rather than shrinking an anonymous total.

## 🐛 A bug I introduced, and the test that pins it

The class-string keying is safe for facades: they publish `RunFacadeInterface::class` and resolve `RunFacade::class`, two distinct ids. **Two bindings have no separate interface** (`PhpConfigReader`, `LintFacade`), so publishing the class the body asks the container for made the binding **re-enter itself** until the stack ran out.

PHP 8.5 caught it. Its new `zend.max_allowed_stack_size` guard turns this into `Maximum call stack size … Infinite recursion?`; on 8.4 it stayed just under the limit, which is exactly why the local run went green and CI did not.

Both now key on a constant with the constraint written down, and `ProviderBindingIdTest` pins the class of bug. It is worth pinning because the failure is not a resolution error; it is a stack overflow raised deep inside the container, only when something first resolves that binding, which may be one unrelated integration test.

## ✅ Gacela 2.0: verified, not assumed

Run against `dev-main` (`006c6af`) before reverting the constraint. Every breaking change was already satisfied:

| Removed in 2.0 | Phel |
|---|---|
| `AbstractDependencyProvider` | 0 uses; providers are already `<Module>Provider extends AbstractProvider` |
| `*DependencyProvider.php` filename resolution | 0 files |
| `GacelaConfig::addMappingInterface()` | 0 uses |
| `DocBlockResolverAwareTrait` | 0 uses |
| Concrete `Gacela\Container\Container` type-hints | 0 uses |

All **24** classes using `ServiceResolverAwareTrait` already declare `#[ServiceMap]`, so removing the PHPStan pillar suppression costs nothing, and Phel now gets *typed* accessors instead of `mixed`. Nothing redeclares a now-typed constant on `AbstractSetupGacela`/`ConfigInterface`. PHP floor (`>=8.3`) and Symfony range (`^7.0 || ^8.0`) both widen; Phel's `>=8.4` and `^7.3 || ^8.0` sit inside.

Green on `dev-main`: unit 4365, integration 1113, core 6564, Psalm 0, PHPStan L9 0, PHAR 10/10 at 2.15 MiB.

## 📄 Docs 2.0 makes stale

`.agnostic-ai/rules/php.md` said *"Use `DependencyProvider` for cross-module dependencies"*, and `skills/module-new` scaffolds `<ModuleName>DependencyProvider.php`. On 2.0 a module created from that skill **resolves as silently absent**, the exact trap gacela's upgrade guide warns is hardest to debug. Renamed throughout, with the filename-suffix rule spelled out.

## 📊 Measured, not assumed

- **Container `compile()`/`writeCompiledCache()` not wired.** `cache:warm` moves `phel eval` 0.30s → 0.30s; gacela resolution is not the bottleneck. Phel already delegates `cache:warm` to gacela's own command, so 2.0's work there lands for free.
- **`AbstractFactory::make()` not adopted** on compiler paths: it routes construction through container reflection, a deoptimisation where it matters most.
- **Typed config accessors not adopted**: Phel's config is deliberately lenient; they throw on mismatch instead of coercing.
- **`debug:graph --check` not adopted**: `ModuleDependencyCycleTest` already knows Phel's layering and guards the deliberate `Shared -> Compiler` cycle (#2785). Two mechanisms for one invariant is worse than one.

## ⏭️ Follow-up

Bump `gacela-project/gacela` to `^2.0` once it is tagged. Nothing else changes; this branch proves it.

Related to #2859


Chemaclass:
Merged with `gacela-project/gacela: ^1.21`, deliberately. Everything here works on stable 1.21, and the whole suite was verified green against `dev-main` (unit 4365, integration 1113, core 6564, Psalm 0, PHPStan L9 0, PHAR 10/10), so the 2.0 bump is a one-line follow-up rather than a moving dev branch sitting in `composer.lock` where a release could pick it up.

Follow-up once gacela tags 2.0 this week: bump the constraint to `^2.0`. Nothing else changes.

Issue #2879, opened 2026-07-27
Document heredoc for multi-line eval in `phel eval -h`


jasalt:
Follow-up half of [phel-lang.org#280](https://github.com/phel-lang/phel-lang.org/issues/280). That issue covers the website docs; this one covers only the CLI help text in this repo.

## What works today but is undocumented

`phel eval -` reads the expression from stdin, so a heredoc evaluates multiple forms with no quoting problems. Added in #1494, verified on `main` (`93fe4a5cd`):

```bash
./bin/phel eval - <<'PHEL'
(ns app)
(println (php/+ 40 2))
PHEL
# 42
```

## The gap

`EvalCommand::configure()` (`src/php/Run/Infrastructure/Command/EvalCommand.php:43-49`) sets a `setHelp()` block with two examples, neither of them a heredoc:

```
Evaluates a Phel expression (or stdin) and prints its value.

<info>Examples:</info>
  <comment>phel eval '(+ 1 2)'</comment>      Evaluate an inline expression
  <comment>echo '(* 6 7)' | phel eval -</comment>   Evaluate from stdin
```

The argument description at `:53` says `Use "-" to read from stdin.` and stops there, so nothing tells a reader that `-` plus a heredoc is the reliable way to evaluate several forms.

## Why heredoc is worth calling out

From the [brepl heredoc pattern](https://github.com/licht1stein/brepl/#heredoc-pattern-for-reliable-evaluation), which the docs issue quotes:

- No quoting issues: everything between `<<'PHEL'` and `PHEL` is literal, so single quotes, `$`, and backslashes pass through untouched.
- One pattern for every size of input, simple to complex.
- Multi-line friendly, so a `(ns …)` plus several forms reads normally.
- Extending it means adding a line, not re-quoting the shell argument.

This matters most for agents and scripts, which is where shell-quoting failures actually bite.

## Acceptance criteria

- [ ] `EvalCommand`'s `setHelp()` block gains a heredoc example, quoted delimiter (`<<'PHEL'`) so the shell does not interpolate.
- [ ] One line on why to prefer it for multi-form input. Keep the help block short; the reasoning belongs on the website.
- [ ] `phel eval --help` output stays readable at 80 columns.
- [ ] Any existing test asserting the help text is updated (`grep -rn "Evaluate from stdin" tests/`).

## Out of scope

[...]

PR #2880, opened 2026-07-27
docs: add heredoc example for multi-line eval to phel eval -h

jasalt:
There's slight issue with the indented help output being not very copy-paste friendly in `/bin/phel eval -h` output:

image

The heredoc closes only if the `PHEL` doesn't have whitespace before it which get's easily copied from the indented help output:

```
  phel eval - <<'PHEL'
  (ns app)
  (println (+ 40 2))
  PHEL    # <- invalid closing of heredoc
```

Following works:
```
  phel eval - <<'PHEL'
  (ns app)
  (println (+ 40 2))
PHEL 
```

It's bit tricky, should (A) the output be modified to make it easier to copy paste, or (B) should mention be added that it should be made sure that closing `PHEL` would not have whitespace before it... Or alternatively (C) leave as is and let user to discover it themselves, it's documented in the docs website in copy-paste friendly way anyways.

Resolving https://github.com/phel-lang/phel-lang/issues/2879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Add heredoc example to `phel eval -h` help output for reliable multi-line one-shot evaluation
+
 ### Added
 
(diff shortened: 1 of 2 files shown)


jasalt:
Added changelog entry,

Chemaclass (approved):
thanks!

Issue #2886, opened 2026-07-27
`phel eval` / `phel nrepl` require issue


jasalt:
Noted `ns` form `:require` issue when hooking up nREPL to a project. This might have been there for a while, maybe all the while since built-in nREPL was added as I haven't gotten to try it yet much.

- Phel v0.49.0-beta#6d7f719
- PHP 8.5.8 / 8.4

Test environment preparation:
```
mkdir issue-demo && cd issue-demo
mkdir src

# Demo file
cat << EOF > src/demo.phel
(ns demo)
(defn plus [a b] (+ a b))
EOF

composer require phel-lang/phel-lang
```

Following evaluation in REPL (`vendor/bin/phel repl`) works:

```
(ns demo.test
  (:require demo :refer [plus]))
(plus 1 2)
;; => 3
```

However, it fails with `phel eval`:
```
vendor/bin/phel eval - <<'EOF'
(ns demo.test
  (:require demo :refer [plus]))
(plus 1 2)
EOF
```
Erroring: `[PHEL001] Cannot resolve symbol 'plus'.`

Same affects nREPL evaluation. Test client installation:
```
# Babashka
curl -sLO https://raw.githubusercontent.com/babashka/babashka/master/install
chmod +x install
./install --dir ~/.local/bin/

# brepl
curl -fsSL https://raw.githubusercontent.com/licht1stein/brepl/master/brepl -o ~/.local/bin/brepl
chmod +x ~/.local/bin/brepl
echo 'export PATH="$PATH:$HOME/.local/bin"' >> ~/.$(basename $SHELL)rc && exec $SHELL
```

Demo evaluation:
```
vendor/bin/phel nrepl

# other terminal
brepl -p 7888 <<'EOF'
(ns demo.test
  (:require demo :refer [plus]))
(plus 1 2)
EOF
```
Errors: `AnalyzerException: Cannot resolve symbol 'plus'.`

Chemaclass:
Reproduced on `main` (`f0e99d496`), and the cause is a one-line gate. Thanks for the exact repro, it made this quick.

## It is not `:refer`

`:refer` is a red herring. All three spellings fail under `phel eval`, and the same file works under `phel run`:

```bash
phel eval - <<'EOF'   # Cannot resolve symbol 'plus'
(ns demo.test (:require demo :refer [plus]))
(plus 1 2)
EOF

phel eval - <<'EOF'   # Cannot resolve symbol 'demo/plus'
[...]

PR #2890, opened 2026-07-27
fix: phel eval and phel nrepl now resolve :required namespaces

jasalt:
PoC fix to https://github.com/phel-lang/phel-lang/issues/2886 (Pi / kimi-k3):

## Root cause                                                                                                           
                                                                                                                      
 In statement (eval) emit mode, NsEmitter::emitRequiredNamespaces() emits runtime code that locates and evalFiles     
 required namespaces — but it only populated the source-dir list when phel.core/*repl-mode* was set. Only phel repl   
 sets that flag, so under phel eval / phel nrepl the dir list was empty, getDependenciesForNamespace([], ['demo'])    
 returned nothing, demo.phel was never loaded, and the analyzer then failed with Cannot resolve symbol 'plus' on the  
 referring form.                                                                                                      
                                                                                                                      
 The *repl-mode* gate was added in ad089a376 with a narrower intent: keep runtime dependency loading out of builds    
 (phel build evaluates each form while compiling). Gating on repl-mode was over-broad.                                
                                                                                                                      
## Fix                                                                                                                  
                                                                                                                      
 src/php/Compiler/Domain/Emitter/OutputEmitter/NodeEmitter/NsEmitter.php — the gate is now                            
 !\Phel\Build\BuildFacade::isBuildMode(): builds still skip runtime loading (the original concern), while every other 
 eval context (repl, phel eval, nREPL, run/test) resolves and loads requires.                                         
                                                                                                                      
## Tests                                                                                                                
                                                                                                                      
 - New tests/php/Integration/Run/EvalNamespaceRequire/ — plain eval session (no repl-mode) with :refer and :as        
   requires. Confirmed it fails on master with exactly your error (Cannot resolve symbol 'plus') and passes with the  
   fix.                                                                                                               
 - Repaired RequireBuildModeTest — it was vacuous on master: nothing has read its phel.repl/src-dirs setup since      
   a9664b758, and its fixture's (php/constant "__DIR__") can never execute (fatals in plain PHP), so the require      
   silently never fired. Fixed the fixture to use phel's real __DIR__ magic symbol, added a phel-config.php (src-dirs 
   + enable-compiled-code-cache => false, since warm cache hits skip runtime dep loading by design — frameworks       
   preload deps in order), and added an assertion that the namespace actually loads. It now pins both intended        
   semantics: require-loads happen, and they happen in build mode.                                                    
 - Updated NsEmitterTest and KebabNamespaceRequireTest for the new gate.                                              
                                                                                                                      
## Verification                                                                                                         
                                                                                                                      
 - Reporter scenarios: phel eval (stdin + inline arg), phel repl, phel nrepl (real bencode client) → all print 3      
 - phel build/run/test behavior unchanged (build output diffed against master)                                        
 - Full Unit + Integration suites pass (only pre-existing env failures: missing bigint ext, watcher prefs — identical 
   on master); phel's own phel test: 6564/6564                                                                        
 - phpstan, psalm, php-cs-fixer clean; CHANGELOG entry added under Unreleased → Fixed


tests/php/Integration/Run/RequireBuildMode/Fixtures/example/game.phel
jasalt, line 1: Irrelevant code style change.

Chemaclass:
Thanks @jasalt, this was genuinely useful rather than just a duplicate. #2892 landed instead, but three things came from here, and one of them was load-bearing.

**Adopted from this PR:**

1. **`BuildFacade::isBuildMode()`.** My version read `*build-mode*` out of the registry by munged string with `addslashes`, reinventing an API that already existed. Yours is shorter and emits cleaner PHP.
2. **The cache-forcing `phel-config.php`.** This one mattered most. My strengthened test passed on a cold cache and **failed on a warm one**: in cache mode the `ns` emission carries no dependency loading, so the require under test never fires. Without `enable-compiled-code-cache => false` the test was non-deterministic, which is the same cache masking that made the original bug pass CI for so long. With it, green on three consecutive runs.
3. **`__DIR__`** in the fixture instead of `(php/dirname *file*)`. Verified both resolve to the same directory, so yours is simply shorter.

**Independent convergence on the diagnosis**, which is the strongest signal either of us had: you reached the same gate (build mode, not `*repl-mode*`) and hit the same two latent breaks in `RequireBuildMode` (no `phel.core` compiled, and `(php/constant "__DIR__")` throwing because PHP's `constant()` reads defined constants rather than compile-time magic ones). Two separate paths to the same answer.

**What #2892 added on top:** nREPL verified end to end over raw bencode rather than by inference, an assertion that `*repl-mode*` is *gone* from the emitted output and not merely that `isBuildMode()` is present, and a note on why setting `*repl-mode*` from `EvalCommand`/nREPL would have been the wrong fix: that flag also suppresses duplicate-definition errors (`GlobalEnvironment::shouldThrowOnDuplicateDefinition()`) and injects the `phel.repl` refers (`ReplReferInjector`), so it would have widened two unrelated behaviours.

The merge commit carries `Co-authored-by: jasalt`.

Also filed #2891 from your repro: a `:require` that resolves to no file is silent in every mode, which is why this surfaced as `Cannot resolve symbol` at the call site instead of an error at the require. Different cause, so it is tracked separately.

Closing in favour of #2892.

PR #2893, opened 2026-07-27
feat: nrepl server writes the bound port to .nrepl-port

jasalt:
Testing this for a bit before ready for review ...

## Commit message

phel nrepl now maintains a Clojure-standard .nrepl-port file in the CWD:

 New file — src/php/Nrepl/Infrastructure/NreplPortFile.php
 - write(int $port) creates .nrepl-port containing the port (throws a clear RuntimeException if the directory is unusable); delete() removes it, idempotently.

 src/php/Nrepl/Infrastructure/Command/NreplCommand.php
 - After the socket binds, writes .nrepl-port with $server->port() — so --port=0 records the real randomly-bound port.
 - Removal on every stop path:
     - try/finally around $server->run() for the graceful path,
     - register_shutdown_function backstop for fatal-error exits, - new SIGINT/SIGTERM handlers (when ext-pcntl is present) that call $server->stop(), so Ctrl+C/kill make the accept loop return and the file is removed instead of being left pointing at a dead server. Handlers and the shutdown function are registered before the file is written, so there's no signal window.
 - Help text documents the behavior.

 src/php/Nrepl/NreplFactory.php — createPortFile(string $directory) following the
 module's factory convention.

 Tests
 - tests/php/Unit/Nrepl/Infrastructure/NreplPortFileTest.php — write/overwrite/delete/missing-dir (5 tests).
 - tests/php/Integration/Nrepl/NreplCommandPortFileTest.php — spawns the real bin/phel nrepl --port=0 in a temp project, asserts the file appears, that its port accepts TCP connections, then SIGTERMs and asserts exit 0 and file removal.

 CHANGELOG.md — entry under Unreleased.

 Verification: 111 Nrepl unit+integration tests, 39 architecture tests pass;
 php-cs-fixer, PHPStan, Psalm, Rector clean; manual hammer runs (SIGINT/SIGTERM
 ×12) all shut down cleanly with the file removed, and a failed bind writes
 nothing.


Chemaclass:
Completed and reviewed. Pushed on top:

- **`delete()` is now ownership-aware.** `register_shutdown_function` fires on every exit, including one where `write()` never ran (a failed bind, for instance). It would then remove a `.nrepl-port` that a *different* server had written in the same directory, making that server undiscoverable. `NreplPortFile` tracks whether this instance wrote the file and `delete()` no-ops otherwise. Unit test added for it (the class stops being `readonly`; the directory stays so).
- **`getcwd()` returning false** no longer reaches the message as an empty directory name; it falls back to `.`, which still names the working directory for `file_put_contents`.
- **`src/php/Nrepl/CLAUDE.md`** documents the port-file contract: written once listening, removed on the `finally`/shutdown/signal paths, no-op without ext-pcntl, and the ownership rule.
- Merged `main`, plus a php-cs-fixer and a Rector formatting fix the new files needed.

Two behaviours left as they are, both matching Clojure: a second server in the same directory overwrites the first one's file, and the file lands in the CWD rather than a resolved project root.

`composer test` green (6586 core assertions).

Chemaclass (approved):
Approving. The port file is the right shape: written only once the socket is listening (so `--port=0` records the real bound port), removed on the graceful path, the fatal path and both signals, and `NreplSocketServer::stop()` is idempotent so the doubled `stop()` between the handler and the `finally` is harmless.

The integration test earns its keep: it spawns the real `bin/phel nrepl --port=0` in a temp project, connects to the port it read from the file, then SIGTERMs and asserts both a zero exit and the file's removal. It skips cleanly without ext-pcntl, which is what keeps Windows green.

Factory boundary is respected (`createPortFile()` news a class from its own module), and `.gitignore` covers the artifact.

Chemaclass:
CI failure was a bug in the new test, not in the feature.

`waitForExit()` read the exit status with `proc_close()` after a `proc_get_status()` loop had already seen the process stop. `proc_get_status()` reaps the child itself, so `proc_close()` had nothing left to wait for and answered `-1`; the assertion on a graceful SIGTERM shutdown then failed. The exit code now comes from the status array where it actually lives, and a timeout fails with a message instead of returning a value that looks like an exit code.

Passed locally on macOS by timing luck, which is why it only showed on ubuntu.

One more fix while in here: an unwritable working directory used to abort `phel nrepl` entirely, since `write()` throws and the outer catch returns FAILURE. Losing the discovery file should not cost the session, so it now reports the reason plus the explicit port and keeps listening:

```
nREPL server started on 127.0.0.1:59986
Cannot write .nrepl-port to "/private/tmp/phel-ro-test". Connect on port 59986 explicitly.
```

`composer test` green, and the integration suite passes under paratest (the CI runner) as well as phpunit.

jasalt:
Good improvements, cheers.

PR #2895, opened 2026-07-27
fix: nREPL server reports session namespace changes to client

### LINKED ISSUE #2894: nREPL not updating namespace to Cider prompt
When `./bin/phel nrepl` session starts, Emacs Cider client shows prompt `user>` correctly, but after changing namespace `(ns foo)` the prompt does update accordingly with the updated namespace while `*ns*` shows it. `phel repl` does have the issue but shows the updated namespace in prompt.

jasalt:
Closes #2894

## 🤔 Background

The two REPLs read the current namespace from different sources:

- **`phel repl`** builds its prompt from live compiler state — `ReplPrompt` reads `GlobalEnvironment::getNs()`, which `(ns foo)` updates. Always correct.
- **`phel nrepl`** reports the namespace in the `ns` field of eval responses from `Session::$namespace`, which was initialized to `'user'` and never updated (`Session::setNamespace()` had zero callers). CIDER drives its prompt from that response field, so it stayed `user>` forever while `*ns*` — evaluated against the real global environment — correctly said `foo`.

The same field is also what an editor primes its prompt from on connect, and neither of the two things it sends first carried an `ns`: an empty init eval (`cider-repl-init-code ""`) was rejected as a missing `code` param, and CIDER's default Clojure-only init code fails to evaluate. Result: a fresh connection sat at `nil>`.

## 💡 Goal

Report the session's real namespace in every eval response — including error frames and the connect-time init eval — so editor clients (CIDER, Calva, ...) render the prompt the terminal REPL already gets right.

## 🔖 Changes

- `EvalResultResponder` — the single funnel shared by `eval`, `load-file`, `reload` and `run-tests` — syncs the session namespace from the compiler's `GlobalEnvironment` after every eval, via `Munge::displayNs`: the exact source and normalization the terminal REPL prompt uses. Failed and incomplete evals are safe, since `StructuredEvaluator` restores the environment snapshot on failure, making the sync a no-op there.
- The `ns` is attached to eval **error** frames too, so a client can still initialize its prompt when the first eval errors.
- `EvalOp` distinguishes an absent `code` param from an empty one: missing gets the protocol's `no-code` status (matching the reference nREPL, which replies `["error" "no-code" "done"]`), empty/whitespace-only is a no-op replying `done` with the current `ns` — the reference nREPL reads it as EOF the same way.
- `OpRequest::optionalStringParam()` gives ops that "absent vs. empty" distinction without reaching into `raw`.
- `Session::DEFAULT_NAMESPACE` replaces the duplicated `'user'` literal; `Session::namespace()` and `EvalOp`'s docblocks now say what is actually true — eval runs in the compiler's global environment shared by all sessions, and the session namespace is the mirror of it that the `ns` response field and `LookupOp` read.
- `NreplFactory` / `NreplProvider` provide `CompilerFacadeInterface` to the responder; `src/php/Nrepl/CLAUDE.md` and `CHANGELOG.md` document the behaviour.

**Side benefit:** `lookup`/`info`/`eldoc` resolve unqualified symbols against the session namespace, so symbol resolution after an `ns` switch is now correct too.

### ✅ Verification

- **Integration test** over a live socket (`NreplServerTest`): empty init eval → `user`, failed eval → `user`, `(+ 1 2)` → `user`, `(ns foo)` → `foo`, and it sticks. Fails without the fix (`user` instead of `foo`).
- **Unit tests** for the sync (backslash→dot normalization, uninitialized environment, empty environment namespace, session-less eval, empty-code paths, non-string `code` param) plus a new `OpRequestTest`.
- Full local suite green: 4417 unit, 1138 integration, 6575 core (`phel test`), php-cs-fixer, Psalm, PHPStan, Rector.
- **Live check** against a real `./bin/phel nrepl` server with a raw bencode client: `user` → `foo` → `foo`.


jasalt:
image

Starting to look good.

Chemaclass:
Reviewed and pushed two follow-ups on top:

- merged `main` (only CHANGELOG conflicted)
- `respond()` now answers an incomplete form before syncing the namespace. `StructuredEvaluator` restores its environment snapshot before reporting an unfinished form, so the namespace has not moved and the frame carries no `ns`; syncing first wrote the session a value it already held. Regression test added.

The namespace being read from the process-wide `GlobalEnvironment` means two connected editors share it: A doing `(ns foo)` moves B's prompt. That is not from this PR (evaluation was always shared, and B's code always ran in A's namespace) and this only stops the `ns` field from claiming otherwise. Tracked separately in #2906.

`composer test` green, 126 nREPL tests.

Chemaclass (approved):
Approving. The fix is at the right layer: `EvalResultResponder` is the single funnel for eval, load-file, reload and run-tests, and it mirrors the same `GlobalEnvironment::getNs()` the terminal REPL prompt reads, normalized through `Munge::displayNs`. `Session::setNamespace()` having had zero callers was the whole bug.

Wiring follows the module rules (`CompilerFacadeInterface` via `NreplProvider`, no cross-module `new`, `CLAUDE.md` dependency table updated), and `optionalStringParam()` is the right way to separate an absent `code` param from an empty one, so a missing param can answer with the protocol's `no-code` status.

One client-visible change worth noting in review: a missing `code` param now returns `no-code` instead of `eval-error`. That matches reference nREPL and the CHANGELOG covers it.

Issue #2876, opened 2026-07-27
Decide the one way to refer to a PHP class once `\` stops being a namespace separator


Chemaclass:
Feedback from @jasalt on the 1.0 roadmap (#2859):

> Are there any downsides to leaving `\` out for referring PHP classes? It could be easier if there was simply one way to refer them. As `\` blows up the Clojure reader it would be nice to leave it out in my opinion.

## Current behaviour, verified on `main`

`\` is **overloaded**. It means two different things depending on position:

```phel
(new DateTime "2020-01-01")          ; ✅ works: global-namespace PHP class, no backslash
(new \Phel\Lang\Keyword "x")         ; ✅ resolves: leading \ marks a PHP FQCN
(new Phel\Lang\Keyword "x")          ; ❌ PHEL001 Cannot resolve symbol 'Phel\Lang\Keyword'
```

The third fails because without a leading `\`, `Foo\Bar` is read as a **Phel namespace** reference using the deprecated `\` namespace separator, the one deprecated in favour of `.` by #1567 (closed) and still shipped.

So today `\` is simultaneously:

1. the Phel namespace separator (deprecated by #1567, `.` is the target, removal tracked in #2827), and
2. the marker for a PHP fully-qualified class name.

That overload is the actual source of the friction jasalt describes, and it is also why `\` "blows up the Clojure reader": a Clojure reader treats `\x` as a character literal.

## Why this is worth deciding at 1.0

`docs/spec/language-surface.md` freezes the reader syntax for `1.x`. `\` is the only reader-level item whose fate is still open; #1567 delivered the deprecation and the `.` migration path, and #2827 owns whether the separator meaning is removed. Once meaning (1) is gone, `\` becomes unambiguous (it marks a PHP FQCN and nothing else) and Phel gains exactly the "one way to refer to them" jasalt asks for, without removing anything users rely on.

## What this issue is for

Deciding and writing down the answer to: **once `\` stops being a namespace separator (#2827), is a leading `\` still required for a namespaced PHP class?**

Options to evaluate, with the trade-off each carries:

- **A. Keep the leading `\` as the FQCN marker.** Unambiguous, no reader change, matches PHP's own spelling. Costs the Clojure-reader friction jasalt raises.
- **B. Allow `Foo\Bar` without the leading `\` once the separator meaning is gone.** Fewer characters, one spelling. Needs a resolution rule for "is this a Phel symbol or a PHP class?" and a decision on what wins on collision.
[...]

Issue #2827, opened 2026-07-25
Next major: remove the `\` namespace separator (last deprecated surface item)


Chemaclass:
**One item is left: the `\` namespace separator.** Its fate is decided, its date is decided, and what remains is execution at the next major. This issue is the execution tracker plus anything new that lands on the deprecation list before then.

## Done, verified absent from `main`

| Group | Removed in | Verified by |
|---|---|---|
| Reader syntax (`#\| \|#`, bare `#`, `\|()`, `,` / `,@` unquote, `foo$`) | #2864 | `grep -rn DEPRECATABLE_TYPES src/php/` → no hits |
| `^:reference` fn-param alias | #2863 | `MethodEmitter::isByRef()` reads only `by-ref` |
| `set-meta!`, `phel.test/print-summary` | #2863 | absent from `src/phel/`; `print-summary-block` is a different private helper |
| CLI aliases `index --out`, `config --json` | #2862 | `src/php/Shared/Console/DeprecatedOptionWarner.php` deleted |
| Legacy REPL history migration | #2862 | `ReplHistoryPathResolver.php` is 28 lines, no migration branch |

Their before/after entries live in [`docs/migration/removed-deprecated-core-fns.md`](https://github.com/phel-lang/phel-lang/blob/main/docs/migration/removed-deprecated-core-fns.md) and the reader-syntax list in `docs/spec/language-surface.md`.

## Decided: `\` retires at the next major, not at 1.0

[ADR 0015](https://github.com/phel-lang/phel-lang/blob/main/docs/adr/0015-a-php-class-is-named-with-dots.md) settled this on 2026-08-02, reviewed by @jasalt. §4: the leading `\` retires **with** the separator, at the next major, and is never promoted to a permanent marker. §5: the `def`-shadows-a-class warning becomes a refusal in that same major.

Three things pin the date, so it is not reopenable without superseding an accepted ADR:

- ADR 0015: "**Not at 1.0.** `\` keeps working for all of `1.x`."
- `docs/migration/upgrade-0.49-to-1.0.md` promises the same in writing.
- [ADR 0014](https://github.com/phel-lang/phel-lang/blob/main/docs/adr/0014-announce-the-separator-deprecation.md): the notice period only *starts* at `1.0.0`, when the deprecation begins announcing without a flag.

[...]

PR #2911, opened 2026-07-28
fix(compiler): resolve all-caps PHP classes before constants

jasalt:
## 🤔 Background

This came up while diagnosing backslash and `php/` free alternate forms for calling `php/PDO::ATTR_ERRMODE`. Seems that all caps `PDO` was getting read as constant instead of class name.

Status before:
- `\PDO/ATTR_ERRMODE` -> works, but fails without backslash
- `(.-ATTR_ERRMODE \PDO)` -> works, fails without backslash
- `php/PDO::ATTR_ERRMODE` works

Afterwards following work, no backslash required:
- `PDO/ATTR_ERRMODE`
- `(.-ATTR_ERRMODE PDO)`

AI used for diagnosis / PoC implementation (Pi agent + gpt-5.6-sol)

## Commit message

PDO failed to resolve as a class because its all-uppercase name matched Phel's global-constant pattern first. As a result, PDO/ATTR_ERRMODE, (.-ATTR_ERRMODE PDO), and (new PDO ...) raised "Undefined constant PDO".

Resolve existing PHP classes, interfaces, traits, and enums before the global-constant fallback, including through :use aliases. If a class and global constant share a name, the class wins; php/NAME remains the explicit constant escape hatch.

Document the resolution rule and add coverage for PDO, aliases, and class/constant collisions.

## Other considerations

Should be checked that the design makes sense / aligns with the Phel direction. Can be used as reference for better fix or used with possible modifications.

Related recent issues:

- https://github.com/phel-lang/phel-lang/issues/2876 
- https://github.com/phel-lang/phel-lang/issues/2827


Chemaclass (approved):
Correct fix at the right layer, and verified end to end.

The root cause is real: `looksLikePhpConstantName()` ran before the bare-class fallback, so `PDO` matched `/^[A-Z_][A-Z0-9_]*$/` and became a `PhpVarNode` before it could ever be a class. `QualifiedMemberExpander::isStaticMethodOnly()` then saw a non-`PhpClassNameNode` and fell through to `(php/:: PDO ATTR_ERRMODE)`, which emits `PDO` as a constant. Reproduced on `main` for all three spellings, all three fixed on this branch.

Checked the ordering swap for collateral damage and found none: the first arm of `looksLikeBareClassName` requires a lowercase character, so it never overlapped the constant pattern, and only all-caps names change meaning. `PHP_EOL`, `JSON_HEX_AMP`, `M_PI`, `SORT_REGULAR` and `NAN` all still read as constants.

The autoloading existence check is the right call and the comment defending it is accurate. The cost worry does not hold up: composer memoizes misses (`ClassLoader::$missingClasses`), and it is compile-time only. Measured 0.032 ms for ten distinct cold names across class/interface/trait probes.

Pushed one refactor commit on top:

- `enum_exists($name, false)` was unreachable. Enums are classes, so the preceding autoloading `class_exists()` already returned true. Removed.
- The PR added a fifth spelling of the class-like existence predicate, and the existing four disagreed about which kinds they covered. Extracted `Compiler/Domain/Analyzer/PhpClassLike::exists()` and wired it into `SymbolResolver` (3 sites) and `UseAliasRegistrar`. Kept it inside `Compiler` rather than `Shared`, since every `Shared` class is public and this is an internal helper. Left `QualifiedMemberExpander` and `InterfaceImplementationsAnalyzer` on their narrower checks on purpose: widening them to traits changes analysis behaviour and belongs in its own change.
- Collapsed the four repeated `getNamespace() === null` guards into `resolveBareHostSymbol()`, so the collision ordering this PR is about lives in one named place with its rationale.
- Extracted the duplicated class-plus-constant test fixture into `DefinesClassConstantCollisionTrait`, and added `PhpClassLikeTest`. One of its cases pins the subtle part: an interface defined by an on-demand loader is found only because the class probe already spent the single autoload attempt, so it fails if anyone switches to `class_exists($name, false)`.
- Trimmed the spec paragraph, which stated the rule twice with a full sentence bolded.

`composer test` green locally on PHP 8.5.8, and all 20 CI checks pass.

Unrelated, for a follow-up: `UseAliasRegistrar::register()` carries two consecutive docblocks, so PHP honours only the second and the prose one is invisible to IDEs and static analysis.

PR #2935, opened 2026-08-02
docs(adr): record how a PHP class is named after the separator goes

### LINKED ISSUE #2876: Decide the one way to refer to a PHP class once `\` stops being a namespace separator

Chemaclass:
## 🤔 Background

#2876 asks which spelling replaces `\` for a PHP class. It stayed open because two questions kept being conflated: `\` is the namespace **separator**, and a leading `\` is a **marker** meaning "this is a host class". Dropping the separator does not decide the marker.

Two arguments from @jasalt settle it, and this record only follows them to their conclusion:

- [#2859 (comment)](https://github.com/phel-lang/phel-lang/issues/2859#issuecomment-5152721973): write Phel's own sources in the dot spelling. Done in #2926, which is what makes any argument from them honest.
- [#2876 (comment)](https://github.com/phel-lang/phel-lang/issues/2876#issuecomment-5156046021): Clojure refuses a `def` that shadows a class, which is what makes a bare class name safe there. Shipped as a warning in #2934.

Closes #2876

## 💡 Goal

One spelling, and it is the one a Clojure reader already knows.

```phel
(new DateTime "now")                  ; a root class, bare
(new Phel.Lang.Keyword "x")           ; a namespaced class, dotted
Phel.Lang.Symbol                      ; in value position
```

## 🔖 Changes

Documentation only. Nothing in `1.x` changes, and no code moves.

**[ADR 0015](../blob/main/docs/adr/0015-a-php-class-is-named-with-dots.md)**, status **Proposed**:

1. `.` is the separator, for Phel namespaces and PHP class names alike.
2. A class reference is identified lexically, by an upper-case first segment. No reflection, so a name's meaning never depends on what happened to be loaded.
3. A lower-case-initial namespace (`phpDocumentor\Reflection\…`) is reached by importing it, `(:use phpDocumentor.Reflection.DocBlock)`, which is Phel's `:import`. One of the 99 PSR-4 namespaces installed here needs it.
4. **The leading `\` retires with the separator** at the next major, rather than being promoted to a permanent marker. `\Phel.Lang.Symbol` is deliberately never made to work: I checked, it currently fails with `Undefined constant "Phel"`, and making it work would add a fourth spelling on the way to having one.
5. The `def`-shadows-a-class warning becomes a **refusal** in that same major. That is the guarantee Clojure gives, and it is what makes a bare name safe.

`\` keeps working for all of `1.x`, as the upgrade guide already promises, and its notice period starts at `1.0.0` per ADR 0014. #2827 carries out the removal.

Also updated: the migration guide gains a "What replaces it" section, and the language-surface and divergences pages point at the record.

## 🙋 Why Proposed rather than Accepted

The Clojure-alignment argument is @jasalt's, so the record should carry his review before it is frozen. Nothing in `1.x` depends on accepting it: both spellings work today either way, and the only thing this changes now is that the destination is written down instead of being re-argued.

Worth his eye in particular: point 3, whether an import is the right answer for a lower-case-initial vendor namespace, or whether that case deserves resolution of its own.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,12 +23,14 @@ All notable changes to this project will be documented in this file.
 - `phel eval -h` shows a heredoc example for multi-line evaluation
 - The normative [language-surface spec](docs/spec/language-surface.md) and [Clojure divergences](docs/spec/clojure-divergences.md). A test compares the special-form table against the analyzer's dispatch registry, so the spec cannot drift from the compiler
 - The [stability policy](docs/stability.md), plus migration guides for [0.49 to 1.0](docs/migration/upgrade-0.49-to-1.0.md), [live deprecations](docs/migration/deprecated-surface.md) and [removed core fns](docs/migration/removed-deprecated-core-fns.md)
-- [Architecture decision records](docs/adr/README.md): 14 records covering the decisions that read as oversights and get re-proposed, each naming the test that fails when it is broken
+- [Architecture decision records](docs/adr/README.md): 15 records covering the decisions that read as oversights and get re-proposed, each naming the test that fails when it is broken
 - Compatibility gates that fail the build on drift: snapshots of the public PHP surface (`composer api-surface:update`) and of every public `phel.*` definition and arity (`composer core-api:update`), plus a `:doc` gate on public definitions
 - Every class outside the public PHP surface carries `@internal`, so the split reaches IDEs and static analysis. A test pins both directions
 - CI: macOS is a supported platform (Windows stays declared best-effort), coverage is published and gated at 85%, benchmarks are asserted against the base revision, and mutation testing runs weekly over `src/php/Lang/` and the analyzer (MSI floor 80)
 
 ### Changed
+
+- The spelling a PHP class will have after `\` goes is recorded: dotted or bare, with no leading marker, and the marker retires with the separator rather than becoming permanent ([ADR 0015](docs/adr/0015-a-php-class-is-named-with-dots.md), #2876). Nothing changes in `1.x`, where both spellings work
 
 - A `def` whose name is a loadable PHP class warns. The definition wins from there on, so `(def DateTime "shadow")` used to make `(new DateTime)` fail with `Class "shadow" not found` and nothing said why. Clojure refuses the `def`; Phel warns for now and names `\DateTime` as the spelling that still reaches the class, with the refusal scheduled for the major that drops the leading `\` (#2876)
 
--- a/docs/adr/0015-a-php-class-is-named-with-dots.md
+++ b/docs/adr/0015-a-php-class-is-named-with-dots.md
@@ -1,0 +1,120 @@
+# ADR 0015: A PHP class is named with dots and no leading marker
+
+- **Status**: Proposed. Nothing in `1.x` depends on accepting it, and it should
+  carry @jasalt's review before it does: the Clojure-alignment argument is his
+  ([#2876](https://github.com/phel-lang/phel-lang/issues/2876#issuecomment-5156046021),
+  [#2859](https://github.com/phel-lang/phel-lang/issues/2859#issuecomment-5152721973)),
+  and this record only follows it to its conclusion.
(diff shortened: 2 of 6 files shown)


jasalt:
Comment moved to https://github.com/phel-lang/phel-lang/issues/2876#issuecomment-5159253536

Issue #2941, opened 2026-08-03
Refactor Phel core & stdlib & tests to use less raw PHP interop and lower level spellings


jasalt:
When generating Phel code LLM coding tools often pick up Phel core & stdlib sources as example which in many cases uses "lower level" spellings than what is currently available.

Proposed goal would be to make Phel core a good example utilizing "high level" language features (functions/macros and syntax sugar) where it make sense and keep raw PHP interop at minimum. Changes should also not affect performance negatively.

Some examples:

## Prefer Phel core / stdlib functions over `php/*` where available

The `php/` prefix should be used mainly for calling PHP functions that don't have "Clojure-like" core library function available.

- `php/aget` / `php/aset`
  - Replacement: `aget` / `aset`
  - Current usage: https://github.com/search?q=repo%3Aphel-lang%2Fphel-lang+php%2Faget&type=code
  - Related ADR: https://github.com/phel-lang/phel-lang/blob/b18417d983df7aa02480a1094fe0cc8f21030d49/docs/adr/0007-clojure-style-interop-is-the-source-spelling.md#decision
  - PR https://github.com/phel-lang/phel-lang/pull/2942

- `php/preg_match` and other PHP regex functions
  - Replacement: Use `re-` prefixed Phel regex functions
  - Usages: https://github.com/search?q=repo%3Aphel-lang%2Fphel-lang+php%2Fpreg_match&type=code
  - Not all usages have direct equivalent. Performance considerations with e.g. https://github.com/phel-lang/phel-lang/blob/b18417d983df7aa02480a1094fe0cc8f21030d49/src/phel/string.phel?

- `(php/intval x)`
  - Replacement: `(int x)`
  - Usages: https://github.com/search?q=repo%3Aphel-lang%2Fphel-lang+php%2Fintval&type=code
  - Performance considerations?

- `php/str_replace`
  - Replacement: `phel.string/replace`

- `php/array`
  - Replacement: `php-indexed-array` / `php-associative-array`

## Use reader macros / syntax sugar where available

Not as high value but could be considered also.

- `to-php-array`
  - Replacement: `#php` reader macro `(to-php-array x)` => `#php x`
  - Current usage: https://github.com/search?q=repo%3Aphel-lang%2Fphel-lang%20to-php-array&type=code

jasalt:
[...]

PR #2942, opened 2026-08-03
refactor(stdlib): prefer `aget` / `aset` array ops spelling without leading `php/`

jasalt:
Converts usage of `php/aget` and `php/aset` to Clojure-compatible `aget` and `aset` in `src/phel/` and `tests/phel` where possible.

For now, PHP sources are not touched and docs haven't been confirmed to promote this Clojure-compatible syntax.

Related issue: https://github.com/phel-lang/phel-lang/issues/2941


jasalt:
Broad sed replacement `find ./src/phel -type f -exec sed -i -E 's#php/aget([^-]|$)#aget\1#g' {} +` crossed bootstrap boundaries and made tests fail:                                                          
                                                                   
 - core.phel uses aget before it is defined.                                                                                  
 - (declare aget) only creates a nil placeholder, causing runtime “null is not callable”.                                     
 - core/defs.phel invokes aget during macro bootstrapping.                                                                    
 - core/arrays.phel’s aget implementation became recursively self-calling.                                                    
 - The remaining def-, defn, defstruct, etc. errors were cascading failures because phel.core never loaded.

jasalt:
Added changelog entry

Chemaclass:
Reviewed, and the conversion itself is right: the split you landed on is principled (`core/arrays.phel` keeps `php/aset` where the macro would reference itself during bootstrap, `(declare aget)` covers the reads that compile before `core/arrays` loads), and the docstrings that taught the old spelling were updated with it.

Two things do not survive a measurement, and both turned out to be defects in `aget` / `aset` rather than in this PR:

**1. `aset` evaluated its value expression twice.** The macro spliced the value form into its expansion twice:

```phel
(macroexpand '(aset a 0 (f x)))
;; => (do (php/aset a 0 (f x)) (f x))
```

Harmless for a variable, quadratic-then-exponential for a recursive call, because each level doubles. `json.phel:26` and `transit.phel:238` are exactly that shape (`(aset arr (encode-key k) (encode-value v))`), so on this branch `phel.json/encode` grows exponentially with nesting depth:

| nesting depth | `main` | this branch |
|---|---|---|
| 6 | 0.03 ms | 0.45 ms |
| 10 | 0.03 ms | 9.94 ms |
| 14 | 0.03 ms | 149 ms |
| 16 | 0.03 ms | 712 ms |

**2. `aget` reached every read through `& indices`.** Collecting the rest arguments costs more than the array read it wraps: ~1.6 µs per call against ~0.13 µs for a raw `php/aget`. That is why the standard library still wrote `php/aget` on the hot paths, and converting `str` (`core/strings.phel:*` via `val-to-str`) and `get` on PHP arrays (`core/sequences.phel`) made both 3-4x slower on this branch.

Neither is your bug to carry: `aset` should have bound its value once (`set!` in `core/interop.phel` always has), and `aget` should have declared its single-index arity. Both are fixed in #2944, with the evaluation-count regression tests that were missing.

I cherry-picked your two commits on top of that fix and re-measured: `json/encode` at depth 16 goes back to 0.03 ms and stays flat, `str` and `get` return to `main`'s range, `composer test-core` is 6603/0. **So this PR is good as written** — it only needed the floor under it to be solid.

Merging #2944 first, then this one after a rebase (the only conflict is the `CHANGELOG.md` bullet).

One optional follow-up, not a blocker: with `aget` no longer paying for rest arguments, the sites you deliberately left on `php/aget` inside `core/arrays.phel` (the `typed-array` copy loop, the pair loop) could go too, if you want the file to read consistently. The bootstrap-order ones must obviously stay.

Chemaclass:
#2944 has landed, so I merged `main` into this branch for you. What that touched, so nothing looks unexplained:

- `CHANGELOG.md`: the only real conflict. Your bullet now sits below #2944's `aget` one and carries the `(#2941, #2942)` reference the neighbouring entries all have, plus a sentence naming where the raw spelling deliberately stays (`core/defs.phel`, the eager defs in `core.phel`, `aget`/`aset`'s own definitions) so the split reads as a decision rather than an omission.
- `tests/phel/core/basic-constructors.phel`: git auto-merged your `test-aset-nested` change with #2944's two new evaluation-count tests. I switched those two new tests' assertions from `php/aget` to `aget` so the file does not mix spellings right after your PR unmixed it.
- `src/phel/core/arrays.phel` auto-merged cleanly.

Re-measured on the merged branch: `phel.json/encode` of a 16-deep map is 0.04 ms and flat across depths, `composer test-core` is 6603/0. Merging once CI is green.

One more site to your credit that I had not spotted: `src/phel/core/seq-fns.phel:877,889`, `phel->php`'s recursive PHP-array and associative branches, had the same exponential shape as `json.phel` and `transit.phel`. That one would have been the worst of the three in practice, since `phel->php` is on every interop boundary. Fixed by #2944 along with the rest.

PR #2943, opened 2026-08-03
fix(core): `slurp` validates its path, and `re-find` stops paying for matches it discards

jasalt:
Related to #2941. Picks up @jasalt's branch and answers the two concerns raised on it, both of which turned out to be real.

## 🤔 Background

The original conversion could not land as written: `re-find` cost 50x a raw `php/preg_match` predicate, so `str` on a float and `parse-long` came out ~4x slower, and the `io.phel` site changed behaviour in two directions at once. Rather than revert those sites, this makes the high-level spelling worth using, the same way #2944 did for `aget`/`aset`.

**`slurp` has never validated anything.** `php/preg_match` returns `0` on no match, `0` is truthy in Phel, so `(when-not is-url? ...)` never fired:

```phel
# on main
(slurp "/tmp/nope.txt")   ;; => "Failed to read from: /tmp/nope.txt" — the "File not found" branch is dead
(slurp "/some/dir")       ;; => "" — a directory reads as the empty string
```

Making the guard live is only half a fix. The check recognised `https?|ftp` and nothing else, so every other stream wrapper fell through to `is_file`/`is_dir`, which cannot answer for one:

```phel
# guard live, three-scheme check
(slurp "data://text/plain;base64,aGVsbG8=")  ;; => throws "File not found" — on main this returns "hello"
```

`php://stdin`, `php://memory`, `zlib://`, `phar://` and any host-registered wrapper went the same way. Nothing in the repo slurps a wrapper, which is why CI stayed green on it.

**`re-find` paid for data its callers discarded.** It ran the whole match array through `php->phel` — allocating a persistent structure — then `count`, then `first`, to hand back one string. That is why core still compared integers against `1`, and that comparison is exactly what made `slurp`'s guard fail silently.

## 💡 Goal

Fix `slurp`, and make asking a regex a question cheap enough that core has no reason to reach for `php/preg_match` and re-introduce the same class of bug.

## 🔖 Changes

**`src/phel/core/io.phel`**

- `re-find` reads element zero directly for a group-less pattern, and builds its vector straight from the match array when the pattern has groups. Dropping `php->phel` also stops routing named groups through a hash map, whose iteration order is not the match order; reading the array keeps PHP's order.
- `re-matches` uses `some?` and compares the full match with `php/===`.
- New private `stream-path?`: is the path's scheme one `php/stream_get_wrappers` reports. `slurp` skips the filesystem checks for those and runs them for everything else, so the guards work and wrappers still read. Checking the *shape* alone was not enough: PHP reads a scheme it has no wrapper for as a literal filename, and a relative directory may legally be named `something:`, so `(slurp "schemelike:///")` would have read back as `""`. `path-status-silenced` answers `is_dir`/`is_file` with warnings off, since both warn for a scheme-shaped path with no wrapper behind it.
- The four regex helpers move to the head of the file, since `slurp` calls them and a definition has to precede its caller.

**`src/phel/core.phel`** — unchanged against `main`. The first commit on this branch added `(declare re-find)` for `core/strings.phel`; that file no longer calls a regex, so the declaration is gone again and the net diff here is empty.

**`src/phel/core/strings.phel`** — `float-to-str` drops its regex: `php/strpbrk` answers "does this already read as a float" in one scan, rather than compiling a pattern to look for three characters.

**`src/phel/core/parsing.phel`** — `parse-long` asks `re-matches`. The `^...$` anchors are gone with it, since `re-matches` *is* the anchor.

**`src/phel/core/interop.phel`** — the two `(= 1 (php/preg_match ...))` predicates in `static-property-place?` read `(some? (re-find ...))`. This runs at macro-expansion time, for every `set!`.

**Tests** — `slurp` gains coverage for the missing-file message, a directory, a `data://` read and an unregistered scheme (`tests/phel/core/file-operation.phel:9` only asserted the exception *type*, which is how the wrong message survived). Plus `re-find` on a zero-length match and on an optional group that did not participate, and `parse-long` on `+7`, `+-5`, `1e5`, `12a`, and `str` on a float in exponent form.

## ✅ Verification

200k calls each, warmed, three interleaved runs against `main`, minimum of each:

| | `main` | this branch |
|---|---|---|
| `(re-find "/[.eE]/" "1.5")` | 612 ms | **39 ms** |
| `(re-find "/(\d+)-(\d+)/" "12-34")` | 2731 ms | **1874 ms** |
| `(str 1.5)` | 194 ms | 193 ms |
| `(parse-long "12345")` | 175 ms | **56 ms** |

Behaviour diffed against `main` over float rendering (including `NaN`, `±Infinity`, exponent form), ten `parse-long` inputs, `re-find`/`re-matches`/`re-seq` across no-match, zero-length, group, optional-group and named-group cases, and `set!` place resolution: identical everywhere except the two intended `slurp` changes.

`composer test-core` 6618 passed / 0 failed, `--testsuite=unit` 4491 OK, `--testsuite=integration` 1175 OK, `phel format --dry-run` clean, `phel lint src/phel` 0 errors.

## Left alone deliberately

`re-find` with named groups returns `[[0 "123"] ["x" "123"] [1 "123"]]` rather than the positional vector its docstring promises, because PHP fills the match array under both the name and the number. This branch preserves that shape rather than changing a public contract inside a fix, and pins it in a test now that reading the match array in match order makes it deterministic. Worth its own issue.

## Review

@jasalt's two flagged concerns both held up under measurement, and an adversarial pass over this diff found one more: the `stream-path?` false positive described above, plus the CHANGELOG wording, both fixed in `866e4abc5`. Everything else it probed — `re-find` against the old `php->phel` version across named, mixed, `\K`, non-participating and zero-length matches, `parse-long` against unicode digits, embedded newlines, overflow and unstripped whitespace, `strpbrk` against nine float renderings, and every `php/===` substitution — agreed with `main` byte for byte.


jasalt:
I didn't do more tailored benchmarking if changes affected usage in hot paths but here's benchmark results from the included benchmark suite: 

```
Performance comparison against origin/main (b18417d98) completed.                                                                                                                                    
                                                                                                                                                                                                    
 ┌───────────────────────┬─────────────┬────────────┬───────┐                                                                                                                                         
 │ Benchmark             │ origin/main │ Commit     │ Delta │                                                                                                                                         
 ├───────────────────────┼─────────────┼────────────┼───────┤                                                                                                                                         
 │ LoadChain warm        │ 46.102 ms   │ 46.256 ms  │ +0.3% │                                                                                                                                         
 ├───────────────────────┼─────────────┼────────────┼───────┤                                                                                                                                         
 │ LoadChain fresh       │ 46.164 ms   │ 46.812 ms  │ +1.4% │                                                                                                                                         
 ├───────────────────────┼─────────────┼────────────┼───────┤                                                                                                                                         
 │ Core compilation      │ 81.030 ms   │ 83.577 ms  │ +3.1% │                                                                                                                                         
 ├───────────────────────┼─────────────┼────────────┼───────┤                                                                                                                                         
 │ REPL lazy boot        │ 79.098 ms   │ 79.938 ms  │ +1.1% │                                                                                                                                         
 ├───────────────────────┼─────────────┼────────────┼───────┤                                                                                                                                         
 │ REPL eager boot       │ 111.586 ms  │ 112.609 ms │ +0.9% │                                                                                                                                         
 ├───────────────────────┼─────────────┼────────────┼───────┤                                                                                                                                         
 │ Phel run (20 samples) │ 50.371 ms   │ 49.998 ms  │ −0.7% │                                                                                                                                         
 └───────────────────────┴─────────────┴────────────┴───────┘                                                                                                                                         
                                                                                                                                                                                                      
 No significant regression detected; all changes are within the ±5% noise threshold.
```
Measured with PHPBench 1.7 using the existing startup/compile benchmarks:                                                                                                                            
                                                                                                                                                                                                      
 - Build/LoadChainBench                                                                                                                                                                               
 - Phel/PhelBench                                                                                                                                                                                     
 - Run/ReplBootBench                                                                                                                                                                                  
                                                                                                                                                                                                      
 Method:                                                                                                                                                                                              
                                                                                                                                                                                                      
 - Ran origin/main in a separate Git worktree.                                                                                                                                                        
 - Ran the current commit in the working tree.                                                                                                                                                        
 - Cleared .phel/cache before each run.                                                                                                                                                               
 - PHP 8.5.8, Xdebug disabled, OPcache disabled, 1 GB memory limit.                                                                                                                                   
 - PHPBench remote executor, --iterations=10 --revs=10.                                                                                                                                               
 - Compared PHPBench’s reported Mo timings manually.                                                                                                                                                  
 - Re-ran PhelBench::bench_phel_run with 20 iterations and 1 revolution for confirmation.

jasalt:
Should think if there was cleaner way to resolve the load order load order issue (Consideration A) if this is worth pursuing further.

jasalt:
Added changelog entry

Chemaclass:
Measured this one rather than eyeballed it, because your two flagged concerns both turn out to matter, and there is a third thing in here that nobody asked for and that we want.

**The `io.phel` change is a bug fix, not a refactor.** `php/preg_match` returns `0` on no match, and `0` is truthy in Phel, so `slurp`'s `(when-not is-url? ...)` guard never fired. Both validations under it have been dead:

```phel
# on main
(slurp "/tmp/nope.txt")  ;; => InvalidArgumentException "Failed to read from: /tmp/nope.txt"
(slurp "/tmp/somedir")   ;; => "" — no throw at all
```

```phel
# on this branch
(slurp "/tmp/nope.txt")  ;; => InvalidArgumentException "File not found: /tmp/nope.txt"
(slurp "/tmp/somedir")   ;; => InvalidArgumentException "Cannot read directory: /tmp/somedir"
```

That is worth landing on its own, and it deserves what a fix gets: a `### Fixed` CHANGELOG bullet naming the changed messages (a directory silently reading as `""` is the user-visible half), and tests. `tests/phel/core/file-operation.phel:7` currently asserts only *that* a missing file throws, which is why the wrong message passed for so long. I swept the rest of `src/phel/` for the same trap: `ai.phel:402`, `http.phel:46` and `http.phel:88` use `one?`, `string.phel:161` uses `not= 0`. `slurp` was the only broken one.

**The two runtime sites regress about 4x.** `re-find` allocates a match array, runs `preg_match`, then converts the result through `php->phel`, `count` and `first` — all discarded when the caller only wants a yes/no. 200k calls, five interleaved runs per branch, minimum of each:

| | `main` | this branch |
|---|---|---|
| `(str 1.5)` — `format-float` in `core/strings.phel` | 182 ms | 802 ms |
| `(parse-long "12345")` — `core/parsing.phel` | 187 ms | 795 ms |

Every run on this branch sat above every run on `main`, so this is not noise. The repo's benchmark suite is startup-and-compile shaped (`LoadChainBench`, `PhelBench`, `ReplBootBench`), which is why your table came back clean at +3.1%: none of those benchmarks call `str` on a float in a loop. Your instinct in note (B) was right.

So the split I would make:

1. **Keep** the `io.phel` change, as a fix, with tests and a `### Fixed` entry.
2. **Drop** `core/strings.phel` and `core/parsing.phel` for now. They are the hot paths, and the honest answer is that #2941's advice cannot be followed there until a predicate exists that does not pay for the match data. `interop.phel` is macro-expansion-time only, so it is borderline: it costs compile time, which is where your +3.1% came from.
3. The thing that would make all of these convertible is a boolean predicate in core, something like `(re-find? re s)` compiling down to `(php/=== 1 (php/preg_match re s))`. That is new public surface and a call for @Chemaclass, so worth raising on #2941 rather than deciding inside this PR.

Two smaller notes:

- `(not (nil? x))` is `(some? x)`, which already exists at `src/phel/core/booleans.phel:286`. Six sites here. Using the lower-level spelling in a PR about preferring higher-level spellings is the kind of thing #2941 is trying to fix.
- `(declare re-find)` does work — I confirmed the call site late-binds, so the real `defn` in `core/io.phel` is what runs. But the underlying oddity is that `re-find` and `re-matches` live in `core/io.phel` at all, so a string function has to reach into the IO file. Loading `core/io` earlier only swaps which file needs the forward declaration; giving the regex functions their own home ahead of `core/strings` removes the `declare` entirely. Also a follow-up, not this PR.

Related: #2942 uncovered two comparable defects in `aget`/`aset`, fixed in #2944. Same pattern as here — the low-level spelling survived in core because the high-level one was not yet a free replacement.

Chemaclass:
Correction to my note above, and it changes the recommendation rather than adding to it: **making that guard live is not purely a fix. It also breaks reads that work today.**

The URL check only recognises `https?|ftp`, so once `when-not is-url?` actually fires, every other stream wrapper falls through to `is-file`/`is-dir`, which know nothing about non-filesystem streams, and gets rejected before `file_get_contents` is ever reached. Applied to a clean `main` checkout:

```phel
# main
(slurp "data://text/plain;base64,aGVsbG8=")  ;; => "hello"

# with this change
(slurp "data://text/plain;base64,aGVsbG8=")  ;; => InvalidArgumentException "File not found: data://text/plain;base64,aGVsbG8="
```

`php://stdin`, `php://memory`, `php://temp`, `zlib://`, `phar://` and any custom registered wrapper go the same way. Nothing inside the repo slurps a stream wrapper (`cli.phel:274` and `http.phel:270` reach for `php/fopen` and `php/file_get_contents` directly), which is why CI is green, but `(slurp "php://stdin")` is an obvious thing for user code to do.

So the `io.phel` site needs a decision, not just tests:

- **The guard should recognise more than three schemes.** Anything with a `scheme://` prefix, or better `(php/in_array scheme (php/stream_get_wrappers))`, should skip the filesystem validation. That keeps the two real fixes (a missing file naming itself, a directory throwing rather than reading as `""`) without breaking wrappers.
- Then the tests worth adding are: a directory, a missing file asserting the *message*, and a `data://` read. `tests/phel/core/file-operation.phel:7` asserts only the exception type, which is how the wrong message survived this long.

Everything else in my earlier comment stands: the 4x runtime cost on `core/strings.phel` and `core/parsing.phel`, `some?` over `(not (nil? ...))`, and the regex functions wanting a home outside `core/io.phel`. On that last point, one correction of emphasis: loading `core/io` before `core/strings` is not the cheaper trade, because `core/io` calls `str` twelve times, so the forward declaration would just move to the busier side. A pure `core/regex.phel` loaded ahead of both is the thing that removes the `declare` outright.

#2944 has landed, so the `aget`/`aset` half of #2941 is unblocked.

PR #2953, opened 2026-08-08
docs: fix invalid :closed? predicate in schema validation docs

jasalt:
The closed-map option is **`:closed`** (as in `metosin/malli`), not `:closed?`. The validator reads `(get (schema-options schema) :closed)` (`schema/validator.phel:258`, `explainer.phel:107`). So `[:map {:closed? true} ...]` from the doc would silently be treated as **open**. Use `{:closed true}`.

REPL example:
```clojure
$ ./bin/phel
Welcome to the Phel Repl (v0.49.0-beta#1fe571c)

(ns schema-demo
  (:require phel.schema :as s))

;; Incorrectly documented keyword
(s/validate [:map {:closed? true} [:id :int]] {:id 1 :extra 2})
;; => true

;; Correct
(s/validate [:map {:closed true} [:id :int]] {:id 1 :extra 2})
;; => false
```

### DIFF
--- a/resources/agents/tasks/validate-with-schema.md
+++ b/resources/agents/tasks/validate-with-schema.md
@@ -9,7 +9,7 @@
   (:require phel.schema :as s))
 
 (def User
-  [:map {:closed? true}
+  [:map {:closed true}
    [:id    :int]
    [:email [:re #"^[^@]+@[^@]+$"]]
    [:age   [:maybe :int]]])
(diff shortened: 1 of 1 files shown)


Chemaclass (approved):
Verified against the source and at runtime: `validator.phel:258` and `explainer.phel:107` both read `(get (schema-options schema) :closed)`, so `{:closed? true}` was silently ignored and the map stayed open.

```
closed? key: true
closed  key: false
```

Both occurrences in the file are fixed, and `:closed?` appears nowhere else in the repo. Matches the spelling already used in `src/phel/schema.phel:50` and the schema tests. Thanks!

Issue #2973, opened 2026-08-08
perf: a variadic parameter costs 8-12x a fixed one, per call


Chemaclass:
## The measurement

A `&` parameter costs 8 to 12 times a fixed one, per call, before the body does anything. 300k iterations, net of an empty-closure floor:

| signature | per call | vs fixed |
|---|---|---|
| `[a b c]` | 0.095µs | 1x |
| `[a b & more]` | 0.80µs | **8.4x** |
| `[& args]` | 0.99µs | **10.4x** |
| `[a & [b]]` | 1.18µs | **12.4x** |

The destructured form `& [opt]` is the worst, and it is the one `get`, `get-in` and `nth` reached for to express an optional argument.

## Why this is the systemic answer

It accounts for the results already measured across #2960, #2962, #2964 and #2965. In every case the split between "reorder the branches" and "split the arities" put the larger half on the arity:

| | reorder only | plus fixed arities |
|---|---|---|
| `get` | -27% | **-59%** |

It also explains why phel functions read as slow next to hand-written PHP. `$arr['a'] ?? null` is 1.5µs per 200k; `(get persistent-map k)` was 116µs for the same count, and roughly a fifth of that gap is the persistent map being a different data structure. The rest was call overhead.

## What this is not

Two things that look like the same problem and are not:

- **Persistent collections are not free and should not be.** `$pmap->find(k)` costs 23.8µs against `$arr[k]`'s 1.5µs. That is structural sharing, not overhead, and no arity change touches it.
- **Specialised call sites already pay nothing.** `(str "x" "y")`, `(php-indexed-array 1 2 3)` and `(not= 1 2)` all measure at the floor, because the emitter lowers them when the analyser has types. The runtime function is what an *untagged* call site reaches, and that is what these issues are about.

## Scope of the opportunity

Remaining `phel.core` functions with a variadic parameter, by head-position uses in `src/phel/`:

| uses | fn | | uses | fn |
|---|---|---|---|---|
| 315 | `str` | | 55 | `reduce` |
| 71 | `map` | | 53 | `atom` |
| 68 | `into` | | 28 | `not=` |
| 64 | `hash-set` | | 26 | `concat` |
| 63 | `swap!` | | 22 | `filter` |
| 58 | `println` | | 20 | `take` |

[...]

PR #2983, opened 2026-08-08
perf(core): give the comparison operators fixed arities and a bool return

### LINKED ISSUE #2979: perf(core): comparison operators pay 8-14x on untagged operands
## Measured

Untagged operands, 200k iterations net of floor, against `php/===` at 33.5ms:

| | phel | ratio |
|---|---|---|
| `(not= x y)` | 475.0ms | **14.2x** |
| `(identical? x y)` | 259.7ms | 7.8x |
| `(= x y)` | 258.9ms | 7.7x |

## Cause

The same `case`-over-a-rest-argument shape as the arithmetic operators:

```phel
(defn = [a & more]
  (case (count more)
    0 true
    1 (equals1 a (first more))
    (if (equals1 a (first more)) (recur (first more) (next more)) false)))
```

The two-argument call, which is nearly every call, allocates `more`, calls core `count` on it, runs a `case`, and calls `first` to recover the value the caller already passed. `not=` costs more again because it delegates through `apply`.

`<`, `<=`, `>`, `>=` share the shape.

## Not the whole picture

`(not= 1 2)` on literals measures **at the floor**: the emitter lowers it when the analyser has types. This is the untagged path, which is what a comparison of two function parameters takes.

## Scope

Fixed arities for the two-argument case on `=`, `not=`, `identical?`, `<`, `<=`, `>`, `>=`, keeping the variadic tail for chained comparisons.

Constraints:

- `=` is value equality and `identical?` is not; neither should start delegating to the other.
- The chained arities must keep short-circuiting: `(< 1 2 0)` is false without evaluating past the failure, and `(= a b c)` stops at the first mismatch.
- `(=)` and `(= x)` are `true`. Keep them.
- `not=` should stop routing through `apply` for the two-argument case.

Rationale: #2973

Chemaclass:
## 🤔 Background

Same `case`-over-a-rest-argument shape as the arithmetic operators in #2982, across `=`, `not=`, `identical?`, `<`, `<=`, `>` and `>=`.

## 🔖 Changes

Untagged operands, 200k iterations net of floor:

| | before | after | |
|---|---|---|---|
| `(not= x y)` | 475.0ms | **95.0ms** | 5.0x |
| `(identical? x y)` | 259.7ms | **69.9ms** | 3.7x |
| `(= x y)` | 258.9ms | **73.2ms** | 3.5x |
| `(< x y)` | 592.1ms | **391.0ms** | 1.5x |
| `php/===` raw | | 34.2ms | |

`<` lags the others because `lt2` itself costs about 1.75µs, which an arity split does not touch. Worth its own look; not this change.

Chained arities keep comparing each argument with the next and stop at the first failure, now through an explicit `loop` rather than `recur` into a variadic.

## ⚠️ The part that needs your judgement

`^bool` is declared on every arity. **Without it the emitted code lost the `: bool` return** it used to infer on inlined comparisons, which the `Inline/` fixtures caught as a regression.

Declaring it turns out to *strengthen* inference rather than merely restore it:

- a `defn` whose body is `(= x 5)` now emits `: bool`
- `(if (= x nil) 0 1)` now emits `: int`

Both are correct. The second contradicted a fixture named **`defn-infer-phel-core-eq-nil-stays-untagged`**. Its expectation was conservative rather than required, so I renamed it to `defn-infer-phel-core-eq-nil-infers-int` rather than regenerate it under a name that now asserts the opposite of what it checks. **If that fixture was pinning a deliberate decision to stay conservative, this is the change to push back on.**

`tests/phel/core/cmp-dispatch.phel` covers every arity of all seven, chained comparisons including ones broken in the middle, and mixed numeric-tower operands.

Rationale for the series: #2973

Closes #2979

## Update: a shadowed nil guard, found before merge

A measurement sweep flagged that `<` had **not** followed `=` here: 391.0ms against `=`'s 73.2ms. The arity split was not the problem.

`booleans.phel` defines a fixed-arity `assert-non-nil [a b]`, and its docblock says it is local *"because `booleans` loads before the arithmetic module"*. Load order defeats exactly that: `core/math` defines a **variadic** `assert-non-nil [& xs]` in the same `phel.core` namespace and loads **after**, so the short name resolved to the variadic one and every ordered comparison allocated a rest argument. The local guard was dead code.

Renaming it `assert-non-nil-pair` is the whole fix:

```
(< x y)   391.0ms -> 84.8ms
(>= x y)          -> 86.6ms
php/< raw            19.5ms
```

**7.0x** against the 592.1ms measured before the arity split, and the ratio against raw PHP falls from about 30x to 4.3x. `<` now tracks `=` instead of lagging it.

nil is still rejected, and that is tested. math's variadic `assert-non-nil` stays, because the bit operators use it.

This is why the PR was held rather than merged on green.


Chemaclass:
Rebased onto current main (this branch was 14 behind, which is why the PHAR smoke tests had gone
red) and settled the fixture question that was holding it.

## The fixture question, answered

I had renamed `defn-infer-phel-core-eq-nil-stays-untagged` to `-infers-int`, and flagged that I
was not sure whether the original name pinned a deliberate decision. It did, and it is a
different decision from the one this PR touches.

The fixture arrived with #2080, whose commit message says:

> a body with no primitive literal anywhere (e.g. `(+ a b)`) stays untagged and preserves
> BigInt / Ratio dispatch

The untagged thing is the **parameter**. `(= x nil)` has no numeric literal sibling to
disambiguate, so `$x` must not be inferred as `int`, or `BigInt` and `Ratio` arguments would
stop dispatching.

That decision is untouched here. The signature this branch emits is:

```php
public function __invoke($x): int {
```

`$x` still untyped; only the return type added, which is a strictly different axis. So the
fixture is now named `defn-infer-phel-core-eq-nil-param-stays-untagged`, which keeps saying what
it guards, and the `: int` in the fixture body speaks for itself.

## New evidence for merging this

While auditing `apply` I decomposed why `max` and `min` are slow, and it lands here. Measured in
one process against an empty-closure floor, floor subtracted:

```
(> a b)          2.720 us          (max a b)  2.959 us
numeric-nan?     0.045 us  x2
```

`max`'s pair arity is `(cond (numeric-nan? x) … (numeric-nan? y) … (> x y) x :else y)`, so
**92% of what `max` costs is the single `>` call**. `max` is not slow; `>` is, and `max` inherits
it.

With this branch rebased:

| call | main | this branch | speedup |
|---|---|---|---|
| `(> a b)` | 2.720 us | 0.333 us | 8.2x |
| `(< a b)` | 2.709 us | 0.328 us | 8.3x |
| `(>= a b)` | 2.713 us | 0.320 us | 8.5x |
| `(= a b)` | 1.051 us | 0.133 us | 7.9x |
| `(max a b)` | 2.959 us | 0.550 us | **5.4x, without touching max** |
| `(min a b)` | 2.977 us | 0.550 us | **5.4x, without touching min** |

For scale, `(+ a b)` costs 0.147 us, so the comparison operators were running roughly 18x an
arithmetic operator of the same shape. They are now in the same range.

`composer test` green on the rebase: 6968 Phel tests, 4557 unit, 1187 integration, quality clean.

PR #3011, opened 2026-08-09
perf(core): give the five core constructors fixed arities

### LINKED ISSUE #3010: perf(core): the five core constructors are bare variadics, costing 3x on every call
## Summary

`list`, `vector`, `queue`, `hash-map` and `array-map` are the five most fundamental
constructors in the language, and all five are spelled as a bare variadic in
`src/phel/core.phel`:

```phel
(def list
  {:doc "..."}
  (fn [& xs]
    (Phel/list (apply php/array xs))))
```

`vector`, `queue`, `hash-map` and `array-map` are identical modulo the constructor called.
They sit under the comment "Basic methods for quasiquote and destructure", which is exactly
why this matters: they are on the expansion path, not just in user code.

Per #2973 a `&` parameter costs 8-12x a fixed one per call before the body runs. These are
the hottest functions in the language that still pay it.

## Measured

Both spellings compiled in one process and timed against an empty-closure floor taken in
that same process, 200k revs each, floor subtracted:

| args | variadic (today) | fixed arities | speedup |
|---|---|---|---|
| 0 | 0.918 us | 0.297 us | 3.1x |
| 1 | 1.698 us | 0.542 us | 3.1x |
| 3 | 2.453 us | 0.947 us | 2.6x |
| 5 (falls through to the variadic tail) | 3.155 us | 2.724 us | 1.2x |

For scale, in the same process a raw PHP closure taking one argument costs 0.023 us and
`phel.core/identity` costs 0.024 us. The dispatch mechanism itself is free; a bare `(list 1)`
costs 74x a PHP closure call purely because of how it is declared.

The prototype measured was:

```phel
(fn
  ([] (Phel/list (php/array)))
  ([a] (Phel/list (php/array a)))
  ([a b] (Phel/list (php/array a b)))
  ([a b c] (Phel/list (php/array a b c)))
  ([a b c & more] (Phel/list (apply php/array a b c more))))
```

It compiles and returns the correct `PersistentList`.

## Why this one is free of the `max` trade-off

Declaring fixed arities on `max` (#2992) moved a no-argument call from a runtime
`ArgumentCountError` to a compile-time PHEL002, which was a real if minor BC change. That
does not apply here, because the variadic tail keeps accepting every argument count. Odd
argument counts to `hash-map` still reach `Phel/map` and still throw there, exactly as today.

## Scope

One PR covering all five, because it is the same edit five times in the same block of
`core.phel` and splitting it would be noise. Points worth settling while implementing:

- `hash-map` and `array-map` take key/value pairs, so their fixed arities should be even
  (0, 2, 4, 6) with the variadic tail after.
- Check whether the emitter already specialises any of these at the call site before adding
  arities, so the two mechanisms do not overlap confusingly.
- These are `(def name (fn ...))` rather than `defn` because they are defined before `defn`
  exists. Multi-arity `fn` works there (the prototype above compiled in that form), but this
  is bootstrap-early code: a mistake here breaks the entire stdlib load rather than one
  function, so `composer test` is the gate, not a focused test run. That failure mode has
  bitten before, in #3004.

## Acceptance

- all five constructors keep their current behaviour for every argument count, including the
  odd-count `hash-map` error
- a benchmark subject covers at least `list` and `hash-map` at small arity, so the win cannot
  be silently reverted (this is the gap #3005 exists to close)
- `composer test` green, including `test-core`

Chemaclass:
## 🤔 Background

`list`, `vector`, `queue`, `hash-map` and `array-map` are the five most
fundamental constructors in the language, and all five were bare variadics:

```phel
(def list
  {:doc "..."}
  (fn [& xs]
    (Phel/list (apply php/array xs))))
```

Per #2973 a `&` parameter costs several times a fixed one per call before the
body runs.

**This PR was opened before #3016, and #3016 changed what it is for.** That
change lowers a literal `(list 1 2 3)` straight to `\Phel::list([1, 2, 3])` at
the call site, so a literal call no longer reaches these functions at all. The
original benchmark table here measured literal calls and is now obsolete: those
numbers belong to #3016.

What the lowering deliberately declines is every call where the constructor is
not in call-head position, because there the argument count is not known at
compile time:

- a local alias, `(let [f list] (f 1 2 3))`
- `apply`
- passing the constructor as a value, `(map list xs)`

Those calls still pay the variadic in full, and they are what this PR is for.

## 💡 Goal

Remove the variadic tax from the calls the literal lowering cannot reach,
without changing what any call does.

## 🔖 Changes

- Declare real arities for the small argument counts on all five, keeping a
  variadic tail.
- The map constructors additionally declare their **odd** arities. Those exist
  only so that an odd argument count keeps reaching `Phel/map` and failing
  there with the message it always had. With only the even arities declared, an
  odd count matches no arity and fails earlier, on dispatch, with a less useful
  exception.
- New `tests/phel/core/constructor-runtime-arities.phel`. The existing
  `constructor-dispatch.phel` pins the #3016 lowering, so every call in it is
  lowered and none of them dispatches on an arity: it cannot see this change at
  all. The new file routes every call through a `let`-bound alias, which is a
  `LocalVarNode` that the lowering declines, and covers each declared arity, one
  past the tail, `apply`, and use as a value.
- Corrected the comment above the constructors. It credited the quasiquote
  expansion path, but quasiquote expands to literal calls, which #3016 now
  lowers, so that is no longer where the cost is.

### Measured

On top of #3016, so these are the calls that remain. Empty-closure floor taken
in the same process and subtracted, 200k revs per row, arguments passed through
an alias so nothing can be folded away:

| call | before | after | speedup |
|---|---|---|---|
| `(f 1)` where `f` is `list` | 1.673 us | 0.489 us | 3.42x |
| `(f 1 2 3)` where `f` is `list` | 2.445 us | 0.887 us | 2.76x |
| `(f 1 2 3)` where `f` is `vector` | 2.314 us | 0.805 us | 2.87x |
| `(f 1 2)` where `f` is `queue` | 2.685 us | 1.254 us | 2.14x |
| `(f :a 1)` where `f` is `hash-map` | 2.463 us | 0.976 us | 2.52x |
| `(f :a 1 :b 2)` where `f` is `hash-map` | 3.297 us | 1.431 us | 2.30x |
| `(apply list [1 2 3])` | 3.363 us | 1.789 us | 1.88x |
| `(f 1 2 3 4 5)` where `f` is `list`, the tail | 3.157 us | 2.680 us | 1.18x |
| `(map list xs)` over 10, realized | 37.859 us | 24.261 us | 1.56x |
| `(list 1 2 3)`, literal | 0.987 us | 0.973 us | unchanged, #3016 owns it |

The literal row is the control: it is already lowered, so it should not move,
and it does not.

### No BC cost

Unlike the arity split on `max` (#2992), nothing moves from a runtime error to
a compile-time one, because the variadic tail still accepts every argument
count. The odd-count errors were diffed against `main` and are byte-identical:

```
RuntimeException: An even number of elements must be provided to build a map, got 1
```

Deleting the odd arities turns four assertions in the new test red with
`InvalidArgumentException: No matching function arity`, so that guard bites.

`composer core-api:update` reports the API snapshot unchanged, and
`composer test-all` passes with 7,156 core tests.

### Depends on #3029

Growing `core.phel` by 1,945 bytes tripped four compiler bench subjects that
read `src/phel/core.phel` as their input, at +31%, +31%, +27% and +35% against
a 25% tolerance. Nothing got slower per byte; the input got bigger on one side
of the comparison only. #3029 freezes that corpus under `tests/php/Benchmark`,
which is the side the job already pins to the pull request for both runs. This
branch is built on it, and with it in place those subjects move by -0.8%,
+2.9%, -2.7% and +1.5%.

### Known limitation, tracked separately

`phel doc list` still renders `(list & xs)` rather than the five arities, while
`phel doc hash-set` renders all of that one's. That string is identical before
and after this PR, and it still describes the accepted call shapes correctly,
so nothing regresses; it is simply less complete than reality now. The cause is
pre-existing and already filed as #3012: a multi-arity `defn` reports every
arity, a `def` over an `fn` reports one. These five cannot be `defn`, because
they are defined before `defn` exists.

Closes #3010

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@ Runtime core:
 - Give the closures returned by `partial`, `fnil` and `comp` real arities: 30x, 50x and 8x faster to call (#3021)
 - Guard the twelve nil-rejecting arithmetic functions with the fixed-arity macros instead of the variadic helper: `round` and `floor` 25x, `quot`, `rem` and `mod` 14 to 17x, `even?` and `odd?` 6 to 7x (#3018)
 - Compile a literal `(list ...)`, `(vector ...)`, `(queue ...)`, `(hash-map ...)` or `(array-map ...)` straight to its `Phel` factory: 2.2 to 3.6 times faster (#3014)
+- Give `list`, `vector`, `queue`, `hash-map` and `array-map` fixed arities, so the calls the literal lowering cannot reach get faster too: 2.1 to 3.4 times through an alias or `apply`, 1.5 times passed as a value to `map` (#3010)
 - Inline PHP array constructors and use `php-indexed-array` across core and standard libraries (#2941)
 - Speed up `re-find`, `re-matches` and `parse-long` by avoiding full match-map conversion (#2941 #2943)
 - Give `aget` a fixed single-index arity and require an index, matching Clojure (#2941)
--- a/src/phel/core.phel
+++ b/src/phel/core.phel
@@ -28,36 +28,73 @@
 ;; Basic methods for quasiquote and destructure
 ;; --------------------------------------------
 
+;; Each constructor declares real arities for the small argument counts it is
+;; actually called with, and keeps a variadic tail for the rest. A `&`
+;; parameter costs several times a fixed one per call before the body runs
+;; (#2973), and these five sit on the quasiquote and destructuring expansion
+;; paths, so they pay it more often than anything else in the language.
+;; The tail still accepts every argument count, so nothing that used to be a
+;; runtime error becomes a compile-time one, and an odd argument count still
+;; reaches `Phel/map` and still throws there (#3010).
+
 (def list
   {:doc "Creates a new list. If no argument is provided, an empty list is created. Shortcut: '()"
    :example "(list 1 2 3) ; => '(1 2 3)"}
-  (fn [& xs]
-    (Phel/list (apply php/array xs))))
+  (fn
+    ([] (Phel/list (php/array)))
+    ([a] (Phel/list (php/array a)))
+    ([a b] (Phel/list (php/array a b)))
+    ([a b c] (Phel/list (php/array a b c)))
+    ([a b c & more] (Phel/list (apply php/array a b c more)))))
 
 (def vector
   {:doc "Creates a new vector. If no argument is provided, an empty vector is created. Shortcut: []"
    :example "(vector 1 2 3) ; => [1 2 3]"}
-  (fn [& xs]
-    (Phel/vector (apply php/array xs))))
+  (fn
+    ([] (Phel/vector (php/array)))
+    ([a] (Phel/vector (php/array a)))
+    ([a b] (Phel/vector (php/array a b)))
+    ([a b c] (Phel/vector (php/array a b c)))
+    ([a b c & more] (Phel/vector (apply php/array a b c more)))))
 
 (def queue
   {:doc "Creates a persistent FIFO queue. With no arguments returns an empty queue; with arguments returns a queue with the values pushed in order so that the first argument is at the front."
    :example "(queue 1 2 3) ; => <-(1 2 3)-<"}
-  (fn [& xs]
-    (Phel/queue (apply php/array xs))))
+  (fn
+    ([] (Phel/queue (php/array)))
+    ([a] (Phel/queue (php/array a)))
+    ([a b] (Phel/queue (php/array a b)))
+    ([a b c] (Phel/queue (php/array a b c)))
+    ([a b c & more] (Phel/queue (apply php/array a b c more)))))
 
+;; The map constructors take key/value pairs, so the arities that matter are
+;; the even ones: 2 and 4 cover the overwhelming majority of literal maps. The
+;; odd arities in between are declared anyway, and deliberately: without them
+;; an odd argument count would match no arity at all and fail on dispatch,
+;; where today it reaches `Phel/map` and throws there. Keeping them routes
+;; every count to the same constructor and leaves that error exactly as it was.
 (def hash-map
   {:doc "Creates a new hash map. If no argument is provided, an empty hash map is created. The number of parameters must be even. Shortcut: {}"
    :example "(hash-map :a 1 :b 2) ; => {:a 1, :b 2}"}
-  (fn [& xs]
-    (Phel/map (apply php/array xs))))
+  (fn
+    ([] (Phel/map (php/array)))
+    ([k] (Phel/map (php/array k)))
+    ([k v] (Phel/map (php/array k v)))
+    ([k v k2] (Phel/map (php/array k v k2)))
+    ([k v k2 v2] (Phel/map (php/array k v k2 v2)))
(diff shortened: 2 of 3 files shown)


Chemaclass:
CI is red on `Bench tests`, and the failure is real rather than noisy. Four variants trip the
+25% gate, all of them compile-time benchmarks whose input is `src/phel/core.phel`, which is
the file this PR grows:

| subject | change |
|---|---|
| `LexerBench::bench_lex_core` | +31.16% |
| `ParserBench::bench_parse_core` | +31.17% |
| `ReaderResultCacheBench::bench_cold_front_half` | +26.58% |
| `ReaderResultCacheBench::bench_warm_front_half` | +35.17% |

Reproduced locally, so it is not a one-off CI sample:

```
main     LexerBench::bench_lex_core   1.096  1.104  1.110  1.088  1.070 ms
this PR  LexerBench::bench_lex_core   1.435  1.436  1.429  1.418        ms
```

`core.phel` goes from 8750 to 10695 bytes, +22.2%, and the added lines are dense in forms
rather than prose, which is why the token work lands slightly above the byte growth.

So this is not a slowdown in the lexer or the parser. It is the cost of compiling a bigger
`core.phel`, and the gate is behaving correctly by flagging it.

**The trade.** Roughly +1.7 ms once per cold compile of `core.phel`, against 1.4 to 1.5 us
saved on every constructor call. Break-even is around 1100 constructor calls, which any real
program passes immediately. The cold compile is also the path that the PHAR's precompiled
stdlib and the compiled-code cache exist to avoid, so most runs never pay it.

That said, the project has spent real effort getting cold stdlib compile down, so +22% of
`core.phel` is not something to wave through on my own judgement.

**There may be a strictly better design.** The same win could come from lowering a
known-arity call site in the emitter, `(list a b c)` straight to `Phel::list([a, b, c])`,
alongside the existing specialisation families. That would add nothing to `core.phel`, cost
nothing at compile time, and be faster still, since it removes the call rather than making it
cheaper. It would not cover `apply` or higher-order uses, which would keep today's variadic
and stay as slow as they are now.

Holding this PR rather than merging or tuning it until the gate goes quiet.

Chemaclass:
Parked as draft. Taking the emitter route first: lowering a known-arity `(list a b c)` call site straight to `\Phel::list([...])` gets the same win with no `core.phel` growth and no compile cost, so it dominates this approach wherever it applies. This PR stays open because it is the fallback if that turns out not to cover enough call sites, and because it is the only one of the two that helps `apply` and higher-order uses.

Chemaclass:
The emitter route landed as #3016, and it did not fall short: 2.2x to 3.6x on direct calls, no `core.phel` growth, and the bench gate green. It also beats this PR where this PR was weakest, the counts past the last fixed arity: 2.67x against 1.16x for `(list x 2 3 4 5)`.

What remains unique to this branch is `apply` and higher-order use (`(map list …)`), which #3016 leaves at today's speed because neither puts the constructor in call-head position. That is roughly 4.3us per call, and buying it still costs the same +22% of `core.phel` and the same red bench gate.

Leaving this parked rather than closing it, since that trade is worth a deliberate decision rather than a silent one.

PR #3013, opened 2026-08-09
docs: use canonical `:add` instead of `'add` in validate-with-schema.md

jasalt:
Small docs cleanup:

`instrument!` accepts any key for the name, but a keyword `:add` is canonical. The quoted symbol `'add` previously used here works at runtime but is non-idiomatic as every example in instrument.phel, schema.phel, and the test suite consistently uses keywords.


Chemaclass (approved):
Correct: `instrument!` takes "any key" and its own `:example` uses `:add`, so the doc was the odd one out. Thanks.

Issue #3037, opened 2026-08-10
LSP completion for PHP superglobals


jasalt:
Phel's PHP-interop completion supports PHP global functions such as
`php/str_replace`, but does not offer PHP superglobals such as `php/$_SERVER`.

## Expected behavior

When completion is requested after a PHP superglobal prefix, Phel LSP should
return applicable superglobal variables. For example:

```phel
(php/$_|)
```

should offer at least `$_SERVER`, and analogous prefixes should support the
standard PHP superglobals:

```text
$GLOBALS  $_SERVER  $_GET  $_POST  $_FILES  $_COOKIE  $_SESSION
$_REQUEST  $_ENV
```

The completion item should preserve the `php/` prefix, so selecting
`$_SERVER` produces `php/$_SERVER`.

## Headless reproduction

From a Phel checkout with dependencies installed:

```bash
php <<'PHP'
<?php
require 'vendor/autoload.php';

$resolver = new \Phel\Api\Application\PhpInteropContextResolver();

foreach (['(php/str', '(php/$_'] as $source) {
    $context = $resolver->resolve($source, 1, strlen($source) + 1);
    printf("%s => kind=%s prefix=%s\n", $source, $context->kind, $context->prefix);
}
PHP
```

Actual output:

```text
(php/str => kind=global-function prefix=str
(php/$_ => kind=none prefix=
```

The first input reaches PHP function completion. The second input has no PHP
interop completion context, so an LSP `textDocument/completion` request at
`(php/$_|)` returns no superglobal candidates.

## Cause

`PhpInteropContextResolver` recognizes PHP global-function completion with a
word-only prefix (`php\/(\w*)`). A PHP variable prefix begins with `$`, so it
does not match. In addition, `PhpInteropReflector::globalFunctions()` uses
`get_defined_functions()`, which only lists functions; PHP superglobals are
variables and cannot be obtained from that API.

## Suggested implementation

Add a distinct PHP-global-variable completion context for `php/$...`, rather
than treating superglobals as functions. Provide a static catalog of PHP
superglobals (their names are language-defined), filter it by the typed prefix,
and return normal LSP completion items for matching candidates.

The existing global-function path should remain unchanged. The LSP completion
range/item text should allow clients to replace only the portion after `php/`,
so the resulting Phel source retains the interop prefix.

## Acceptance criteria

- Completion at `php/$_|` includes `$_SERVER`.
[...]

PR #3038, opened 2026-08-10
feat: PHP superglobal LSP completion

jasalt:
Key changes:
 - Added a distinct global-variable PHP interop context.
 - Added the standard PHP superglobal catalog with prefix filtering.
 - Added variable-kind completion support.
 - Added LSP textEdit ranges that retain the php/ prefix.
 - Added resolver, reflector, and LSP integration coverage.
 - PHP function completion remains unchanged.

 Validation:
 - Focused tests pass.
 - PHPStan passes.
 - Code style check passes.
 - Full unit suite has unrelated environment failures due to missing BCMath and an inotify watcher expectation.

___

PoC implementation for https://github.com/phel-lang/phel-lang/issues/3037
AI used: Pi, gpt-5.6-sol (medium), prompt "Work on *issue text*"


jasalt:
Generated code not human tested / reviewed yet.
Testing this next with Emacs Eglot LSP client manually to see how it behaves...

jasalt:
Dang, I think earlier unnecessary ad-hoc edit got slipped into the PR.. need to clean up. (DONE)

jasalt:
Pleasing rector...

jasalt:
Looks good on client, had to do some config on that side to recognize the completion point correctly.

image

jasalt:
Seems usable now, continuing testing... 
Open for review with caveat that this should be treated as LLM coded PoC.

jasalt:
Changelog entry added.

Chemaclass (commented):
Reviewed the diff. The shape is right and it follows the Api module's existing conventions: a new `KIND_GLOBAL_VARIABLE` context rather than overloading the function path, the catalog behind `PhpInteropReflector::globalVariables()` next to its siblings, and prefix filtering reusing the existing case-insensitive `matches()` so it behaves like every other completion. Tests cover the resolver, the reflector and the LSP integration, and the public API snapshot is regenerated. All four workflows are green.

Two things before it lands.

**1. It conflicts with `main`, but only in `CHANGELOG.md`.** Nine PRs merged today, so the `## Unreleased` section moved under you. Nothing else conflicts: I test-merged locally and `CHANGELOG.md` is the only file. A rebase should be a one line resolution.

**2. `src/php/Api/CLAUDE.md` needs the two new facts.** The repo rule (`.claude/rules/modules.md`) asks for it when a module gains a key public method or a new constraint, and that file documents both collaborators you touched:

- the `PhpInteropReflector` row enumerates its methods, so `globalVariables()` belongs in it;
- the `PhpInteropContextResolver` row lists the positions it resolves, so the `php/$<name>` position belongs there.

Worth stating explicitly in that second row that a superglobal is a *language-defined catalog* rather than something reflected, since that is the one thing about this feature a reader would otherwise assume works like the classmap lookups around it.

Nothing blocking in the implementation itself. Happy to do either of the above if you would rather not.

Chemaclass:
Reviewed and pushed 5 follow-up commits. The feature itself was right: the context kind, the catalog, and the `textEdit` were all the correct calls, and the `php/$` branch is correctly ordered before the global-function branch (`$` is not a word character, so the function branch would never see it). What follows is what I changed and why.

### 1. The interop grammar was forked across two modules

`CompletionHandler::globalVariableRange()` carried a second copy of the resolver's `php/(\$\w*)$` regex. The Api owns that grammar; the handler only ever needed the shape of the `$...` token under the cursor. Narrowed it to `/\$\w*$/` and left the "is this a completable position" question where it belongs.

Two fixes fell out of the rewrite:

- The span now runs to the **end** of the token, so completing inside `php/$_SE|RVER` replaces the whole name instead of leaving a `RVER` tail behind.
- Reading the line through a new `Document::lineAt()` drops a second newline-splitting implementation and reuses the store's own CRLF normalisation. `Document::wordAt()` now shares it.

Also moved the `textEdit` construction out of the handler loop and into `CompletionConverter`, where the rest of the LSP wire shapes are built. The converter gates on `Completion::KIND_VARIABLE`, so the single range computed per request cannot leak onto the Phel symbols in the same list. `CompletionConverter` had no unit test at all; it has one now.

Small thing while in there: `PositionConverter` is the canonical home for `Range` and `TextEdit`, so `CompletionItem` imports them instead of inlining the shape.

### 2. Two surfaces of the same feature were dead

- **REPL and nREPL.** They run their own `php/` completion through `ReplCompleter`, which only knows functions and classes, so `php/$_S` returned nothing there.
- **Hover.** `phpInteropHoverAt` fell through to null, even though the LSP had just offered the symbol and every other interop kind answers it.

Both read the catalog now. To make that possible the nine names moved from a private const in `PhpInteropReflector` to `PhpSymbolCatalog::SUPERGLOBALS`, which already owns the other two `php/` symbol sets, mapped to a one-line description each so completion and hover have something to show beyond the bare label.

### 3. One real bug, found by a test

Hovering directly on the `$` sigil resolved nothing: `CursorText::wordEndColumn()` treated `$` as a word boundary and handed the resolver a bare `$` prefix. Teaching it to step over a leading sigil fixes superglobals and static properties (`\Foo/$bar`) alike, since both carry the sigil in their name.

### Verification

- Verified all nine names are genuinely reachable as `php/$NAME` by running them, not by assumption. `phel.core` and `phel.http` already use five of them.
- Pinned the resolver ordering with tests for the neighbours the new branch could have stolen: `php/str_rep`, `php/new \Foo`, `\Countab`, `Class/$prop`, plus the bare-sigil, outside-a-form, and string/comment cases.
- Full local gate green: php-cs-fixer, PHPStan, Psalm, Rector, 4605 unit tests, 7341 core tests. The BCMath and inotify failures noted in the description did not reproduce here.

The branch had also drifted into conflict; it is mergeable again.

Chemaclass:
One more, found by re-reviewing the neighbourhood of the fix rather than the fix itself.

`PhpInteropReflector::staticMembers()` labels a static property with its sigil (`$prop`, per ADR 0013) but tagged it `KIND_LOCAL`. Only `KIND_VARIABLE` earns the explicit `textEdit`, so accepting a static-property completion at `\Foo/$pr|op` left the typed `$` in place and wrote `\Foo/$$prop`. That is the same corruption the superglobal work fixes, one line away in the same file.

The sigil is the whole criterion, so `KIND_VARIABLE` now means "label carries a `$`", which covers superglobals and static properties alike. Instance properties stay `KIND_LOCAL`, since they complete without a sigil and a client's own word range already handles them. Both kinds map to the same LSP `CompletionItemKind`, so nothing changes on the wire beyond the edit itself.

This predates the PR, but it shares a root cause and costs one line, so it seemed worse to leave it sitting next to the fix. Happy to split it into a separate PR if you would rather keep this one scoped to superglobals.

Local gate green again, and CI was green on the previous push.

PR #3047, opened 2026-08-11
chore(deps): update all dependencies to gacela 2.2, and declare the module boundaries it can now check

Chemaclass:
## 🤔 Background

Routine dependency sweep, with `gacela-project/gacela` 2.0.0 → 2.2.0 as the one that actually moved something. 2.1 added Gacela's architecture rules to Psalm and 2.2 added declarative module-dependency rules, a config schema declaration and Symfony/Laravel bridges.

## 💡 Goal

Land the updates and keep both analysers green without weakening any rule.

## 🔖 Changes

`gacela` 2.0.0 → 2.2.0, plus `amp`, `paratest`, `php-cs-fixer`, `infection`, `phpstan`, `phpstan-deprecation-rules`, `rector`, `symfony/console` and `symfony/routing` to current patch/minor. No phel behaviour change.

### psalm.xml: the Gacela architecture rules 2.1 added

37 findings, in three new issue types. Every one is a decision **already argued site by site in `phpstan.neon`**, which has run the same three rules since 2.0. These entries mirror it rather than restating the reasoning:

| Psalm issue | phpstan identifier | Files |
|---|---|---|
| `GacelaFacadeInterfaceDrift` | `gacela.facadeInterfaceDrift` | Api, Build, Compiler, Interop, Run facades |
| `GacelaFacadeOnlyDelegates` | `gacela.facadeOnlyDelegates` | `Build/BuildFacade.php` |
| `GacelaSuffixExtends` | `gacela.suffixExtends` | the 8 pinned there, plus one |

Two mechanics worth knowing:

- They go through `<PluginIssue name="...">`. Psalm's XSD only accepts core issue names as elements, so `<GacelaSuffixExtends>` fails config parsing.
- **phpstan keeps sole ownership of the drift detection.** It pins each path with a `count`, so a genuinely new offender makes a count short by one and gets reported. Psalm has no count, so these are blanket per-file suppressions. A comment in `psalm.xml` says to pin the count there before adding a path here.

`Phel\Lang\TypeFactory` is on psalm's `suffixExtends` list and not phpstan's, because phpstan does not report it. `Lang` is a leaf with no Gacela pillars at all (`src/php/CLAUDE.md`), so it is the same category as the other eight. The lists are pinned against what each analyser actually reports rather than kept identical, and the file says so.

### rector 2.6 deleted `SetList::STRICT_BOOLEANS`

The set file and its boolean rules are both gone from the package. Its one survivor, `DisallowedEmptyRuleFixerRector`, is in `CodeQualityLevel::RULES`, which `SetList::CODE_QUALITY` already loads, so dropping the set loses nothing that still ships.

rector's updated `CODE_QUALITY` then rewrote one guard:

```php
-            if ($shared === null) {
+            if (!$shared instanceof LocalVarNode) {
```

`parseEqualityCall` is annotated `array{0: LocalVarNode, 1: mixed}|null` and `$shared` is only ever assigned from it, so `?LocalVarNode` makes the two identical.

### Held back on purpose

`phpunit` stays at 13.1.14. 13.3.0 requires `sebastian/diff ^9.0` and `vimeo/psalm` 6.16.1 caps it at `^8.0`; `paratest` 7.24.0 follows phpunit. Both wait for psalm rather than being forced.

### New 2.2 features, assessed separately

No new gacela command surfaced in `phel list`: `FrameworkCommands.php` registers a fixed set, so `debug:graph`, `debug:config` and `stubs:publish` are not exposed. The Symfony and Laravel bridges do not apply. The declarative module-rules file and `declareConfigSchema()` are being evaluated on their own merits rather than adopted because they are new.

### Validation

`composer test` exits 0: phpstan and psalm clean, rector and cs-fixer clean, 4641 PHPUnit, 7350 core tests.


Chemaclass:
Second commit adopts the one 2.2 feature that earns its place: `module-rules.json`, gacela's declarative module-dependency rules, read by **both** analysers.

Five boundaries that until now existed only in prose:

| Rule | Already written in |
|---|---|
| nothing imports `Console` | `Console/CLAUDE.md`: "the graph's sink ... nothing depends on it. Keep it that way" |
| `Filesystem` is a leaf | `Filesystem/CLAUDE.md` |
| `Fiber` is a leaf | `Fiber/CLAUDE.md` |
| `Compiler` reaches only `Filesystem` outside the shared kernel | `Compiler/CLAUDE.md` + ADR 0004 |
| `Command` reaches only `Compiler`, type-only | `Command/CLAUDE.md` |

Nothing new was decided, so no ADR. Each rule held before it was added.

**Proven live, not assumed.** Replacing the `Command` rule with `allow: []` reports, in both readers:

```
phpstan: TextExceptionPrinter.php:36  Phel\Command must not depend on Phel\Compiler
         gacela.declaredModuleDependency
psalm:   GacelaDeclaredModuleDependency - TextExceptionPrinter.php:36:1
```

The shipped rules report nothing. A boundary check that quietly turns itself off is the failure this kind of gate actually has, so it got a control.

Two things worth knowing, both now in the config comments:

**phpstan's result cache is not keyed on `module-rules.json`**, and a cached finding keeps the `reason` string it was written with. `--clear-cache` does not help, it clears a different cache; `phpstan clear-result-cache` does. Three runs read green while the file said otherwise before I worked that out. CI starts cold, so this bites locally only.

**`Lang`, `Shared` and `Config` cannot be governed.** They are declared `sharedNamespaces`, and gacela's CLI graph derives modules from Facade classes, which those three do not have. Rules naming them would be inert under the analysers and would fail `debug:graph --check --rules` outright. The `sharedNamespace` lists in `phpstan.neon` and `psalm.xml` must stay identical for one file to mean one thing.

`debug:graph --check --rules` is deliberately **not** in CI: it gates cycles in the same pass and would need an allowed-cycles file for the four pairs ADR 0004 accepts, which `ModuleDependencyCycleTest` already pins.

### On the other 2.2 features

`declareConfigSchema()`: skipped. `validate:config` does print "No schema declared", so it would light up, but it means restating the 21 keys `PhelConfig`'s typed properties already enforce, with nothing keeping the two in sync, and `debug:config` (where most of the payoff is) is not registered in `FrameworkCommands`. Two sources of truth to make an advisory command print more.

`swapModuleFactory()`/`swapModuleConfig()`/`swapModuleProvider()`, `stubs:publish`, and the Symfony and Laravel bridges: no current use here.

Issue #3059, opened 2026-08-12
LSP navigation to Phel namespace from namespace identifier


jasalt:
- Phel v0.49.0-beta#c0bdab8
- Emacs 30.2 eglot LSP client

Phel LSP server doesn't seem provide navigation to namespace file from namespace identifiers, only from namespace members.

Example:
```clojure
(ns demo
  (:require phel.pprint :refer [pprint]))
```

When cursor is on top of `phel.pprint` namespace identifier and LSP navigation to the symbol is triggered (e.g. Emacs `xref-find-definitions`), the navigation source is not found as LSP server is returning `nil` to client triggering `xref--not-found-error: No definitions found for: LSP identifier at point`. Server response parsed as `nil` by Eglot LSP client:

```emacs-lisp
(eglot--request
     (eglot-current-server)
     :textDocument/definition
     '(:textDocument
       (:uri "file:///path/to/demo.phel")
       :position (:line 1 :character 17)))
;; => nil
```

When cursor is on top of latter `pprint` after `:refer:`, LSP navigation works as expected opening the `pprint.phel` namespace. Functioning server response parsed by Eglot LSP client:

```emacs-lisp
(eglot--request
     (eglot-current-server)
     :textDocument/definition
     '(:textDocument
       (:uri "file:///path/to/demo.phel")
       :position (:line 1 :character 35)))
;; => (:uri "file:///path/to/out/phel/pprint.phel" :range (:start (:line 99 :character 5) :end (:line 99 :character 11)))
```

jasalt:
Phel LSP Eglot client session with navigation failing for extra context. 

https://github.com/user-attachments/assets/741b8c82-8af9-4efd-a141-1e43d6e7d3d7

Showcases also that navigation to user namespace from `'supplier-product-curator.domain.catalog` fails while it works from symbol with namespace member `supplier-product-curator.domain.catalog/transition-result`.

Chemaclass:
Verified fixed on main (`e354b60c3`), shipped in 0.50.0 by #3060 and #3158. The PR said "Issue:" rather than "Closes", which is why this stayed open.

Drove `textDocument/definition` end to end against a two-file project for every shape in the report and the video, cursor on the namespace itself each time:

| cursor on | resolves to |
|---|---|
| `phel.pprint` inside `(:require ...)` | `pprint.phel:1` (the `ns` form) |
| `app.domain.catalog` inside `(:require ...)` | `catalog.phel:1` |
[...]

PR #3060, opened 2026-08-12
Add LSP definition navigation for namespace identifiers

jasalt:
Index namespace declaration locations and resolve namespace identifiers in textDocument/definition requests, allowing navigation from require forms to the corresponding namespace source file.

Issue: https://github.com/phel-lang/phel-lang/issues/3059

___

**Phel version:** `v0.49.0-beta#69de235`

Fixed LSP definition navigation for namespace identifiers.

Previously, navigating from a namespace in a require form such as:

```clojure
(ns demo
  (:require phel.pprint :refer [pprint]))
```

returned `nil` when the cursor was on `phel.pprint`. Navigation only worked from referenced members such as `pprint`.

The LSP index now records namespace declaration locations and resolves namespace identifiers—including dotted and internal namespace spellings—to their corresponding source files. Member navigation remains unchanged.

Added coverage for namespace extraction, namespace resolution, namespace-only indexes, and the end-to-end LSP definition handler. Targeted tests pass: **26 tests, 52 assertions**.

___
AI assisted PR, Pi session: https://pi.dev/session/#8a372efbde6f39e243b652427b9090e4


jasalt:
LSP client testing looks good, comparing to https://github.com/phel-lang/phel-lang/issues/3059#issuecomment-5263267079 now navigation to namespace symbols works to both Phel stdlib and user namespaces. 

[Screencast_20260812_095213.webm](https://github.com/user-attachments/assets/a1f24942-2163-40dd-9090-62db70808936)

jasalt:
Fixing rector issues.

jasalt:
Fixed the PHP-CS-Fixer CI issue

jasalt:
Need to ensure these run locally next time, Pi coding agent doesn't seem to pick them up automatically. 
EDIT: User error on my part, didn't run `./bin/phel agent install claude` which advices Pi also to run `composer fix` in context. Expected that Pi had picked up something without any installs (like in case of AGENTS.md in project root), but project requires explicit install command first.

jasalt:
Open for review.
Vibe coded on my part (Pi & gpt-5.6-sol medium) so needs proper architectural review.
Continuing with manual testing using Eglot LSP client...

Chemaclass:
Reviewed and pushed a follow-up commit. The feature and its coverage were sound; the storage choice had a collision I could not merge past.

**The blocker.** The `ns` declaration was stored in `ProjectIndex::$references` under a `"<namespace>/"` key. `SymbolKey::resolve()` builds that exact string from an empty symbol name:

```php
return $namespace === '' ? $symbol : $namespace . '/' . $symbol;
```

So `findReferences($index, 'phel.pprint', '')` returned the `ns` declaration as if it were a reference. Two ways to reach it:

- `JsonRpcDispatcher` coerces a missing `symbol` param to `''`, so any daemon client that omits the field hits it.
- `RenameHandler` feeds those locations straight into `buildWorkspaceEdit`, using the location's real `endCol`. That emits a text edit spanning the namespace name inside a different file's `(ns ...)` form.

**The fix.** `ProjectIndex` now carries its own `$namespaceLocations`, keyed by `Munge::canonicalNs()`:

- `references` keeps its documented `namespace/name` shape, so the collision is gone and its two readers and two JSON consumers stop seeing a phantom entry.
- Lookup is one hash hit. The previous fallback scanned every reference in the project on each miss, on the interactive go-to-definition path.
- The `\` to `.` normalization reuses `Munge::canonicalNs()` instead of an inline `str_replace` in an LSP handler, and it now happens at the write site.
- `findNamespace` moved off `Lsp\...\SymbolResolver` onto the index itself. It is an index query, not a cursor-word heuristic, so Api, Lint and the daemon get it too.

**Two things worth flagging:**

`namespaces()` needed a guard. Definition namespaces keep the source spelling (`fixtures\foo`) while the new keys are canonical (`fixtures.foo`), so a naive merge counted one namespace twice. `countNamespaces()` now only grows for namespaces that genuinely have no definitions.

The public API snapshot caught the change. Adding a trailing optional constructor parameter is non-breaking per `docs/stability.md`, so the snapshot is updated and the CHANGELOG gained the matching `Added` line. Worth noting the sentinel-key version passed that gate silently while still changing the `phel index --output` JSON: the shape shifted without the surface test having anything to see.

Also added an integration test over the real indexer. Every other test hand-built the index, so nothing proved the pipeline populated it from actual source. The fixture declares `(ns fixtures\foo)`, which exercises the backslash path with real data.

Chemaclass:
Second pass, from a correctness and test-quality review. Two things it caught.

**The feature was unreachable for backslash-spelled namespaces.** `Document::isIdentChar()` did not include `\`, so `wordAt` split the name before it ever reached the index:

```
(:require fixtures\foo :refer [greet])   cursor => 'fixtures'
(fixtures\foo/greet 1)                   cursor => 'foo/greet'
```

Neither matches an index key. The storage layer resolved both spellings correctly and no cursor could deliver them. The second line is the older bug: qualified symbols written with `\` have never resolved from a click, in definition, hover, find-references or rename alike. ADR 0008 keeps `\` parsing everywhere it used to, so the LSP quietly not supporting it was an inconsistency rather than a policy.

One character in the identifier class fixes both. The test that pinned the old boundary was a characterization of the regex rather than a decision: it arrived in a cleanup pass (#1526) and its comment restated the behavior instead of defending it. It now asserts the whole word, and a backslash-spelled require is covered end to end through `wordAt` rather than only at the data layer.

**One test I added could not fail.** `ReferenceFinder::find()` only ever reads `references`, so asserting it ignores a populated `namespaceLocations` proved nothing: it passes whether or not any of this exists. Replaced with a guard over the real indexer, which asserts no reference key ends in `/` and that `findReferences($index, $ns, '')` comes back empty. I verified that one has teeth by restoring the sentinel write and watching it fail:

```
Failed asserting that 'fixtures\uses-core/' ends not with "/".
```

Also added coverage for the precedence rule, since definition lookup runs before namespace lookup and that is now user-observable.

**Known limitation, not fixed here:** two files declaring the same namespace keep only one location, and which one depends on directory traversal order. `references` appends across files, `namespaceLocations` overwrites. Worth a follow-up if duplicate `ns` declarations are something the index should report rather than silently collapse.

Local `composer test-compiler` and `composer test-quality` both green.

PR #3120, opened 2026-08-13
refactor(core): use string.phel functions over php interop

jasalt:
Prefer phel.string functions over equivalent PHP string interop where semantics are compatible. Document intentional raw PHP usage where byte indexing, core bootstrap dependencies, TTY inspection, or performance make the native operation preferable.

Related issue: https://github.com/phel-lang/phel-lang/issues/2941

## 🤔 Background

Phel core used PHP string functions directly even where equivalent `phel.string` functions existed. This made it unclear whether native PHP interop was intentional or an oversight.

## 💡 Goal

Prefer Phel standard-library string functions where semantics are compatible, while documenting intentional raw PHP usage.

## 🔖 Changes

- Replaced applicable PHP string calls with `phel.string` functions.
- Used higher-level predicates such as `starts-with?`, `includes?`, and `ends-with?`.
- Replaced suitable character mappings with `escape`.
- Preserved raw PHP for byte-oriented operations, core bootstrap dependency constraints, TTY inspection, and performance-sensitive paths.
- Added rationale comments for retained PHP interop.
- Fixed an `s/subs` end-index regression in AI response extraction.
- Updated affected tests and examples.

All affected tests pass.

### Benchmark

Ran the Phel benchmark suite before and after the commit:

```sh
bin/phel bench tests/phel/bench/core.phel \
  --revs=10000 --iterations=7 --warmup=2
```

Result summary:

- All 9 benchmarks completed successfully in both runs.
- Post-change means were lower across all measured benchmarks, ranging from approximately **4% to 26% faster**.
- No benchmark regression was observed.
- Benchmark variability remained low, with relative standard deviation below **1.5%**.
- These benchmarks are general core benchmarks rather than direct microbenchmarks of every converted string call, so results should be treated as indicative rather than a definitive performance claim.

### DIFF
--- a/src/phel/ai.phel
+++ b/src/phel/ai.phel
@@ -396,7 +396,10 @@ Useful for building multi-turn conversations."
 (defn- extract-json-from-response
   "Extracts JSON from an AI response, handling markdown code blocks."
   [response]
-  (let [trimmed (s/trim response)]
+  ;; Native `trim`: `phel.string/trim` is a `/u` regex that returns nil on a
+  ;; subject that is not valid UTF-8, and a truncated model stream is exactly
+  ;; the case these functions exist to report as a `RuntimeException`.
+  (let [trimmed (php/trim response)]
     (if (s/starts-with? trimmed "{")
       trimmed
       (let [matches (php-indexed-array)
@@ -411,7 +414,8 @@ Useful for building multi-turn conversations."
 (defn- extract-json-array-from-response
   "Extracts a JSON array from a model response, slicing between the first '[' and the last ']' when the array isn't already at the start."
   [response]
-  (let [trimmed (s/trim response)]
+  ;; Native `trim`, for the same reason as `extract-json-from-response` above.
+  (let [trimmed (php/trim response)]
     (if (s/starts-with? trimmed "[")
       trimmed
       (slice-between trimmed "[" "]" "Could not extract JSON array from AI response: "))))
--- a/src/phel/ai.phel
+++ b/src/phel/ai.phel
@@ -1,5 +1,6 @@
 (ns phel.ai
   (:require phel.http-client :as hc)
+  (:require phel.string :as s)
   (:require phel.json)
   (:use RuntimeException))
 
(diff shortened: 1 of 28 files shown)


Chemaclass:
Nice cleanup, and thanks for documenting *why* the raw PHP stays in the places it does. Three things from measuring this against `main`.

## 1. `(keyword ns nm)` is about 1.95x slower

The one behavioural change in `src/phel/core/booleans.phel`:

```phel
-(Keyword/createForNamespace (when-not (php/=== nil ns) (php/strval ns))
-                            (php/strval nm))
+(Keyword/createForNamespace (when-not (php/=== nil ns) (str ns))
+                            (str nm))
```

`str`'s one-argument arity is `(val-to-str x)`, a Phel function call, where `php/strval` compiles to a direct PHP call. Two of them per namespaced keyword.

Interleaved A/B against `main`, three alternating pairs, 32 constructions per revolution:

| | `php/strval` | `str` |
|---|---|---|
| `(keyword "ns" "nm")` | 12.67us, 12.67us, 12.73us | 24.95us, 24.69us, 25.26us |
| `(keyword "nm")` (control) | 13.06us | 12.98us |

The one-argument arity is untouched by the change and stays flat, which is what says the gap is real rather than drift.

Not necessarily a blocker, it depends how hot you consider namespaced keyword construction, but it is a measurable cost rather than a pure refactor, so it seemed worth surfacing rather than leaving to be found later.

## 2. The two conversions disagree on non-strings

```
(php/strval true)  => "1"       (str true)  => "true"
(php/strval false) => ""        (str false) => "false"
```

Same for both on strings, symbols and floats. So `(keyword true)` and friends change meaning. Exotic input, and arguably `str` is the better answer, but it is a semantic change hiding inside a refactor commit.

## 3. This branch predates #3121

It is based on 9ae62b980 (#3112). #3121 landed after that and also touched `core/booleans.phel` and `core/math.phel`, adding `^bool` return tags to fifteen predicates. A rebase onto current `main` will conflict in both files.

Happy to help with the rebase or to take the `keyword` hunk separately if that is easier.

jasalt:
@Chemaclass Nice, thank's for benchmarking. For 1. I'd keep the faster php/strval with tiny hint on function docs about the performance reason.

jasalt:
Also, good catch for the differences. Good to avoid semantic changes, I tried to avoid them with this but only did checking quite quickly so some slipped.

jasalt:
You can finish it off @Chemaclass with the rebase, thank's.

Chemaclass:
Finished off, thanks @jasalt. Rebased onto current `main` (`612bc523d`), no conflicts.

**1.** `keyword` keeps `php/strval`, with the hint in the docstring since the rendering is part of the contract. The 1.95x number stays in a `;;` comment, out of `phel doc`.

**2.** Five more conversions turned out not to be equivalent, so those are back on the native call, each with a comment and a test:

| | why |
|---|---|
| `test.phel` `escape-for-string-diff` | `replace` is `preg_replace`; `\\` in a replacement is one backslash, so it stopped doubling |
| `http.phel` `content-type-base` | `trim` is a `/u` regex: nil on invalid UTF-8, and leaves the NUL `php/trim` strips |
| `http.phel` `headers-from-server` | `starts-with?` asserts string, and PHP casts numeric array keys to int |
| `ai.phel` extractors | same `trim` nil, replaced the `RuntimeException` with an assertion error |
| `base64.phel` | padding sized in characters, applied in bytes; `escape` also 161x slower than `strtr` |

**3.** Two smaller ones: moved the rationale out of `re-pattern` and `char`'s docstrings (both public, it was shipping to the site) into `;;` comments, and fixed `file-seq`'s `:example`, which used an `s/` alias that cannot resolve inside `phel.core`.

The rest of the sweep is untouched and reads better for it. Gate green: `test-core` 8243, unit 4673, integration 1227, static analysis clean, lint 0 errors.

Chemaclass (approved):
Rebased, gate green, and the non-equivalent conversions are reverted with tests. Good to go.

PR #3131, opened 2026-08-13
feat(lsp): let binding navigation

jasalt:
## 🤔 Background

LSP go-to-definition previously only resolved project-indexed definitions and namespaces. Lexically bound locals such as `let` and `loop` bindings could not navigate to their binding site.

## 💡 Goal

Provide correct go-to-definition navigation from local usages to their lexical binding symbols.

## 🔖 Changes

- Add `LocalBindingResolver`, using best-effort reader forms to resolve `let` and `loop` locals.
- Respect progressive binding scope, nested shadowing, repeated binding names, destructuring, and qualified-symbol exclusion.
- Support vector, rest, nested, and map destructuring (`:keys`, `:as`, direct keys); avoid treating `:or` default expressions as bindings.
- Resolve locals before project-index definitions in `DefinitionHandler`.
- Allow local definition resolution before project indexing is complete.
- Only recognise un-namespaced `let`/`loop` heads, matching compiler dispatch.
- Add API facade method, public API snapshot, API docs, CHANGELOG entry, and comprehensive resolver/LSP tests.

## Demo

[Screencast_20260814_011756.webm](https://github.com/user-attachments/assets/ea8d4819-c3b7-4dbd-b393-f02e10abfc7c)

### Follow-up

`fn`, `foreach`, `catch` and the `if-let` family are modelled as scopes. `defn` (multi-arity) and `for`/`dofor` (patterns interleaved with verbs and expressions) are not: they hide the names they rebind so nothing resolves to an outer binding of the same name, and resolving them properly is still open.

Hover, find-references and rename do not handle locals at all yet.


Chemaclass:
Reviewed this in depth and pushed the fixes onto the branch. The feature is good. It had six defects, all reproduced against the real resolver before touching anything.

**The one that mattered most: it returned wrong answers.**

Only `let` and `loop` opened a scope. Every other binder was walked with the outer scope intact. So go-to-definition on the inner `a` here jumped to the outer binding:

```phel
(let [a 1] (fn [a] a))
(let [a 1] (foreach [a xs] a))
(let [a 1] (for [a :in xs] a))
```

The follow-up note reads as "no jump yet". It was a wrong jump. Worse than nothing.

`if-let` and `when-let` had the same shape of problem, and they are everywhere in `cli.phel`, `test.phel` and `router.phel`. Unrecognised head, so the body usage fell through to `SymbolResolver::find`, which scans every namespace by bare name and lands on whatever global matches first.

The rule now is: **be right or be silent.** `fn`, `foreach`, `catch` and the `if-let`/`when-let`/`if-some`/`when-some`/`when-first` family are modelled directly. `if-let` and `if-some` expand their `else` outside the binding, so the else branch does not see it. `defn` may be multi-arity and `for`/`dofor` interleave patterns, verbs and expressions, so their shapes are not modelled. They hide the names they rebind instead. Same for `binding`, which rebinds dynamic vars rather than introducing locals. A name they cover resolves to nothing, never to an outer binding.

**The caret fix is the one users will feel.**

`isAt()` compared against an exclusive end column with `<`. `Document::wordAt()` still reports the word when the caret rests one past the last character, which is exactly where an editor leaves it after you type or double-click a name. Navigation died there. One character to the left it worked.

The rest:

- `walk()` never descended into set literals, so a usage inside `#{...}` was invisible.
- No notion of quoting, so `'(let [x 1] x)` was walked as code and answered with a location inside inert data.
- The binding pattern itself was never walked, so a cursor on a declaration fell through to the project index and could jump to an unrelated global. It resolves to itself now.
- `:or` values are default expressions evaluated in the enclosing scope. They were skipped, not walked, so the class comment overclaimed. They are walked as usages now.
- `toLocation()` sized the range with `strlen()`. The lexer counts columns in codepoints, so a multibyte name overshot its end.
- The `.` marker exclusion had no basis. `&` is the only marker any deconstructor recognises.

**Perf.** Qualified words now return before the parse. `Symbol::getName()` never contains a `/`, so they could never match a local, yet they paid the full whole-buffer read: 15ms on `seq-fns.phel`, on every request.

**Tests.** 20 new resolver cases, plus `LocalDefinitionLspTest` driving the real resolver through the real handler. `DefinitionHandlerTest` stubs `resolveLocalBinding` with a hand-written `Location`, so nothing checked that the resolver's 1-based `Location` and the 0-based LSP range agree. Every new test was mutation-checked: revert the fix, the test fails.

Two things I left alone.

Hover, find-references and rename still do nothing on a local. Worth a follow-up.

And a pre-existing off-by-one, unrelated to this PR but adjacent. `SymbolExtractor` passes the raw 0-based `SourceLocation` column straight into `Location`, but `PositionConverter` subtracts one on the way out. Go-to-definition on a `def`, document symbols, find-references and rename all report a column one to the left. `SymbolExtractorTest` pins the current behaviour, so it needs its own PR.

Chemaclass (approved):
Reviewed in depth, fixed six defects on the branch, CI green. Details in the comment above.

PR #3237, opened 2026-08-18
feat(balance): add conservative boundary and surplus closer repairs

jasalt:
Attempt at improving the `phel balance` command based on cases it was failing with coding agent use with a budget llm model. Generated code, may contain logical mistakes. Continuing with manual testing..

_____

### Background

`phel balance --fix` previously repaired only by appending missing closing delimiters at EOF. That is unsafe when a later top-level form follows an incomplete form: appending can silently nest that later form instead of restoring the intended boundary.

### Solution

This PR keeps append-only repair as the default and adds explicit opt-in strategies:

```text
phel balance file --fix --repair=boundary
phel balance file --fix --repair=delete-unexpected
```

- `append` remains the default and repairs only when appending at EOF is safe.
- `boundary` inserts missing closers before a detected subsequent top-level form.
- `delete-unexpected` removes exactly one surplus closer when it yields valid, balanced Phel.

Mismatched closers are always refused: changing either the opener or the closer can yield valid syntax, so selecting either would guess at program intent.

Extended repair candidates are lexed, parsed with `parseAll()`, and balance-scanned again before writing. Writes are atomic and preserve existing permissions where possible. Boundary repair preserves LF and CRLF line endings.

### Compatibility

`BalanceFacade::balance(array $paths, bool $fix = false)` remains compatible. Its new optional `RepairStrategy` parameter defaults to `RepairStrategy::Append`; existing CLI behavior is unchanged.

### Tests

Coverage includes:

- Boundary insertion before a subsequent definition.
- CRLF-preserving boundary insertion.
- Surplus closer deletion.
- Refusal of mismatched closers, including detection-only runs.
- Existing append-only behavior.

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -14,7 +14,7 @@ enables tab-completion (setup in the [README](../README.md)).
 | `agent-install` | Install agent skill files (Claude, Cursor, Codex, Gemini, Copilot, Aider) into the current project |
 | `analyze` | Run semantic analysis on a single Phel source file and emit JSON diagnostics |
 | `api-daemon` | Long-running JSON-RPC daemon exposing the Api semantic-analysis facade over stdio (for tooling) |
-| `balance` | Report unbalanced `()`, `[]`, `{}` in Phel files; append the missing closers with `--fix` |
+| `balance` | Report unbalanced `()`, `[]`, `{}` in Phel files; repair with `--fix` (`--repair=append|boundary|delete-unexpected`) |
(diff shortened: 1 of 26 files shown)


jasalt:
Unsure if these are ideal for Phel core or should exist in separate tool.
Anyways, continuing experimenting with this branch for the time being.

jasalt:
Decided keep this as experiment outside Phel for now. Having multiple balancing command options does not seem like the best dx so will reconsider later if there’s more clear idea.

PR #3274, opened 2026-09-09
fix(repl): accept (exit) and (quit), and say so in the banner

### LINKED ISSUE #3272: fix(repl): the banner says to type `exit`, and `(exit)` is an unresolved symbol
## Summary

The REPL banner tells the user to type `exit`. Bare `exit` works. `(exit)`, which is what a Lisp user reaches for, is an unresolved symbol with three unrelated suggestions.

## Repro

Phel `main` (1194f22d1), PHP 8.5.10, macOS.

```sh
printf '(exit)\nexit\n' | phel repl
```

```
Welcome to the Phel Repl (v0.51.0)
[PHEL001] Cannot resolve symbol 'exit'. Did you mean 'empty', 'eval', or 'next'?
in string:1

1| (exit)
    ^^^^

Bye!
```

`exit` is handled as a REPL command string in `src/php/Run/Infrastructure/Command/ReplCommand.php`, not as a definition. Nothing in `src/phel/` defines `exit` or `quit`, so any parenthesised form fails.

## Why it matters for 1.0

It is the first thing a user types after the banner tells them to, in the shape the language trains them to use, and it produces an error naming `empty`, `eval` and `next`. Clojure users reach for `(exit)` or `(System/exit 0)`; Python and Node both accept the call form alongside the bare word. Small, and it is the last impression of every REPL session.

## Proposal

Accept `(exit)` and `(quit)` as the same REPL command the bare words already are, and mention both forms in the banner:

```
Type (exit) or press Ctrl-D to exit.
```

Handling it at the REPL command layer keeps `exit` out of `phel.core`, which matters because `docs/spec/language-surface.md` freezes that surface for 1.x. If it is preferable to have a real definition, it belongs in `phel\repl` rather than `phel\core`.

Optionally accept `(exit <code>)` and use it as the process exit code, which is what makes the call form worth having over the bare word.

## Acceptance

- A test asserting `(exit)`, `(quit)`, `exit` and `quit` all end the session with status 0.
- A test asserting the banner names the form the session accepts.
- A test asserting `(exit 2)` exits with status 2, if that part is taken.

dchaudhari7177:
Fixes #3272.

Takes the proposal as written: handled at the REPL command layer, not as a definition, so `exit` stays out of `phel\core` and `docs/spec/language-surface.md` keeps the surface it freezes for 1.x.

```
Welcome to the Phel Repl (v0.51.0)
Type (exit) or press Ctrl-D to exit.
```

`exit`, `quit`, `(exit)` and `(quit)` all end the session. `(exit <code>)` was the optional part and I took it, because it is the only thing the call form can do that the bare word cannot — `ExitException` carries the status, `loopReadLineAndAnalyze()` returns it, and `execute()` returns that instead of a fixed `self::SUCCESS`.

## Matching exactly, and not more

The risk in a change like this is swallowing code that merely mentions `exit`. The match is anchored and the pattern is the whole of the logic:

```php
private const string EXIT_PATTERN = '/^(?:exit|quit|\(\s*(?:exit|quit)(?:\s+(\d+))?\s*\))$/';
```

Run against every shape I could think of:

```
input                  result
---------------------------------------------
'exit'                 EXIT status 0
'quit'                 EXIT status 0
'(exit)'               EXIT status 0
'(quit)'               EXIT status 0
'  (exit)  '           EXIT status 0
'( exit )'             EXIT status 0
'(exit 2)'             EXIT status 2
'(exit  7 )'           EXIT status 7
'(quit 1)'             EXIT status 1
'(exit-code)'          evaluate
'exits'                evaluate
'(println "exit")'     evaluate
'(exit x)'             evaluate
'(exit 2 3)'           evaluate
'exit 2'               evaluate
'(exitquit)'           evaluate
'(map exit xs)'        evaluate
''                     evaluate
'(+ 1 2)'              evaluate
```

`(exit x)` and `(exit 2 3)` staying ordinary code is deliberate: the command understands one optional integer, and a form it does not understand should fail as a form rather than quietly end the session on an argument it ignored.

## Tests

`tests/php/Integration/Run/Command/Repl/ReplExitTest.php`, covering the three acceptance criteria and one more:

- all four spellings, plus whitespace variants, end the session with status 0 and print `Bye!`
- `(exit 2)` exits with status 2
- the banner names a form the session accepts — it looks the banner up in the transcript and asserts `(exit)` appears in it, so the two cannot drift apart again
- the six near-misses above reach a second prompt instead of ending the session

They follow `ReplBareStdClassTest` and use the existing `ReplCommandTestTrait` / `ReplTestIo` wiring.

## Verification gap, stated plainly

**I could not run the suite.** `composer.json` requires PHP `>= 8.4` and the code uses `new Foo()->bar()`, which is 8.4 syntax; this machine has 8.3.6, where `php -l` fails on `ReplCommand.php` *on unmodified `main`* as well. So the tests are for CI to run.

What I did verify locally is the part I actually wrote: the table above comes from extracting `EXIT_PATTERN` out of the committed file with a regex and running the same `trim()` + `preg_match()` the command runs, so it is the shipped pattern being tested, not a retyped copy of it.

## Also changed

`CHANGELOG.md` gains an entry under Unreleased → Fixed.

## Note

This PR was written with AI assistance. I understand the change and can defend or revise it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- REPL: `(exit)` and `(quit)` end the session, as `exit` and `quit` already did, and the banner names the call form. `(exit <code>)` sets the process status. (#3272)
+- REPL: `(exit)` and `(quit)` end the session like `exit` and `quit`, `(exit <code>)` sets the exit status, and the banner names the call form. (#3272)
 
 ## [0.51.0](https://github.com/phel-lang/phel-lang/compare/v0.50.0...v0.51.0) - 2026-09-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+### Fixed
+
+- REPL: `(exit)` and `(quit)` end the session, as `exit` and `quit` already did, and the banner names the call form. `(exit <code>)` sets the process status. (#3272)
+
 ## [0.51.0](https://github.com/phel-lang/phel-lang/compare/v0.50.0...v0.51.0) - 2026-09-05
 
 ### Added
--- a/src/php/Run/Domain/Repl/ExitException.php
+++ b/src/php/Run/Domain/Repl/ExitException.php
@@ -23,9 +23,6 @@ final class ExitException extends RuntimeException
         return new self('Exit from REPL!', $status);
     }
 
-    /**
-     * The status the process should leave with, so `(exit 2)` can mean 2.
-     */
     public function status(): int
     {
         return $this->status;
--- a/src/php/Run/Domain/Repl/ExitException.php
+++ b/src/php/Run/Domain/Repl/ExitException.php
@@ -11,8 +11,23 @@ use RuntimeException;
  */
 final class ExitException extends RuntimeException
 {
-    public static function fromRepl(): self
+    private function __construct(
+        string $message,
+        private readonly int $status,
+    ) {
+        parent::__construct($message);
+    }
+
+    public static function fromRepl(int $status = 0): self
     {
-        return new self('Exit from REPL!');
+        return new self('Exit from REPL!', $status);
+    }
+
+    /**
+     * The status the process should leave with, so `(exit 2)` can mean 2.
+     */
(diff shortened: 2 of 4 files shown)


Chemaclass (approved):
thanks!!

PR #3299, opened 2026-09-10
feat(compiler)!: reject php/new, php/->, php/:: and set-var as source

Chemaclass:
## 🤔 Background

The consolidation comment on #2859 lists `php/new`, `php/->`, `php/::` and `set-var` under "**1.0 removes**". Three of them cannot be removed. The analyzer emits them:

```
NAME_PHP_NEW                  created by the compiler: 1 site
NAME_PHP_OBJECT_CALL          created by the compiler: 2 sites
NAME_PHP_OBJECT_STATIC_CALL   created by the compiler: 2 sites
```

`(new \C 1)` becomes `(php/new \C 1)`, `(.m o)` becomes `(php/-> o (m))`, and the `binding` macro expands to `set-var`. Deleting them leaves the compiler with nothing to emit, which is why ADR 0007 rejected removal in as many words.

Leaving the deprecation as it was is the other bad option: a notice behind `--warn-deprecations` for the whole of `1.x` means the migration never completes and the removal list stays untrue.

## 💡 Goal

Make the removal list honest without breaking the compiler. From the outside it reads as removal; internally nothing moves.

## 🔖 Changes

**The rule.** Writing one of the four is an error carrying `PHEL012`, with no flag to turn it off. `php/` at large is untouched: `php/aget`, `php/aset`, `php/apush`, `php/aunset` and their `-in` variants, `php/oset`, `php/ref`, `php/callable` and every `(php/some_function …)` are exactly as they were. That is the rule ADR 0007 protects, `php/` marks host access.

```
$ phel eval '(php/new \DateTime)'
[PHEL012] "php/new" is no longer valid source for constructing a PHP object. Use "(new \Foo arg)" or "(\Foo. arg)" instead.
in string:1

1| (php/new \DateTime)
   ^^^^^^^^^^^^^^^^^^^
```

**How the two roles are told apart.** By the phase that produced the form. The check runs in the reader, on each top-level form as it is read:

- the reader only sees typed source, so every head it finds was written by somebody
- macro output is built during analysis and never passes through the reader, so `binding` keeps emitting `set-var` and `set!` keeps emitting `php/::`
- a quasiquote template is already lowered, so the name sits in it as quoted data, not as a list head
- a plain `quote` is skipped: `'(php/new \C)` is data

A first version checked in the analyzer and told the roles apart by source location. It rejected `set-var` from a user test calling `binding` (PHAR smoke). `BreakSymbol` and `FnPrePostConditionRewriter` no longer need unlocated heads.

**Everything else that had to move**, all of it a consequence rather than a choice:

- The stdlib's own five uses of these forms were rewritten to the public spellings. The one that stays is the `php/::` inside `set!`, which is its compilation target.
- 48 emitter fixtures wrote the target forms as input. Rewritten to the public spelling and regenerated; a check confirms no fixture's expected output changed for any reason other than the input rewrite or the message change below.
- The diagnostics that named a rejected spelling moved to #3300 (merged).
- The opt-in deprecation channel has no syntax deprecation left. The tests that used `php/new` as their example now use `to-php-array`, a deprecated definition.

**Recorded as [ADR 0018](docs/adr/0018-superseded-forms-are-rejected-as-source.md)**, amending ADR 0007. It also records two consequences worth knowing: a method call on a local compiles to an `is_string` ternary because the shorthand cannot prove the target is not a class name, and the editor still completes forms that can no longer be written.

## ✅ Tests

- `SupersededFormRejectorTest`: each of the four heads, a head nested in a larger form, a quoted subtree, a `php/*` form that stays.
- `SupersededFormRejectionTest`: every Clojure-style shorthand still compiles, every rejected form is rejected. The old deprecation test became this.
- `LanguageSurfaceSpecTest` pins the spec table against the rejector.

Full gate green: 10125 core assertions, unit and integration suites, static analysis. `PharExecutionTest` green against a fresh PHAR.

Refs #2859


Chemaclass:
Marking this draft. The approach does not hold, and the reason is worth writing down before anyone retries it.

Rejecting these four as source needs the compiler to tell "somebody wrote this" from "the compiler generated this". I tried both discriminators available today, and each one fails in a different environment:

**Expansion origin** (what this branch currently does). Ask whether `SourceLocation::getExpansionOrigin()` points inside `src/phel`. Works in a source checkout. Fails inside the PHAR, where that path is `phar://…/src/phel` and the check comes out false, so **every `deftest` in a scaffolded project stops compiling**:

```
$ phel.phar init --no-interaction && phel.phar test
[PHEL012] "set-var" is no longer valid source for a var's root value.
in tests/main_test.phel:5
```

**No source location.** Have the stdlib build its heads with `(symbol "set-var")` and `(symbol "php/::")` instead of quoting them, so a generated head carries no location and the rule becomes structural rather than path-based. Fixes the PHAR. Breaks under **nested** macro expansion: `deftest` expands to `binding`, which expands to `set-var`, and the outer expansion stamps a location onto the freshly built symbol. Same failure, now everywhere rather than only in the PHAR.

So the prerequisite is a real marker for a compiler-generated form, one that survives macro expansion. Metadata on the head symbol is the obvious candidate. That is its own change, and it should land before this one rather than inside it.

What is already worth keeping from this branch, independent of the decision:

- `BreakSymbol` and `FnPrePostConditionRewriter` copy a source location onto heads their sibling expressions deliberately leave unlocated.
- `(.-$foo o)` reports `Cannot resolve symbol '.-$foo'`. The message explaining that a sigil names a static property is reachable only through `php/->`, and it recommends `(php/:: \Foo $foo)`, which this change would forbid.
- The dynamic-`new` runtime error says `php/new expects a class name or object`.
- Five `.phel` fixture projects under `tests/php/Integration` still wrote these forms and passed only because their `.phel/cache` directories were warm from before the change. A clean checkout compiles them.

One consequence for the plan on #2859: step 2 there says settling the `php/*` wording is "no code". That is right for the wording, but the removal step behind it is not a small change.

JesusValeraDev (approved):

